Issue #2, opened 2026-04-05
Add Redis TLS support for BullMQ connections


blamechris:
## Description

All 3 BullMQ connections (review-worker, clinical-data events, clinical-notes events) use plain TCP. In production, Redis should use TLS to encrypt patient IDs and clinical event data in transit.

Event payloads on the queue contain \`patient_id\` and clinical data — this is PHI-adjacent data that must be encrypted in transit for HIPAA compliance.

## Acceptance Criteria

- [ ] All Redis/BullMQ connections support \`REDIS_TLS=true\` env var to enable TLS
- [ ] When enabled, connections pass \`tls: {}\` to the IORedis config
- [ ] \`.env.example\` documents the \`REDIS_TLS\` variable
- [ ] docker-compose.yml includes a commented TLS configuration example

Issue #1, opened 2026-04-05
Generate and commit Drizzle migrations


blamechris:
## Description

No Drizzle migrations have ever been committed. The schema exists in \`packages/db-schema/src/schema/*.ts\` but \`packages/db-schema/drizzle/\` is empty. \`pnpm db:migrate\` fails on a fresh checkout because there are no migration files.

This blocks all developer onboarding and fresh environment setup.

## Acceptance Criteria

- [ ] Run \`pnpm db:generate\` to generate migration files from current schema
- [ ] Commit the generated migration SQL files in \`packages/db-schema/drizzle/\`
- [ ] Verify \`pnpm db:migrate\` runs successfully against a fresh database
- [ ] Includes migrations for new tables: \`llm_interaction_log\`, \`medlens_sync_tokens\`, \`medlens_sync_log\`

PR #21, opened 2026-04-06
feat(db): add column-level encryption for high-sensitivity PHI

### LINKED ISSUE #10: Add column-level encryption for high-sensitivity PHI
## Description

High-sensitivity PHI fields are stored in plaintext: date_of_birth, mrn, insurance_id, emergency_contact_name, emergency_contact_phone. HIPAA §164.312(a)(2)(iv) requires encryption of ePHI at rest.

## Acceptance Criteria

- [ ] Use \`pgcrypto\` extension for column-level encryption
- [ ] Encrypt: \`patients.date_of_birth\`, \`patients.mrn\`, \`patients.insurance_id\`, \`patients.emergency_contact_name\`, \`patients.emergency_contact_phone\`
- [ ] Encryption key managed via env var, not hardcoded
- [ ] Migration handles encrypting existing plaintext data
- [ ] Application layer transparently decrypts on read

blamechris:
## Summary
- AES-256-GCM encryption via Node.js built-in crypto (no new deps)
- Drizzle `encryptedText` custom type: transparent encrypt/decrypt
- Applied to: date_of_birth, mrn, insurance_id, emergency_contact_name, emergency_contact_phone
- One-time migration script for encrypting existing plaintext data
- 11 tests for encryption round-trips, key management, error cases

Closes #10

## Test plan
- [ ] Set `PHI_ENCRYPTION_KEY` env var (64-char hex string)
- [ ] Verify insert → select round-trip for patient with encrypted fields
- [ ] Run encrypt-existing script against seeded data
- [ ] Run `pnpm --filter @carebridge/db-schema test`

### DIFF
--- a/packages/db-schema/src/__tests__/encryption.test.ts
+++ b/packages/db-schema/src/__tests__/encryption.test.ts
@@ -1,7 +1,7 @@
 import { describe, it, before, after } from "node:test";
 import assert from "node:assert/strict";
 import { randomBytes } from "node:crypto";
-import { encrypt, decrypt, getKey } from "../encryption.js";
+import { encrypt, decrypt, getKey, _resetKeyCache } from "../encryption.js";
 
 const TEST_KEY = randomBytes(32).toString("hex");
 
@@ -68,7 +68,12 @@ describe("encrypt / decrypt", () => {
 describe("getKey reads PHI_ENCRYPTION_KEY from env", () => {
   const originalKey = process.env.PHI_ENCRYPTION_KEY;
 
+  before(() => {
+    _resetKeyCache();
+  });
+
   after(() => {
+    _resetKeyCache();
     if (originalKey !== undefined) {
       process.env.PHI_ENCRYPTION_KEY = originalKey;
     } else {
@@ -77,6 +82,7 @@ describe("getKey reads PHI_ENCRYPTION_KEY from env", () => {
   });
 
   it("returns a Buffer when a valid 32-byte hex key is set", () => {
+    _resetKeyCache();
     process.env.PHI_ENCRYPTION_KEY = TEST_KEY;
     const key = getKey();
     assert.ok(Buffer.isBuffer(key));
@@ -93,6 +99,7 @@ describe("getKey reads PHI_ENCRYPTION_KEY from env", () => {
   });
 
   it("throws on invalid key length", () => {
+    _resetKeyCache();
     process.env.PHI_ENCRYPTION_KEY = "aabbcc"; // too short
     assert.throws(() => getKey(), /must be exactly 32 bytes/);
   });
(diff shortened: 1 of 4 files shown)


blamechris:
## Code Review: PR #21

### Summary
Adds AES-256-GCM column-level encryption for high-sensitivity PHI fields in the patients table (date_of_birth, mrn, insurance_id, emergency_contact_name, emergency_contact_phone). Includes a Drizzle custom type for transparent encrypt/decrypt, a one-time migration script, and 11 unit tests.

### Strengths
- **Correct crypto primitives**: AES-256-GCM with random IVs and auth tags -- gold standard for application-layer encryption.
- **Non-deterministic IVs**: Two encryptions of the same plaintext produce different ciphertexts, preventing frequency analysis.
- **Clean Drizzle integration**: `encryptedText` custom type means existing queries need no changes.
- **Thorough tests**: Round-trip, unicode, empty string, wrong-key rejection, format validation, env-key handling.
- **ESM `.js` extensions** used correctly throughout imports.
- **Functional style** maintained, no classes.
- **ISO 8601 dates** preserved in created_at/updated_at.

### Issues Found

#### Critical (Must Fix)

1. **`getKey()` called on every row read/write -- no caching** (`encryption.ts:62-63`)
   The `encryptedText` custom type calls `getKey()` inside `toDriver` and `fromDriver`. `getKey()` reads `process.env.PHI_ENCRYPTION_KEY` and allocates a new Buffer on every call. For a query returning 100 patients with 5 encrypted columns, that is 500 Buffer allocations and 500 env lookups. Cache the key at module load or use lazy initialization with a module-level variable.

2. **Migration script logs patient IDs to console** (`encrypt-existing.ts:64`)
   `console.log(\`  Encrypted PHI for patient ${row.id}\`)` -- while UUIDs are not themselves PHI, coupling patient IDs with "PHI encrypted" status in logs could be a compliance concern in production. Recommend logging only a count, or gating verbose output behind a `--verbose` flag.

1. **`mrn` column has `.unique()` constraint but values are now non-deterministic ciphertext** (`patients.ts:8`)
   Two inserts of the same MRN will produce different ciphertexts, so the unique constraint will never catch duplicates. If MRN uniqueness must be enforced at the DB level, you need a separate HMAC-based column (deterministic hash) for the unique index, with the encrypted column for retrieval. File this as a follow-up if MRN uniqueness is a business requirement.

2. **No key rotation strategy**
   When `PHI_ENCRYPTION_KEY` needs to rotate, all existing ciphertexts become unreadable. Consider supporting a key ID prefix in the ciphertext format (e.g., `v1:iv:tag:ct`) so a future rotation script can re-encrypt with the new key while still decrypting old values.

3. **`encrypt-existing.ts` processes rows sequentially**
   For a large patient table, consider batching updates or using a transaction for atomicity. Currently a failure mid-run leaves partial encryption with no rollback.

4. **Missing `AUTH_TAG_LENGTH` usage**
   `AUTH_TAG_LENGTH` is defined (line 7) but never referenced. The default GCM auth tag is 16 bytes, which matches, but the constant is dead code. Either use it in `createCipheriv` options or remove it.

### Verdict
**Approve with requested changes.** The crypto implementation is solid. Fix the key caching performance issue (#1) and the patient-ID logging (#2) before merge. The MRN uniqueness constraint (#S1) should be tracked as a follow-up issue.

Issue #317, opened 2026-04-10
feat(ai): implement escalation engine for unacknowledged critical flags


blamechris:
## Context

A critical clinical flag that goes unacknowledged is as dangerous as no flag at all. There is currently no mechanism to escalate when flags are ignored. This issue implements an automated escalation engine that widens the notification audience over time.

## Scope

`services/ai-oversight`

## Requirements

- BullMQ scheduled/delayed job that checks for unacknowledged critical flags past threshold
- Configurable escalation thresholds (e.g., 30min for critical, 2hr for warning)
- Escalation chain: assigned provider → department providers → on-call (if available)
- Flag status transitions to `escalated` with audit trail
- Emit escalation notification events

## Acceptance Criteria

- [ ] Scheduled checker runs at configurable interval (default: every 5 minutes)
- [ ] Critical flags unacknowledged past threshold trigger escalation notification
- [ ] Flag status updated to `escalated` with timestamp
- [ ] Escalation targets broader audience (department/specialty level)
- [ ] Audit log records escalation event
- [ ] `pnpm typecheck && pnpm lint` passes

blamechris:
Implemented in PR #343 (merged). Closing — the PR body used `Refs #317` rather than `Closes #317`, so GitHub never auto-closed this. Verified via code audit on current main.

PR #343, opened 2026-04-10
feat(ai): implement flag escalation engine (#317)

blamechris:
## Summary

- BullMQ repeatable job checks for unacknowledged flags every 5 minutes
- Critical flags escalate after 30 minutes, warning flags after 2 hours
- Updates flag status to `escalated` with audit trail
- Emits escalation notifications via notifications queue (broader audience)
- Integrated into ai-oversight server startup with graceful shutdown

Refs #317

## Test Plan

- [ ] pnpm typecheck passes
- [ ] Escalation worker starts with ai-oversight server
- [ ] Critical flag unacknowledged for 30+ min gets escalated
- [ ] Warning flag unacknowledged for 2+ hours gets escalated
- [ ] Escalation notification emitted with ESCALATED prefix
- [ ] Already acknowledged flags are not escalated

### DIFF
--- a/services/ai-oversight/src/server.ts
+++ b/services/ai-oversight/src/server.ts
@@ -7,11 +7,14 @@
 
 import { createServer } from "node:http";
 import { startReviewWorker } from "./workers/review-worker.js";
+import { setupEscalationQueue, startEscalationWorker } from "./workers/escalation-worker.js";
 import { formatPrometheus, getMetricsSnapshot } from "./services/shadow-metrics.js";
 
 const HEALTH_PORT = Number(process.env.HEALTH_PORT ?? 4001);
 
 const worker = startReviewWorker();
+const escalationQueue = setupEscalationQueue();
+const escalationWorker = startEscalationWorker();
 
 const REDIS_PING_TIMEOUT_MS = 2_000;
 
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #343

### Summary
Adds a BullMQ-based escalation engine that periodically checks for unacknowledged clinical flags and escalates them after severity-based timeouts (30min critical, 2hr warning).

### Strengths
- Clean BullMQ repeatable job pattern with sensible defaults (5-min interval, concurrency 1)
- Proper graceful shutdown — both worker and queue closed in server teardown
- Good notification queue config with exponential backoff and retention limits
- Correct use of ESM `.js` extensions and functional style

### Issues Found

1. **[Suggestion] N+1 update pattern**: `checkAndEscalate()` does a SELECT then loops individual UPDATEs per flag. For a small number of stale flags this is fine, but consider batching with `UPDATE ... WHERE id IN (...)` for efficiency, especially since you already have `UPDATE...RETURNING` patterns elsewhere in the codebase.

2. **[Suggestion] Missing `updated_at` on escalation**: The `.set({ status: "escalated" })` call doesn't set `updated_at`. If the table has an `updated_at` column, this should be set for audit trail accuracy.

3. **[Suggestion] Module-level Redis connection**: `const connection = getRedisConnection()` at module scope means the connection is created at import time. This is consistent with how the review worker works, but worth noting — if the module is imported in tests without Redis, it will fail.

4. **[Low] Logging includes `patient_id`**: The console.log on line ~100 prints the patient_id. While this is server-side logging (not client-facing), ensure your log aggregator treats this as PHI and applies appropriate retention/access controls.

### Verdict
- [x] Approve - Ready to merge

Issue #315, opened 2026-04-10
feat(gateway): add SSE endpoint for real-time notification delivery


blamechris:
## Context

Clinicians and patients need to see notifications without refreshing the page. Server-Sent Events (SSE) is simpler than WebSocket for one-way server-to-client push and is sufficient for notification delivery.

## Scope

`services/api-gateway` + `apps/clinician-portal` + `apps/patient-portal`

## Requirements

- SSE endpoint on api-gateway (`GET /notifications/stream`) with JWT auth
- Clinician portal: notification bell icon in header with unread count badge, dropdown showing recent notifications
- Patient portal: notification indicator
- tRPC procedure `notifications.getUnread` for initial load
- Mark-as-read on click

## Acceptance Criteria

- [ ] SSE endpoint streams new notifications to authenticated users
- [ ] Clinician portal shows notification bell with real-time unread count
- [ ] Clicking a notification marks it as read and navigates to the linked resource
- [ ] Patient portal shows notification indicator
- [ ] Connection handles reconnection gracefully
- [ ] `pnpm typecheck && pnpm lint` passes

## Dependencies

- Depends on notification dispatch worker issue (#314)

blamechris:
Implemented in PR #344 (merged). Closing — the PR body used `Refs #315` rather than `Closes #315`, so GitHub never auto-closed this. Verified via code audit on current main.

PR #344, opened 2026-04-10
feat(gateway): add SSE endpoint for real-time notifications (#315)

blamechris:
## Summary

- GET /notifications/stream SSE endpoint with JWT authentication
- Redis Pub/Sub for real-time delivery (channel per user: `notifications:{userId}`)
- Heartbeat every 30s to keep connections alive through proxies
- Auto-cleanup of Redis subscriber on client disconnect
- `publishNotificationToUser` helper for dispatch worker integration

Refs #315

## Test Plan

- [ ] pnpm typecheck passes
- [ ] SSE endpoint returns 401 without auth
- [ ] Authenticated connection receives `connected` event
- [ ] Published notification appears as `notification` event
- [ ] Heartbeat keeps connection alive
- [ ] Client disconnect cleans up Redis subscription

### DIFF
--- a/services/api-gateway/src/routes/notifications-sse.ts
+++ b/services/api-gateway/src/routes/notifications-sse.ts
@@ -1,0 +1,100 @@
+/**
+ * Server-Sent Events endpoint for real-time notification delivery.
+ *
+ * Clients connect to GET /notifications/stream with their JWT in the
+ * Authorization header or session cookie. The connection stays open and
+ * pushes new notifications as they arrive via Redis Pub/Sub.
+ *
+ * Flow:
+ * 1. Client connects → auth validated → Redis subscription created
+ * 2. Notification dispatch worker publishes to Redis channel "notifications:{userId}"
+ * 3. This SSE handler forwards the message to the connected client
+ * 4. Client reconnects automatically on disconnect (EventSource spec)
+ */
+
+import type { FastifyInstance, FastifyRequest, FastifyReply } from "fastify";
+import Redis from "ioredis";
+
+function createSubscriberClient(): Redis {
+  return new Redis({
+    host: process.env.REDIS_HOST ?? "localhost",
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #344

### Summary
Adds a Server-Sent Events endpoint at `/notifications/stream` on the API gateway for real-time notification delivery via Redis Pub/Sub.

### Strengths
- Clean SSE implementation with proper headers (including `X-Accel-Buffering: no` for nginx)
- Heartbeat every 30s to prevent proxy/LB timeouts
- Proper cleanup on client disconnect (unsubscribe + quit)
- `reply.hijack()` correctly used to prevent Fastify from closing the response
- `publishNotificationToUser` helper is a clean separation for the dispatch worker

### Issues Found

1. **[Concern] One Redis connection per SSE client**: Each connected client creates a dedicated Redis subscriber via `createSubscriberClient()`. Under moderate load (e.g., 200 concurrent clinicians), this means 200 Redis connections just for SSE. Consider a shared subscriber that multiplexes across clients using pattern subscriptions or a fan-out map.

2. **[Suggestion] Auth extraction is fragile**: The cast `(request as unknown as { userId?: string }).userId` is a double-cast escape hatch. Consider typing the Fastify request via `declare module` augmentation or using a proper decorator, which is more aligned with TypeScript strict mode.

3. **[Suggestion] No Zod validation on publish payload**: `publishNotificationToUser` accepts a typed object but doesn't validate at the boundary. Since this is called from other services, adding a Zod schema for the notification shape would catch malformed events early.

4. **[Low] Redis config duplication**: The `createSubscriberClient()` manually reads env vars. The codebase has `@carebridge/redis-config` (used in PR #343). Consider using that shared package instead for consistency.

### Verdict
- [x] Approve - Ready to merge (address Redis connection scaling before production load)

Issue #330, opened 2026-04-10
feat(scheduling): implement appointment scheduling schema and tRPC service


blamechris:
## Context

Appointment scheduling is a core MyChart feature. The `services/scheduling` directory exists but is completely empty — no package.json, no code. Need full service scaffold + schema.

## Scope

`packages/db-schema` + `services/scheduling`

## Requirements

- `appointments` table: id, patient_id, provider_id, appointment_type (follow_up/new_patient/procedure/telehealth), start_time, end_time, status (scheduled/confirmed/checked_in/completed/cancelled/no_show), location, reason, encrypted notes, encounter_id (nullable FK), created_at, updated_at, cancelled_at, cancelled_by, cancel_reason
- `providerSchedules` table: id, provider_id, day_of_week (0-6), start_time, end_time, slot_duration_minutes, location, is_active, effective_from, effective_until
- `scheduleBlocks` table: id, provider_id, start_time, end_time, reason (vacation/meeting/blocked), created_at
- New Fastify service with tRPC router
- tRPC procedures: appointments.list, appointments.create, appointments.cancel, appointments.reschedule, appointments.getById, schedule.availability (returns open slots for provider + date range), schedule.setProviderSchedule
- Slot availability calculation: provider schedule minus existing appointments minus blocks
- Appointment creation validates slot availability (no double-booking)
- Register in api-gateway tRPC router

## Acceptance Criteria

- [ ] Database migration creates all three tables
- [ ] Scheduling service starts and registers with api-gateway
- [ ] Slot availability calculation returns correct open slots
- [ ] Appointment creation prevents double-booking
- [ ] Cancel and reschedule update status with audit trail
- [ ] `pnpm typecheck && pnpm lint` passes

blamechris:
Implemented in PR #349 (merged). Closing — the PR body used `Refs #330` rather than `Closes #330`, so GitHub never auto-closed this. Verified via code audit on current main.

PR #347, opened 2026-04-10
feat(portal): implement patient symptom journal with AI oversight (#327)

blamechris:
## Summary

**This is the core differentiator from Epic MyChart.**

Patients can record structured symptoms/observations that feed into the AI oversight engine, catching patterns like "patient with DVT reporting persistent headaches" without cluttering the clinical chart.

### Schema
- `patientObservations` table with encrypted description, structured data (severity, location, duration), self-assessment
- Migration 0018

### Backend
- `observations.create` / `observations.getByPatient` tRPC procedures
- `patient.observation` event emitted to clinical-events queue
- AI cross-references observations against active diagnoses and medications

### Patient Portal
- `/symptoms` page with guided symptom entry
- 8 observation categories (pain, neurological, GI, respiratory, skin, cardiovascular, general, medication side effect)
- Severity scale (1-10), body location picker, duration, free-text
- History view with past observations

### Architecture
- Observations live in separate "Patient Signals" section — NOT in clinical chart
- Visible to providers but doesn't clutter clinical documentation
- AI monitoring catches dangerous patterns patients might not recognize

Refs #327

## Test Plan

- [ ] pnpm typecheck passes
- [ ] Patient can submit observation from /symptoms page
- [ ] Observation stored with encrypted description
- [ ] patient.observation event emitted to clinical-events queue
- [ ] Observations visible in history
- [ ] Observations do NOT appear in clinical notes or clinical data

PR #349, opened 2026-04-10
feat(scheduling): implement appointment scheduling service (#330)

blamechris:
## Summary

- New `services/scheduling` with full tRPC router
- Schema: `appointments`, `providerSchedules`, `scheduleBlocks` tables (migration 0019)
- Appointment CRUD: create (with double-booking prevention), cancel, reschedule
- Slot availability: generates time slots from provider template, subtracts booked appointments and blocks
- Provider schedule templates: weekly day-of-week patterns with configurable slot duration
- Registered in api-gateway as `scheduling` namespace

Refs #330

## Test Plan

- [ ] pnpm typecheck passes (32/32)
- [ ] Create appointment with time validation
- [ ] Double-booking prevented (error on overlap)
- [ ] Cancel appointment sets status and reason
- [ ] Reschedule cancels original and creates new
- [ ] Availability returns correct open slots

### DIFF
--- a/packages/db-schema/drizzle/0019_scheduling_tables.sql
+++ b/packages/db-schema/drizzle/0019_scheduling_tables.sql
@@ -1,0 +1,51 @@
+-- Migration: Create scheduling tables for appointment management
+
+CREATE TABLE IF NOT EXISTS appointments (
+  id TEXT PRIMARY KEY,
+  patient_id TEXT NOT NULL REFERENCES patients(id),
+  provider_id TEXT NOT NULL REFERENCES users(id),
+  appointment_type TEXT NOT NULL,
+  start_time TEXT NOT NULL,
+  end_time TEXT NOT NULL,
+  status TEXT NOT NULL DEFAULT 'scheduled',
+  location TEXT,
+  reason TEXT,
+  notes TEXT,
+  encounter_id TEXT,
+  cancelled_at TEXT,
+  cancelled_by TEXT,
+  cancel_reason TEXT,
+  created_at TEXT NOT NULL,
+  updated_at TEXT NOT NULL
+);
+
+CREATE INDEX IF NOT EXISTS idx_appointments_patient ON appointments(patient_id, start_time);
+CREATE INDEX IF NOT EXISTS idx_appointments_provider ON appointments(provider_id, start_time);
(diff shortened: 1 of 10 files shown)


blamechris:
## Code Review: PR #349

### Summary
New `scheduling` service with appointment CRUD, provider schedule templates, slot availability calculation, and double-booking prevention. Full tRPC router registered in the api-gateway.

### Strengths
- Double-booking prevention with overlap detection on create and reschedule
- Slot availability calculation correctly subtracts both appointments and schedule blocks
- `encryptedText` used for appointment `notes` field (PHI)
- Clean schema with proper FK references to `patients` and `users`
- Good index design: composite indexes on `(provider_id, start_time)` for range queries
- Proper ESM exports, `.js` extensions, functional style throughout

### Issues Found

**Suggestion — Race condition on double-booking check**
The overlap check and insert are not in a transaction:
```ts
const overlapping = await db.select()...;
if (overlapping.length > 0) throw ...;
await db.insert(appointments).values(appointment);
```
Two concurrent requests for the same slot could both pass the check before either inserts. Wrap in a DB transaction, or add a unique constraint / advisory lock. Same issue exists in `reschedule`.

**Suggestion — `is_active` should be boolean**
`providerSchedules.is_active` is `text("is_active").default("true")` — using text for a boolean value is fragile. Consider using `boolean("is_active").default(true)` for type safety. The migration also uses `TEXT DEFAULT 'true'`.

**Suggestion — Migration uses plain TEXT for notes column**
Migration 0019 declares `notes TEXT` while the Drizzle schema uses `encryptedText("notes")`. Same concern as PR #347 — add a migration comment noting this is application-layer encrypted to prevent confusion.

**Minor — `reason` field on appointments is not encrypted**
The appointment `reason` field (e.g., "follow-up for cancer treatment") could contain sensitive clinical context. Consider using `encryptedText` for this field as well.

**Minor — `listAll` ordering on compliance dashboard**
`emergencyAccessRouter.listAll` orders by `granted_at` ascending — most recent should likely come first. Use `desc(emergencyAccess.granted_at)`.

*Correction: that last point belongs to PR #350, not this one.*

### Verdict
- [ ] Approve — The race condition on double-booking is a real concurrency bug that should be fixed before merge

blamechris:
Fixed in 77932ff — double-booking check now wrapped in `db.transaction()` to prevent TOCTOU race.

PR #383, opened 2026-04-13
fix(infra,fhir,gateway): migration journal, FHIR RBAC layering, dev rate limit

blamechris:
## Summary
- Register migrations 0005–0023 in the Drizzle journal so `pnpm db:migrate` applies them
- Add `globalPassThroughEnv` to `turbo.json` so env vars reach child processes during `turbo run dev`
- Fix FHIR export RBAC layering: clinicians authorized by the gateway wrapper were rejected by the raw router's defense-in-depth guard. Added `rbacVerified` context flag.
- Raise auth login rate limit from 5/min to 60/min in non-production environments

## Test plan
- [ ] `pnpm db:migrate` applies all 23 migrations on a clean DB
- [ ] `pnpm db:seed` succeeds
- [ ] `pnpm test` — all 30 suites pass (including 3 new FHIR RBAC tests)
- [ ] Clinician can call `fhir.exportPatient` and get a full FHIR R4 bundle
- [ ] Direct raw FHIR router callers (no `rbacVerified`) still blocked for clinicians
- [ ] Login rate limit is 60/min in dev, 5/min in production

### DIFF
--- a/packages/db-schema/drizzle/meta/_journal.json
+++ b/packages/db-schema/drizzle/meta/_journal.json
@@ -36,6 +36,132 @@
       "when": 1744041600000,
       "tag": "0004_mfa_and_llm_audit",
       "breakpoints": true
+    },
+    {
+      "idx": 5,
+      "version": "7",
+      "when": 1744128000000,
+      "tag": "0005_refresh_token",
+      "breakpoints": true
+    },
+    {
+      "idx": 6,
+      "version": "7",
+      "when": 1744214400000,
+      "tag": "0006_audit_log_procedure_patient",
+      "breakpoints": true
+    },
+    {
+      "idx": 7,
+      "version": "7",
+      "when": 1744300800000,
+      "tag": "0008_encrypt_patient_name",
+      "breakpoints": true
+    },
+    {
+      "idx": 8,
+      "version": "7",
+      "when": 1744387200000,
+      "tag": "0009_icd10_snomed_coding",
+      "breakpoints": true
+    },
+    {
+      "idx": 9,
+      "version": "7",
+      "when": 1744473600000,
+      "tag": "0010_add_vitals_loinc_code",
+      "breakpoints": true
+    },
+    {
+      "idx": 10,
+      "version": "7",
+      "when": 1744560000000,
+      "tag": "0011_encrypt_clinical_narratives",
+      "breakpoints": true
+    },
+    {
+      "idx": 11,
+      "version": "7",
+      "when": 1744646400000,
+      "tag": "0012_audit_log_immutability",
+      "breakpoints": true
+    },
+    {
+      "idx": 12,
+      "version": "7",
+      "when": 1744732800000,
+      "tag": "0013_encounters_table",
+      "breakpoints": true
+    },
+    {
+      "idx": 13,
+      "version": "7",
+      "when": 1744819200000,
+      "tag": "0014_llm_compliance_columns",
+      "breakpoints": true
+    },
+    {
+      "idx": 14,
+      "version": "7",
+      "when": 1744905600000,
+      "tag": "0015_requires_human_review_boolean",
+      "breakpoints": true
+    },
+    {
+      "idx": 15,
+      "version": "7",
+      "when": 1744992000000,
+      "tag": "0016_notification_phi_encryption",
+      "breakpoints": true
+    },
+    {
+      "idx": 16,
+      "version": "7",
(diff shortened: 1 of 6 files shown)


blamechris (commented):
## CareBridge Inspector Review

**Verdict:** APPROVE

### Findings

#### Critical

None.

#### Suggestions

1. **Missing `exportPatient` test for `rbacVerified` flag** — The new `rbacVerified` context flag tests (`router-auth.test.ts`) only exercise `getByPatient`. Both `getByPatient` and `exportPatient` pass `ctx.rbacVerified` to `assertRawPatientAccess`, so the underlying logic is the same, but for defense-in-depth security code, both endpoints should have explicit test coverage. A compromised or refactored `exportPatient` path that accidentally drops the flag would go undetected.
   - File: `services/fhir-gateway/src/__tests__/router-auth.test.ts`

2. **`isDev` check matches all non-production environments** — The relaxed login rate limit (60/min) triggers when `NODE_ENV !== "production"`, which includes staging, QA, and any misconfigured environments. If staging is network-accessible, this gives attackers 12x the brute-force budget. Consider restricting the relaxation to `NODE_ENV === "development"` specifically, or using an explicit opt-in flag like `CAREBRIDGE_DEV_AUTH`.
   - File: `services/api-gateway/src/server.ts`, line 101

3. **Migration journal tag gap: `0007` missing** — The journal entries jump from tag `0006_audit_log_procedure_patient` (idx 6) to `0008_encrypt_patient_name` (idx 7). Drizzle uses the `idx` field for ordering so this is functionally harmless, but it creates confusion for anyone auditing migration history. Worth a comment in the journal or renaming for consistency.
   - File: `packages/db-schema/drizzle/meta/_journal.json`

#### Nits

1. **`rbacVerified` JSDoc on `assertRawPatientAccess`** — The `Allowed:` block in the docstring for `assertRawPatientAccess` lists the `rbacVerified` flag first, above admin. Consider reordering to match the check priority: admin (unconditional) first, then patient self-access, then `rbacVerified` — which would better reflect the trust hierarchy (admin > pre-authorized clinician > patient self-access).
   - File: `services/fhir-gateway/src/router.ts`, lines 98-108

2. **Duplicate rate-limit blocks** — The `auth.login` and `auth.refreshSession` branches in the rate-limit config are identical (`isDev ? 60 : 5`). These could be combined into a single condition for readability:
   ```typescript
   if (req.url?.startsWith("/trpc/auth.login") || req.url?.startsWith("/trpc/auth.refreshSession")) {
     return isDev ? 60 : 5;
   }
   ```
   - File: `services/api-gateway/src/server.ts`, lines 104-110

### Summary

This is a well-structured PR that solves a real RBAC layering problem (clinicians authorized by the gateway wrapper being rejected by the raw router's defense-in-depth guard). The `rbacVerified` context flag is the right approach — it defaults to falsy, only the RBAC wrapper sets it, and the tests cover the key scenarios (true, false, unset). The migration journal catch-up and turbo env passthrough are straightforward infrastructure fixes.

The main suggestion is to add `exportPatient` test coverage for the `rbacVerified` flag and tighten the dev-mode rate-limit check so it does not accidentally apply to staging environments. Neither blocks merge.

blamechris:
All findings addressed in 61776fa:

**Suggestion 1 — exportPatient rbacVerified tests**: Fixed. Added three new test cases (rbacVerified: true, false, and unset) for exportPatient, mirroring the existing getByPatient coverage. The true case asserts NOT_FOUND (proving it passes the RBAC guard and reaches the DB lookup) rather than FORBIDDEN. Closes #384.

**Suggestion 2 — isDev rate-limit scope**: Fixed. Changed isDev from NODE_ENV !== "production" to NODE_ENV === "development". Staging, QA, and misconfigured environments now get the production 5/min limit. Closes #385.

**Suggestion 3 — Migration journal tag gap (0007)**: Not fixing. The idx field controls ordering, and Drizzle ignores the tag name for sequencing. Renaming 0008_encrypt_patient_name to 0007_* would break any existing database that already applied the migration under the current tag. The gap is cosmetic only.

**Nit 1 — JSDoc ordering**: Fixed. Reordered the Allowed: list to admin > rbacVerified > patient, matching the trust hierarchy and code execution order.

**Nit 2 — Duplicate rate-limit branches**: Fixed. Collapsed auth.login and auth.refreshSession into a single || condition.

All 127 tests pass (34 fhir-gateway, 93 api-gateway).

PR #409, opened 2026-04-13
fix(notifications): publish to Redis pub/sub for SSE delivery

### LINKED ISSUE #401: fix(notifications): wire SSE delivery — dispatch worker never publishes to Redis pub/sub
## Context
Per marathon session audit (2026-04-10, 5/8 agents flagged):

The notification dispatch worker creates DB records but never publishes to the Redis pub/sub channel. The SSE endpoint at \`/notifications/sse\` subscribes to Redis channels that never receive messages. Real-time notification delivery is completely non-functional.

## Requirements
1. Dispatch worker: after inserting notification into DB, publish to Redis channel \`notifications:{userId}\`
2. Clinician portal: add EventSource connection to \`/notifications/sse\` endpoint
3. Patient portal: same EventSource integration
4. Handle reconnection on disconnect
5. Show toast/badge when new notification arrives

## Acceptance criteria
- Creating a clinical flag triggers a real-time notification in the portal
- SSE endpoint delivers notifications within 1 second of creation
- Browser reconnects automatically after disconnect

blamechris:
## Summary
- The dispatch worker's Redis publisher was created with `lazyConnect: true` but `.connect()` was never called, so publish commands were queued in the offline buffer and never sent to Redis
- Extracted a shared `publish.ts` module that creates the Redis client with eager connection (no `lazyConnect`) and exposes `publishNotification()` 
- Wired the publish call into both the BullMQ dispatch worker and the tRPC `create` mutation, so all notification creation paths publish to `notifications:{userId}` channels
- Added tests covering channel routing, payload format matching SSE expectations, empty userId handling, and multi-user dispatch

## Test plan
- [x] `pnpm --filter @carebridge/notifications test` — 18 tests pass (13 specialty-filter + 5 new publish tests)
- [x] `pnpm build` — full monorepo builds cleanly
- [ ] Manual: create a clinical flag and verify SSE endpoint delivers the notification in real time

Closes #401

### DIFF
--- a/services/notifications/src/__tests__/publish.test.ts
+++ b/services/notifications/src/__tests__/publish.test.ts
@@ -1,0 +1,124 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import { publishNotification, setPublisher } from "../publish.js";
+
+/**
+ * Tests for Redis pub/sub notification publishing.
+ *
+ * Verifies that notification creation correctly publishes to the expected
+ * Redis channel with the payload format the SSE endpoint consumes.
+ */
+
+function createMockRedis() {
+  return {
+    publish: vi.fn().mockResolvedValue(1),
+    quit: vi.fn().mockResolvedValue("OK"),
+  };
+}
+
+describe("publishNotification", () => {
+  let mockRedis: ReturnType<typeof createMockRedis>;
+
+  beforeEach(() => {
+    mockRedis = createMockRedis();
+    setPublisher(mockRedis as never);
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it("publishes to the correct Redis channel for a user", async () => {
+    const userId = "user-abc-123";
+    const payload = {
+      id: "notif-001",
+      type: "ai-flag",
+      title: "CRITICAL: Clinical flag — drug interaction",
+      body: "Potential interaction between warfarin and aspirin",
+      link: "/patients?flagId=flag-001",
+      related_flag_id: "flag-001",
+      created_at: "2026-04-12T10:00:00.000Z",
+    };
+
+    await publishNotification(userId, payload);
+
+    expect(mockRedis.publish).toHaveBeenCalledTimes(1);
+    expect(mockRedis.publish).toHaveBeenCalledWith(
+      `notifications:${userId}`,
+      JSON.stringify(payload),
+    );
+  });
+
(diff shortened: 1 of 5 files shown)


blamechris (commented):
## Code Review — PR #409

### Verdict: Approve with Required Fixes

The core fix is correct: the `lazyConnect: true` / never-connected publisher was the root cause of silent SSE delivery failure. Extracting a shared `publish.ts` module and wiring it into both the BullMQ worker and the tRPC `create` mutation is the right approach. Tests are well-structured.

---

### Findings by Severity

#### HIGH — Publish failures break mutation & cause duplicate rows

**router.ts (line ~63):** If Redis is down, `publishNotification()` throws after the DB insert already succeeded. The caller sees an error, but the notification row exists. Any retry creates a duplicate.

**dispatch-worker.ts (line ~163):** Same pattern — Redis publish failure after batch insert triggers a BullMQ retry, which re-inserts all notification rows with new UUIDs.

**Fix:** Wrap publish calls in try/catch in both locations. Publish is best-effort; the SSE path is a bonus, and clients fall back to polling. (Copilot flagged this too — I agree with their suggestion.)

#### MEDIUM — Redis config duplication

`publish.ts` manually constructs `new Redis({ host, port, password, tls })` from env vars. This duplicates the logic in `@carebridge/redis-config` (`getRedisConnection()`). The shared package is the canonical source — use it to avoid drift. (Note: `ioredis` accepts the same shape, so `new Redis(getRedisConnection())` works.)

#### MEDIUM — Public re-export of internal test hooks

`index.ts` now re-exports `getPublisher` and `setPublisher`. These are test-only internals that let external consumers mutate a module-level singleton. Only `publishNotification` and `NotificationPayload` should be in the public API.

#### LOW — Dead code in api-gateway

`services/api-gateway/src/routes/notifications-sse.ts` exports `publishNotificationToUser()` (lines 85-100), which is now superseded by the new `publishNotification()` in `publish.ts`. Should be removed to avoid confusion.

#### LOW — No graceful shutdown for publisher client

The `getPublisher()` singleton has no teardown path. When the service shuts down, the Redis connection is leaked. Consider exposing a `closePublisher()` and calling it in the server's graceful shutdown handler.

---

### Tests

The 5 new tests cover the key paths well: channel routing, payload format, optional fields, empty userId guard, and multi-user dispatch. Good use of `setPublisher` for mock injection.

### Summary

Two fixes should be made before merge (the HIGH items). The MEDIUM and LOW items can be deferred to follow-up issues.

PR #452, opened 2026-04-13
fix(db): encrypt patient diagnosis and notes at rest

### LINKED ISSUE #287: patients.diagnosis and patients.notes stored unencrypted
**File:** packages/db-schema/src/schema/patients.ts:14-15

Other patient PII columns (name, dob, mrn, insurance) use encryptedText() but diagnosis and notes — core PHI — are plain text.

**Fix:** encryptedText() + migration.

blamechris:
## Summary
- Switch `patients.diagnosis` and `patients.notes` from plain `text()` to `encryptedText()` so these PHI fields are encrypted at rest with AES-256-GCM, consistent with all other PHI columns on the patients table (`name`, `mrn`, `insurance_id`, etc.)
- Add migration `0025_encrypt_patient_diagnosis_notes.sql` and journal entry (no DDL change needed since the column type remains `text` at the SQL level; encryption is handled transparently by the Drizzle custom type)
- Existing plaintext rows will need a one-time re-encryption pass before reads will succeed through the ORM (documented in the migration header)

Closes #287

## Test plan
- [x] TypeScript compiles cleanly (`tsc --noEmit`)
- [x] All 272 existing tests pass (no regressions)
- [ ] Verify on dev DB: new writes to `patients.diagnosis`/`patients.notes` produce `iv:authTag:ct` ciphertext
- [ ] Run re-encryption script against any environment with existing plaintext data before deploying

### DIFF
--- a/packages/db-schema/drizzle/0028_encrypt_patient_diagnosis_notes.sql
+++ b/packages/db-schema/drizzle/0028_encrypt_patient_diagnosis_notes.sql
@@ -1,0 +1,26 @@
+-- 0025_encrypt_patient_diagnosis_notes.sql
+--
+-- HIPAA: encrypt patients.diagnosis and patients.notes at rest
+-- using AES-256-GCM via the `encryptedText` Drizzle custom type
+-- (see packages/db-schema/src/encryption.ts).
+--
+-- Both columns are already `text` at the SQL level, and the
+-- `encryptedText` custom type also maps to `text`, so no DDL
+-- change is required. The Drizzle ORM layer now transparently
+-- encrypts on write and decrypts on read.
+--
+-- Fields now encrypted:
+--   patients.diagnosis
+--   patients.notes
+--
+-- IMPORTANT — DATA MIGRATION REQUIRED:
+-- Any existing rows with plaintext values in these columns will
+-- fail to decrypt through the custom type (which expects the
+-- `iv:authTag:ct` format produced by encrypt()). A one-time
+-- re-encryption script must run against the live database to
+-- rewrite every existing row through the encryption pipeline
+-- before application reads will succeed.
+
+-- No-op: column type unchanged (text -> text); encryption is
+-- handled at the ORM layer by the encryptedText custom type.
+SELECT 1;
--- a/packages/db-schema/drizzle/meta/_journal.json
+++ b/packages/db-schema/drizzle/meta/_journal.json
@@ -190,6 +190,13 @@
       "when": 1745942400000,
       "tag": "0027_flag_escalation_tracking",
       "breakpoints": true
+    },
+    {
+      "idx": 27,
+      "version": "7",
+      "when": 1746028800000,
+      "tag": "0028_encrypt_patient_diagnosis_notes",
+      "breakpoints": true
     }
   ]
 }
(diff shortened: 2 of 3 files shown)


blamechris:
## Code Review: PR #452

### Summary
Switches `patients.diagnosis` and `patients.notes` from plain `text()` to `encryptedText()` for at-rest AES-256-GCM encryption, adds migration `0025_encrypt_patient_diagnosis_notes.sql` with a journal entry, and documents the need for a one-time re-encryption pass on existing plaintext rows. Closes #287.

### Strengths
- Change is minimal, consistent with the pattern established in `0008_encrypt_patient_name.sql`, `0011_encrypt_clinical_narratives.sql`, and `0016_notification_phi_encryption.sql` (no DDL needed — the `encryptedText` customType is still `text` at the SQL layer).
- Migration header clearly calls out the one-time re-encryption requirement and retains the `iv:authTag:ct` invariant.
- Shared-type surface (`Patient.diagnosis?: string`) is unaffected — encryption is transparent to callers.
- Journal entry `idx: 24` is the next free slot (existing `0024_notification_urgency.sql` is not indexed in the journal, a pre-existing quirk).

### Issues Found

#### Critical (Must Fix)
| packages/db-schema/src/encrypt-existing.ts | 18-24 | `PHI_COLUMNS` hardcoded list does not include `diagnosis` or `notes`. The PR migration header says existing plaintext rows need a re-encryption pass before application reads will succeed, but the existing script will silently skip these two columns. Any environment with existing patient data will break on read. | Add `"diagnosis"` and `"notes"` to `PHI_COLUMNS`. (Also missing `"name"` from PR #`0008` — add for completeness.) |
| packages/db-schema/src/rotate-keys.ts | 17-23 | Same `PHI_COLUMNS` list is used on key-rotation; missing `diagnosis` and `notes` (and `name`). On the next key rotation, these columns will not be re-encrypted with the new key and will become undecryptable once the previous key is retired. | Add `"name"`, `"diagnosis"`, `"notes"` to `PHI_COLUMNS` so rotation covers every encrypted patient column. |

| packages/db-schema/drizzle/0025_encrypt_patient_diagnosis_notes.sql | 32-36 | Consider a single shared constant (e.g. `PATIENT_PHI_COLUMNS`) exported from `encryption.ts` and imported by both `encrypt-existing.ts` and `rotate-keys.ts`, so new encrypted columns only need to be declared in one place. | Prevents future recurrences of the missing-column bug. |
| PR description | — | Test plan says \"Run re-encryption script against any environment with existing plaintext data before deploying\" but does not call out which script — and the current script does not handle these columns. Clarify. | Operational clarity for release engineer. |

#### Nitpicks (Optional)
- Migration file is a no-op `SELECT 1;` — consistent with prior precedent in this repo (see 0008), so fine. A brief comment pointing at the re-encryption command (`tsx packages/db-schema/src/encrypt-existing.ts`) would save deployers a grep.

### Architecture Notes
Follows the established \"encrypt-at-ORM\" pattern: `encryptedText` is a Drizzle `customType` that encrypts on write and decrypts on read, so the SQL column stays `text` and downstream callers (router, validators, FHIR generator) don't need changes. The PR correctly keeps the shared-type shape unchanged.

The operational gap is that the repo's encryption infrastructure (`encrypt-existing.ts` + `rotate-keys.ts`) still tracks encrypted columns via a hardcoded list that has silently drifted from the schema. This PR exposes but does not cause that drift — however, shipping it without updating the list leaves both scripts broken for these two fields, so the fix belongs here.

### Verdict
- [ ] Approve - Ready to merge
- [x] Request Changes - Critical items above must be addressed before merge
- [ ] Comment - Feedback only

PR #463, opened 2026-04-13
feat(ai): add weight-based dose validation for high-risk medications

### LINKED ISSUE #229: No weight-based or renal/hepatic dose adjustment
**Files:** packages/medical-logic/src/medical-validation.ts:74-96; packages/db-schema/src/schema/clinical-data.ts:5-27

medications table has no patient weight, GFR, hepatic function, or age fields. No rules check creatinine/GFR before dosing. Vancomycin, aminoglycosides, metformin, ACE-i, lithium all need adjustment.

**Risk:** 85yo with GFR 35 prescribed Metformin 1000 BID → lactic acidosis. Most common cause of preventable elderly admissions.

**Fix:** Add weight_kg, eGFR, child_pugh to schema; create dosing-adjustments.ts ruleset; query latest labs at med order time.

blamechris:
## Summary

- Adds `checkWeightBasedDosing()` to `@carebridge/medical-logic` that validates medication doses against weight-adjusted maximums for acetaminophen (75 mg/kg/day, 4g absolute cap), ibuprofen (40 mg/kg/day), vancomycin (20 mg/kg per dose), and gentamicin (7 mg/kg per dose)
- Adds `weight_kg` nullable column to the `patients` table with migration `0025_patient_weight_kg.sql`
- When patient weight is not documented, returns an INFO-level alert recommending weight documentation rather than silently skipping the check
- Includes 17 unit tests covering safe doses, excess doses, missing/invalid weight, brand-name matching, and case-insensitive lookup

Closes #229

## Test plan

- [x] 17 vitest tests pass for all drug rules, edge cases, and missing-weight scenarios
- [ ] Verify migration applies cleanly against dev database (`pnpm db:migrate`)
- [ ] Integration: confirm ai-oversight worker can import and call `checkWeightBasedDosing` with patient context

### DIFF
--- a/packages/db-schema/drizzle/0030_patient_weight_kg.sql
+++ b/packages/db-schema/drizzle/0030_patient_weight_kg.sql
@@ -1,0 +1,7 @@
+-- Migration: Add weight_kg column to patients table
+--
+-- Supports weight-based dose validation in the AI oversight engine.
+-- Weight is stored in kilograms as a nullable real column — null indicates
+-- weight has not yet been documented for the patient.
+
+ALTER TABLE patients ADD COLUMN IF NOT EXISTS weight_kg REAL;
--- a/packages/db-schema/drizzle/meta/_journal.json
+++ b/packages/db-schema/drizzle/meta/_journal.json
@@ -204,6 +204,13 @@
       "when": 1746115200000,
       "tag": "0029_allergy_verification_status",
       "breakpoints": true
+    },
+    {
+      "idx": 29,
+      "version": "7",
+      "when": 1746201600000,
+      "tag": "0030_patient_weight_kg",
+      "breakpoints": true
     }
   ]
 }
--- a/packages/db-schema/src/schema/patients.ts
+++ b/packages/db-schema/src/schema/patients.ts
@@ -1,5 +1,5 @@
 import crypto from "node:crypto";
-import { pgTable, text, boolean, index } from "drizzle-orm/pg-core";
+import { pgTable, text, boolean, real, index } from "drizzle-orm/pg-core";
 import { encryptedText } from "../encryption.js";
 
 export const patients = pgTable("patients", {
@@ -22,9 +22,8 @@ export const patients = pgTable("patients", {
   emergency_contact_name: encryptedText("emergency_contact_name"),
   emergency_contact_phone: encryptedText("emergency_contact_phone"),
   primary_provider_id: text("primary_provider_id"),
-  // Overall patient allergy status: distinguishes "confirmed no allergies" (nkda)
-  // from "never assessed" (unknown) and "has documented allergies" (has_allergies).
-  allergy_status: text("allergy_status").notNull().default("unknown"), // nkda, unknown, has_allergies
+  allergy_status: text("allergy_status").notNull().default("unknown"),
+  weight_kg: real("weight_kg"),
   created_at: text("created_at").notNull(),
   updated_at: text("updated_at").notNull(),
 });
--- a/packages/medical-logic/src/__tests__/weight-based-dosing.test.ts
+++ b/packages/medical-logic/src/__tests__/weight-based-dosing.test.ts
@@ -1,0 +1,224 @@
+import { describe, it, expect } from "vitest";
+import { checkWeightBasedDosing } from "../weight-based-dosing.js";
+
+// ─── No matching drug ──────────────────────────────────────────
+
+describe("checkWeightBasedDosing — unrecognized drug", () => {
+  it("returns no alerts for a drug without weight-based rules", () => {
+    const alerts = checkWeightBasedDosing({
+      medicationName: "Amoxicillin",
+      doseMg: 500,
+      weightKg: 70,
+    });
+    expect(alerts).toHaveLength(0);
+  });
+});
+
+// ─── Missing weight ────────────────────────────────────────────
+
+describe("checkWeightBasedDosing — missing weight", () => {
+  it("returns INFO when weight is undefined", () => {
+    const alerts = checkWeightBasedDosing({
+      medicationName: "Acetaminophen",
+      doseMg: 1000,
+      weightKg: undefined,
+    });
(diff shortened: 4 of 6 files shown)


blamechris:
## Code Review: PR #463

### Summary
Adds `checkWeightBasedDosing()` in `@carebridge/medical-logic` covering acetaminophen, ibuprofen, vancomycin, and gentamicin. Adds `weight_kg` (nullable `real`) to the `patients` table via migration 0025, with 17 vitest cases. Pure function, no service-side wiring included.

### Strengths
- Clean functional module in the right package (`medical-logic`), no side effects — fits the pure-function convention for this package.
- Solid test coverage: safe/excess doses, missing/null/zero/negative weight, brand-name regex matching, case-insensitive lookup, both daily and per-dose modes.
- Sensible safety posture on missing weight: INFO alert rather than silent skip.
- Rule IDs follow the `DOMAIN-CONDITION-SEQ` pattern (`DOSE-WT-APAP-001`, etc.).
- Clear inline citations for each drug's dosing threshold.

### Issues Found

#### Critical (Must Fix)

| `packages/db-schema/drizzle/0025_patient_weight_kg.sql` + `meta/_journal.json` | — | **Migration numbering collision.** `main` already contains `0025_failed_clinical_events.sql` (journal idx=24). This PR adds another entry at idx=24 with tag `0025_patient_weight_kg`. After merge the journal will have two idx=24 entries and two `0025_*` filenames; Drizzle ordering is undefined and applies may skip one. | Rename migration file to `0026_patient_weight_kg.sql` and update journal entry `idx: 25`, `tag: "0026_patient_weight_kg"`. |
| `packages/medical-logic/src/weight-based-dosing.ts` (full file) | — | **Severity type deviates from project convention.** The rules engine and shared types use lowercase `FlagSeverity = "critical" \| "warning" \| "info"` (`packages/shared-types/src/ai-flags.ts:6`, confirmed in `drug-interactions.ts`, `critical-values.ts`). This module introduces uppercase `DoseAlertSeverity = "WARNING" \| "INFO"`, which cannot feed a `ClinicalFlag.severity` directly and will force a string-case shim at every call site. | Import `FlagSeverity` from `@carebridge/shared-types` and use lowercase `"warning"` / `"info"`. Delete the local `DoseAlertSeverity` type. |
| `packages/medical-logic/src/weight-based-dosing.ts` (early-return when `weightKg == null`) | ~95 | **Absolute acetaminophen ceiling is skipped when weight is undocumented.** Current code returns the INFO alert and exits, so a 10 g/day acetaminophen order on a patient without weight on file produces only an INFO "please document weight" — hepatotoxic doses slip through silently. The absolute 4 g/day cap is weight-independent and must still fire. | When `weightKg == null`, still evaluate `absoluteMaxMg` for `mode === "daily"` drugs and emit a WARNING if exceeded; emit the INFO only alongside (or in place of) the weight-based portion. |

| `packages/medical-logic/src/weight-based-dosing.ts` | `WeightBasedDoseInput.doseMg` | Input is named `doseMg` but DB stores `dose_amount` + `dose_unit` (`packages/db-schema/src/schema/clinical-data.ts:10-11`), where unit can be `mg`, `mcg`, `g`, `mL`, etc. Callers who pass a `g` or `mcg` value directly will silently produce wrong results (e.g., "1" gram interpreted as 1 mg = no alert). | Accept `doseAmount` + `doseUnit` and normalize to mg internally (reject or warn for non-mass units like mL). Mirrors `validateMedicationDose` in `medical-validation.ts`. |
| `packages/medical-logic/src/weight-based-dosing.ts` | `DRUG_RULES` name patterns | `/gentamicin/i` and `/vancomycin/i` match topical/ophthalmic formulations where systemic weight-based dosing does not apply (e.g., gentamicin eye drops, vancomycin oral for C. diff). | Either scope the rule to IV/parenteral routes (pass a `route` parameter) or document the limitation and require callers to gate by route. |
| `packages/db-schema/src/schema/patients.ts` | `weight_kg: real("weight_kg")` | `real` is 4-byte float (~7 significant digits). For clinical weights this is acceptable, but the project uses `text`/`encryptedText` for most patient fields. Weight is PHI. | Consider `numeric(6,2)` for precision and whether weight should be `encryptedText` like other patient attributes. At minimum document the typing choice inline. |
| `packages/medical-logic/src/weight-based-dosing.ts` | `dosesPerDay?: number` default of 1 | Defaulting to 1 dose/day for a PRN acetaminophen order (`doseMg: 1000`) produces no alert, but a nurse administering `q4h PRN` could hit 6 g/day. This is a silent under-check. | Make `dosesPerDay` required for `mode: "daily"` drugs (TypeScript narrowing), or treat missing `dosesPerDay` as cause for an INFO alert like the missing-weight case. |

#### Nitpicks (Optional)
- `ruleId` is repeated on alerts that have the same rule; consider a synthetic suffix `...-ABSOLUTE` vs `...-WEIGHT` so downstream dedup can distinguish the two sub-checks.
- Box-drawing comment separators (`// ─── ... ──`) are stylistic; the rest of the codebase uses plain `//` section comments.
- Export `DRUG_RULES` read-only so other modules (e.g., an admin UI listing supported drugs) can inspect the list.

Will create post-review for the dose-unit normalization and route-gating suggestions; these depend on decisions about the `medications` caller shape and are out of scope for this PR.

### Architecture Notes
The module belongs in `@carebridge/medical-logic` (pure, no side effects) — correct placement. However, the PR does not wire the function into the ai-oversight worker (no changes under `services/ai-oversight/src/rules/`). The PR summary acknowledges integration is a separate unchecked item. That is fine to defer, but until wiring lands, this code is dead. Please confirm a follow-up is tracked.

The `patients.weight_kg` column is needed by `buildPatientContext` (`services/ai-oversight/src/workers/context-builder.ts`) to supply weight into this check. Not selected/propagated in this PR — expected in the integration PR.

### Verdict
- [x] **Request Changes** — Migration collision and severity case mismatch are blocking; acetaminophen-without-weight absolute-cap gap is a safety concern.

PR #465, opened 2026-04-13
feat(ai): add age-stratified pediatric vital sign ranges

### LINKED ISSUE #243: No pediatric vital sign ranges — adult norms applied to all ages
**Files:** packages/medical-logic/src/medical-validation.ts:16-25; vitals-trend-chart.tsx:36-52

3yo HR 150 (critical for child) treated as normal (threshold 200). Neonate SBP 60 (normal) treated as low. Massive false negatives + false positive alert fatigue.

**Fix:** Age-stratified ranges per PALS/AHA (neonate/infant/toddler/child/adolescent); require age in validateVital().

blamechris:
## Summary
- Add pediatric vital sign ranges for 5 age groups (neonate, infant, child, school-age, adolescent) covering heart rate, respiratory rate, and systolic blood pressure
- Export `getVitalRangeForAge()`, `classifyAgeGroup()`, and `ageInYearsFromDOB()` from `@carebridge/medical-logic`
- Update `isCriticalVital()` to accept optional `ageYears` parameter, falling back to adult thresholds when age is unknown
- Wire patient DOB (`patient_date_of_birth`) through the critical-values rule so pediatric thresholds apply automatically

## Test plan
- [x] 20 new unit tests covering all age groups, age classification, DOB parsing, and range selection
- [x] All 52 medical-validation tests pass
- [x] All 54 ai-oversight rules tests pass (mock updated for new exports)
- [ ] Manual verification: create a toddler patient and record HR 150 -- should not flag as critical

Closes #243

### DIFF
--- a/packages/medical-logic/src/__tests__/medical-validation.test.ts
+++ b/packages/medical-logic/src/__tests__/medical-validation.test.ts
@@ -6,6 +6,10 @@ import {
   isCriticalVital,
   getVitalSeverity,
   VITAL_DANGER_ZONES,
+  PEDIATRIC_VITAL_RANGES,
+  classifyAgeGroup,
+  ageInYearsFromDOB,
+  getVitalRangeForAge,
 } from "../medical-validation.js";
 
 // ─── validateVital ──────────────────────────────────────────────
@@ -197,6 +201,38 @@ describe("isCriticalVital", () => {
   it("returns false for glucose 300 (above warningHigh but below criticalHigh)", () => {
     expect(isCriticalVital("blood_glucose", 300)).toBe(false);
   });
+
+  it("uses adult thresholds when no age is provided", () => {
+    expect(isCriticalVital("heart_rate", 150)).toBe(false);
+  });
+
+  it("flags HR 150 as critical for a toddler (age 2)", () => {
+    expect(isCriticalVital("heart_rate", 150, 2)).toBe(true);
+  });
+
+  it("does not flag HR 120 as critical for a toddler (age 2)", () => {
+    expect(isCriticalVital("heart_rate", 120, 2)).toBe(false);
+  });
+
+  it("flags HR 165 as critical for a neonate (age 0.01)", () => {
+    expect(isCriticalVital("heart_rate", 165, 0.01)).toBe(true);
+  });
+
+  it("does not flag HR 140 as critical for a neonate (age 0.01)", () => {
+    expect(isCriticalVital("heart_rate", 140, 0.01)).toBe(false);
+  });
+
+  it("flags RR 55 as critical for an infant (age 0.5)", () => {
+    expect(isCriticalVital("respiratory_rate", 55, 0.5)).toBe(true);
+  });
+
+  it("flags low SBP for school-age child", () => {
+    expect(isCriticalVital("blood_pressure", 80, 8)).toBe(true);
+  });
+
+  it("uses adult thresholds for age 25", () => {
+    expect(isCriticalVital("heart_rate", 150, 25)).toBe(false);
+  });
 });
 
 // ─── getVitalSeverity ──────────────────────────────────────────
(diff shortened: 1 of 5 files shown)


blamechris:
## Code Review: PR #465 — add age-stratified pediatric vital sign ranges

### Summary
Adds `PEDIATRIC_VITAL_RANGES` for five age groups (neonate / infant / child / school-age / adolescent), plus helpers `classifyAgeGroup`, `ageInYearsFromDOB`, `getVitalRangeForAge`, and threads an optional `ageYears` through `isCriticalVital`. The `critical-values` rule now pulls `event.data.patient_date_of_birth` and uses it to pick an age-appropriate threshold.

### Strengths
- Pure-function additions live in `@carebridge/medical-logic` where they belong; no runtime deps leak into `shared-types`.
- Backward-compatible API (`ageYears` optional, adult fallback preserved) — no churn for existing callers.
- Strong unit coverage: 20 new tests span each age group, DOB parsing edge cases, and the adult fallback.
- Good failure modes in `ageInYearsFromDOB` — returns `undefined` for missing / unparseable / future DOB rather than throwing.
- Vitest alias for `@carebridge/shared-types` is a pragmatic fix for the monorepo type-only import.

### Issues Found

#### Critical (Must Fix)

| `services/clinical-data/src/repositories/vital-repo.ts` | 31–37 | **Pediatric thresholds never engage in production.** `vital-repo.ts` emits `{ resourceId, vitalType, value }`. The new rule code reads `event.data.patient_date_of_birth` (as well as `type`/`value_primary`), none of which exist on the emitted payload. `checkCriticalValues(event)` runs on the raw event *before* any DB enrichment (see `services/ai-oversight/src/services/review-service.ts:83`), so `patientAgeYears` will always be `undefined` in the deployed pipeline. Unit tests pass because they hand-craft the field. Result: the feature is a no-op — the DVT toddler scenario in the PR body (HR 150 on a 2-yo) still flags as non-critical in production. | Either (a) enrich the event in `createVital`/`updateVital` by joining `patients.date_of_birth` before emitting, OR (b) look up DOB inside `checkCriticalValues` (async + DB call — changes the sync contract), OR (c) move the age lookup into `review-service.ts` so it can pass it alongside the event. Option (a) is closest to the existing pattern and keeps the rule synchronous. Regardless, add an integration test that emits through `createVital` → event bus → `checkCriticalValues` and asserts a toddler HR 150 is flagged. |
| `packages/medical-logic/src/medical-validation.ts` | 55–73 | **Several pediatric thresholds are clinically off and will produce severe alert fatigue or miss critical deltas.** • Adolescent `heart_rate.criticalHigh: 100` — 100 is the upper end of *normal* adolescent HR (PALS 60–100); every sinus tachycardia (exercise, fever, anxiety) fires a critical flag. • Adolescent `blood_pressure.criticalHigh: 140` — 140 SBP is stage-1 HTN, not critical. • Child `heart_rate.criticalHigh: 130` — PALS normal awake child is 80–130; equality trigger (`>= 130`) fires at the exact normal boundary. • Neonate `blood_pressure.criticalHigh: 90` — neonate SBP 60–90 is normal. • Neonate `heart_rate.criticalHigh: 160` — 120–160 is normal resting neonate HR; active/crying neonates commonly hit 180. | Re-derive from AHA/PALS reference tables with a clinician sign-off. Suggested direction: align high thresholds with PALS "tachycardia requiring assessment" (e.g., child > 160, adolescent > 130) and low thresholds with "bradycardia requiring CPR consideration" (e.g., infant < 80, child < 60). Add the source citation in a comment above `PEDIATRIC_VITAL_RANGES`. |

| `packages/medical-logic/src/medical-validation.ts` | 95–108 | `classifyAgeGroup(ageYears)` returns `"adult"` for negative input. That masks a bug (negative age ⇒ silent adult fallback) rather than surfacing it. | Return `undefined` or throw on negative input; let `getVitalRangeForAge` handle the fallback explicitly. |
| `packages/medical-logic/src/medical-validation.ts` | 41–65 | Pediatric ranges cover only HR / RR / SBP. `o2_sat`, `temperature`, and `blood_glucose` fall through to adult thresholds, which is mostly fine, but the hypoglycemia threshold for a neonate (< 40 mg/dL, vs. adult 54) is clinically different. | Add neonatal hypoglycemia threshold, or document the gap explicitly and open a follow-up issue (done — see below). |
| `packages/medical-logic/src/medical-validation.ts` | 110–120 | `ageInYearsFromDOB` uses `new Date(dateOfBirth)` — for a bare `YYYY-MM-DD` string this parses as UTC midnight, and `new Date()` is local time. A patient born 23 h ago in the user's timezone could yield a negative diff → `undefined` in the edge case. | Either require an ISO 8601 with Z, or clamp `diffMs < 0` to `0` rather than returning `undefined`, or use UTC-normalized dates. Low risk in practice but worth documenting. |
| `services/ai-oversight/src/__tests__/rules.test.ts` | 11–54 | The test mock replicates the real implementation. It passes today but will drift; the whole reason the test doesn't catch the production wiring bug is that the mock returns `VITAL_DANGER_ZONES[vitalType]` for `getVitalRangeForAge` regardless of age. | Add at least one integration-style test (no mock of `@carebridge/medical-logic`) that feeds an event with `patient_date_of_birth` and asserts a pediatric-specific threshold fires. |

#### Nitpicks (Optional)
- `PEDIATRIC_VITAL_RANGES` could be `Readonly<...>` or `as const`-typed to prevent accidental mutation.
- The comment on line 95 says "neonate: 0 to ~0.077 years (28 days)" — cleaner to name the constant `NEONATE_MAX_YEARS = 28 / 365.25`.
- Consider exporting `AgeGroup` from the package index so consumers don't have to deep-import.

See the Issues column in the summary below — the glucose/temperature gap and the `buildEventSummary` field-name drift will be tracked separately.

### Architecture Notes
The feature concept is right — age-stratified thresholds belong in `medical-logic` as pure functions, and `isCriticalVital` is the right integration point. The gap is purely in data flow: the event bus payload is authoritative for deterministic rules, and the emitters never learned about DOB. This mirrors a pre-existing field-name mismatch (`vitalType` vs `type`, `value` vs `value_primary`) in `vital-repo.ts` that the rule happens to compensate for by reading a field that doesn't exist — a latent bug this PR makes more visible.

### Verdict
**Request Changes** — the pediatric safety feature cannot ship until the DOB is actually delivered to the rule. The clinical threshold values also need review by a clinician before they go live; shipping the current numbers would likely cause a wave of false critical flags on normal adolescents.

PR #474, opened 2026-04-14
fix(ai): correct creatinine/BUN trend direction for AKI detection

### LINKED ISSUE #213: clinical-safety(p2): creatinine and BUN trend direction wrong — rising AKI trends shown as stable
## Problem
`packages/medical-logic/src/trend-utils.ts` marks both Creatinine and BUN as `"stable"` direction:

```typescript
Creatinine: "stable",
BUN: "stable",
```

**Clinically, both are directional:**
- **Rising creatinine** = worsening renal function, acute kidney injury
- **Rising BUN** = kidney dysfunction, dehydration, catabolic state
- **Falling creatinine/BUN** = improving renal function (good)

A patient with creatinine rising from 1.0 → 1.8 mg/dL in 24 hours (80% increase = AKI criteria) will have a **neutral trend indicator** — no visual or alert-level signal that renal function is deteriorating rapidly.

## Clinical Consequence
- Acute kidney injury goes unnoticed due to "stable" trend classification
- Particularly dangerous post-surgery, post-contrast, or with nephrotoxic drugs (aminoglycosides, NSAIDs, contrast)
- Rising BUN + rising Cr = AKI; currently shown as doubly "stable"

## Fix
```typescript
Creatinine: "down",  // lower is better
BUN: "down",         // lower is better
```
Then the trend analysis will correctly flag upward trends as "concerning" and downward as "improving."

blamechris:
## Summary

Fixes #213 — creatinine and BUN trends were shown as "stable" even when rising, hiding AKI signals from the oversight layer.

Root cause: `LAB_DIRECTIONS` in `packages/medical-logic/src/trend-utils.ts` mapped `Creatinine` and `BUN` to good-direction `"stable"`, which made `getTrendColor` always return `"neutral"` regardless of the change. Functionally this meant a rising kidney-function signal (0.7 → 1.0 → 1.4 mg/dL) rendered the same as a flat series.

## Changes

- Reclassify `Creatinine` and `BUN` as `"down"` (down-is-good), so rising values surface as negative trends. Matches existing handling of ALT/AST.
- Add `classifyTrend(values, tolerance?)` returning `rising` / `falling` / `stable` based on endpoint delta with a tolerance knob — direction independent of the good/bad coloring.
- Add `detectAKI(readings)` implementing KDIGO AKI criteria:
  - Absolute rise >= 0.3 mg/dL within 48 hours, OR
  - Relative rise >= 1.5x baseline within the prior 7 days
  - Returns the triggering criterion, baseline, peak, and KDIGO stage (1/2/3).
- Tests cover the realistic AKI trajectory from the issue (0.7 → 1.0 → 1.4), noisy-but-rising series, stage transitions at 1.5x / 2.0x / 3.0x, and reading-order independence.

## Test plan

- [x] `pnpm --filter @carebridge/medical-logic test` — 80 pass (51 in trend-utils, 29 pre-existing medical-validation).
- [x] `pnpm --filter @carebridge/medical-logic typecheck` — clean.
- [ ] Verify clinician portal lab trend tiles render rising creatinine in the negative/red color (manual QA).
- [ ] Confirm ai-oversight AKI-adjacent rules consume `detectAKI` in a follow-up PR if desired; this PR only adds the helper and fixes the classification.

Closes #213

### DIFF
--- a/packages/medical-logic/src/__tests__/trend-utils.test.ts
+++ b/packages/medical-logic/src/__tests__/trend-utils.test.ts
@@ -1,6 +1,8 @@
 import { describe, it, expect } from "vitest";
 import {
   calculateDelta,
+  classifyTrend,
+  detectAKI,
   getGoodDirection,
   getTrendColor,
   getTrendInfo,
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #474

### Summary
Corrects a patient-safety bug in `packages/medical-logic/src/trend-utils.ts` where `LAB_DIRECTIONS` mapped Creatinine and BUN to `"stable"`, causing rising AKI-signal trends to render as neutral. PR reclassifies both to `"down"` (matching ALT/AST), adds a direction-agnostic `classifyTrend`, and a KDIGO-based `detectAKI` helper. Tests cover the realistic 0.7 -> 1.0 -> 1.4 AKI trajectory and KDIGO stage transitions.

### Strengths
- Root-cause fix aligned with issue #213 and clinically correct — rising creatinine now surfaces as a negative trend.
- Pure-function contract for `@carebridge/medical-logic` is preserved (no side effects, no I/O).
- Tests cover order-independence, noisy-but-rising series, stage transitions at 1.5x / 2.0x / 3.0x, and distinguish the 48h-absolute vs. 7d-relative criterion paths.
- Good `baseline > 0` guard against divide-by-zero in the relative-rise branch.
- KDIGO stage 3 combines the 3.0x ratio rule and the `>= 4.0 mg/dL with acute rise >= 0.5` criterion correctly.

### Issues Found

#### Critical (Must Fix)

| `packages/medical-logic/src/trend-utils.ts` | 263-271 | `detectAKI` criterion #1 triggers on any `peak - min >= 0.3` within the 48h window regardless of time order. This false-positives on resolving AKI (e.g. 1.4 -> 1.0 -> 0.7): the peak precedes the min, the patient is improving, but the helper returns `isAKI: true`. KDIGO defines AKI as a *rise*, so peak must occur after baseline. | Anchor the comparison to the latest reading: compare `latest.value - minPriorTo48h` (min among readings strictly before `latest`). Alternatively, compute the running min up to each point and flag only if a subsequent peak exceeds it by >= 0.3. Same fix applies conceptually to the 7d relative branch. |

| `packages/medical-logic/src/trend-utils.ts` | 166-172 | `classifyTrend` JSDoc claims "Uses both the endpoint delta and a monotonicity check" but implementation only checks first vs. last. Update docstring to match (endpoint-delta only) or implement monotonicity. | Misleading docstring invites callers to rely on noise handling that does not exist. Docstring fix is the lower-risk option. |
| `packages/medical-logic/src/trend-utils.ts` | 247-251 | `detectAKI` uses `Date.parse()` without validating the result. A reading with a malformed timestamp yields `NaN`, which silently distorts sort order and window filters (worst case: AKI missed). | Filter invalid timestamps up front, or assert via Zod at the boundary. Cheap defense given this function feeds the oversight layer. |

#### Nitpicks (Optional)
- `peakIn48h` is computed but unused outside the stage call — fine as written, but could name the variables `baselineIn48h` / `peakIn48h` consistently with the 7d branch.
- `AKIResult.stage` type is `AKIStage | null`; when `isAKI: true`, stage is always non-null. A discriminated union (`{ isAKI: true; stage: AKIStage; ... } | { isAKI: false; ... }`) would let callers skip the null check.

None — the issues above are small enough to fix in this PR.

### Architecture Notes
Cleanly lives in `@carebridge/medical-logic` as pure functions. No service imports or side effects introduced. The PR body correctly flags that ai-oversight rules can adopt `detectAKI` in a follow-up — that is the right sequencing so this PR stays small.

### Verdict
Request Changes — the criterion-1 time-order bug can produce false-positive AKI on resolving patients, which is the opposite of the harm #213 targets. Docstring + timestamp-validation fixes should accompany it. Once those land, this is ready to merge.

PR #492, opened 2026-04-16
fix(ai): dedup flags by trigger_event_id to prevent replay duplicates

### LINKED ISSUE #261: Flag dedup ignores trigger_event_id — duplicate events create duplicate flags
**File:** services/ai-oversight/src/services/flag-service.ts:32-72

LLM dedup is by category+severity within 24h. Replayed event creates separate review_job → second flag with new ID.

**Fix:** Include trigger_event_id in dedup key.

blamechris:
## Summary
- Adds a replay-safety idempotency check at the top of \`createFlag\`: if any existing flag for the same patient (regardless of status) already records one of the incoming \`trigger_event_ids\`, return that flag.
- Uses Postgres jsonb \`@>\` containment per event id, OR'd across them, so the check is correct when a flag cites multiple events.

## Why
The existing \`(patient_id, rule_id)\` and \`(patient_id, category, severity)\` dedup only fires when the prior flag is still \`status='open'\`. When a previously resolved/acknowledged flag's trigger event is replayed (outbox reconciler, BullMQ retry, manual requeue), the partial-unique indexes don't match and a duplicate flag is created for the same clinical event — polluting the clinician inbox and the audit trail.

## Test plan
- [x] New test: flag returned when a trigger_event_id matches regardless of status
- [x] New test: idempotency skipped when trigger_event_ids is empty
- [x] Existing TOCTOU tests updated to account for the extra pre-check SELECT
- [x] \`pnpm --filter @carebridge/ai-oversight test\` — 280/280 passing

Closes #261

### DIFF
--- a/services/ai-oversight/src/__tests__/flag-service.test.ts
+++ b/services/ai-oversight/src/__tests__/flag-service.test.ts
@@ -24,6 +24,7 @@ vi.mock("@carebridge/db-schema", () => ({
     category: "category",
     severity: "severity",
     created_at: "created_at",
+    trigger_event_ids: "trigger_event_ids",
   },
 }));
 
@@ -161,10 +162,11 @@ describe("createFlag", () => {
       ...baseFlag,
       created_at: "2026-01-01T00:00:00.000Z",
     };
-    // First call: dedup SELECT -> [] (no dup found at app level)
-    // Second call (after conflict): fetch winner -> [winnerFlag]
+    // Call sequence: (1) trigger_event_id idempotency check -> [],
+    // (2) app-level rule dedup -> [], (3) post-conflict fetch -> [winnerFlag]
     mockLimit
-      .mockResolvedValueOnce([])       // initial dedup check
+      .mockResolvedValueOnce([])           // idempotency pre-check
+      .mockResolvedValueOnce([])           // initial dedup check
       .mockResolvedValueOnce([winnerFlag]); // post-conflict fetch
 
     const result = await createFlag(baseFlag);
@@ -196,8 +198,10 @@ describe("createFlag", () => {
 
     // ON CONFLICT DO NOTHING returns empty
     mockReturning.mockResolvedValueOnce([]);
-    // First SELECT (app-level dedup) -> no match; second SELECT (post-conflict) -> winner
+    // (1) idempotency pre-check -> []; (2) app-level LLM dedup -> [];
+    // (3) post-conflict fetch -> [winnerFlag]
     mockLimit
+      .mockResolvedValueOnce([])
       .mockResolvedValueOnce([])
       .mockResolvedValueOnce([winnerFlag]);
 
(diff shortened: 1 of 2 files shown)


blamechris:
## Agent Review

### Summary
Small, targeted fix that closes a real replay hole: the existing
`(patient, rule_id)` and `(patient, category, severity)` partial-unique
indexes only constrain `status='open'` flags, so a retry/reconciler replay
of an event whose prior flag was already resolved/acknowledged would
duplicate the flag. Using `trigger_event_ids` as an idempotency key is
the right primitive — it binds to the causing event, not to flag shape.

### Strengths
- Idempotency key is the clinical event id, which is the invariant the
  replay paths (outbox, BullMQ retry, manual requeue) actually share.
- Check is correctly scoped by `patient_id`, so event-id collisions across
  patients cannot leak a flag between charts.
- Empty-array short-circuit is handled (`trigger_event_ids.length > 0`),
  preserving manual-flag callers that have no event id.
- Comment above the block explains *why* this exists — future readers
  won't delete it thinking the partial indexes already cover it.
- Notification side effect is correctly skipped (early return bypasses the
  `emitNotificationEvent` path), so replays won't re-notify clinicians.

### Issues

**Critical:** none.

**Suggestions:**
1. **No GIN index on `trigger_event_ids`.** The new query ORs `jsonb @>`
   containment across each incoming event id. Today `idx_flags_patient
   (patient_id, status)` narrows the scan to a single patient's flags,
   which keeps the cost bounded, but at production scale (long-lived
   patients, years of flags) this becomes a seq-scan per replay. Filed
   as follow-up issue #509 — add
   `CREATE INDEX CONCURRENTLY ... USING gin (trigger_event_ids
   jsonb_path_ops)` before real event volume hits the table.
2. **SQLi via `JSON.stringify([id])`:** reviewed and cleared. Event ids
   originate from internal BullMQ queue events (`event.id`), not from
   external input, and drizzle passes the stringified value as a bound
   parameter via `${...}::jsonb`, not as interpolated SQL text. No
   injection surface.

**Nitpicks:**
1. The returned flag from the idempotency branch is cast as
   `ClinicalFlag` but is not guaranteed to match the *incoming* flag's
   `(rule_id, category, severity)` — it's whatever existing flag first
   cited that event id. This is by design (replay-safety contract), but
   the docstring on `createFlag` still says the return shape matches the
   input for dedup; consider tightening the JSDoc to mention the event-id
   path explicitly.
2. The new query runs before the `if (flag.rule_id)` branch even for
   rule-based flags that already have a stricter dedup path. For the
   common fresh-event case this adds one extra SELECT per `createFlag`.
   Acceptable given the query is single-patient and indexed on
   `patient_id`, but worth measuring once #509 lands.

### Test coverage
- New test `returns existing flag when a trigger_event_id matches,
  regardless of status` exercises the resolved-flag replay branch —
  the exact regression this PR guards against.
- New test `does not short-circuit when trigger_event_ids is empty`
  covers the manual-flag path.
- Both TOCTOU tests correctly added a leading `mockLimit.mockResolvedValueOnce([])`
  for the new pre-check SELECT. The mock chain sequence in the comments
  matches the call order in the code.

### Deferred Items
- #509 — perf(db): GIN index on `trigger_event_ids` for replay-safety lookup

### Verdict
**APPROVE.** Correct fix, right primitive, clean tests, clear docstring.
Perf follow-up tracked in #509; not a blocker.

PR #493, opened 2026-04-16
fix(ai): severity-priority + alert on LLM flag truncation

### LINKED ISSUE #260: MAX_FLAGS=20 silently truncates LLM findings with INFO log only
**File:** packages/phi-sanitizer/src/llm-validator.ts:20-21,172-184

>20 LLM flags silently sliced; no severity-based prioritization before truncation; clinician never sees 21st+ finding.

**Fix:** Sort by severity (critical→warning→info) before truncate; raise cap; ALERT-level log on truncation.

blamechris:
## Summary
- Raise \`MAX_FLAGS\` 20 → 50.
- Sort findings by severity (critical > warning > info) before capping, so criticals never get dropped while info survives.
- Add structured \`TruncationInfo\` on the validation result (receivedCount, keptCount, droppedCount, droppedBySeverity); \`review-service\` logs \`console.error\` at ALERT level when it's set.

## Why
Previously overflow was silently sliced in arbitrary response order with only a warning string. Any critical finding beyond the 20th was dropped and the clinician never saw it.

## Test plan
- [x] New tests: cap at 50 w/ dropped-by-severity breakdown; all criticals survive truncation; warnings drop before criticals; no truncation under the cap
- [x] \`pnpm --filter @carebridge/phi-sanitizer test\` — 13/13 passing
- [x] \`pnpm --filter @carebridge/ai-oversight test\` — 278/278 passing

Closes #260

### DIFF
--- a/packages/phi-sanitizer/src/__tests__/llm-validator.test.ts
+++ b/packages/phi-sanitizer/src/__tests__/llm-validator.test.ts
@@ -63,15 +63,79 @@ describe("validateLLMResponse", () => {
     }
   });
 
-  it("caps flags at 20", () => {
-    const flags = Array.from({ length: 25 }, () => makeFlag());
+  it("caps flags at the configured MAX and records truncation detail", () => {
+    // 60 info-severity flags — above the 50 MAX_FLAGS cap.
+    const flags = Array.from({ length: 60 }, () =>
+      makeFlag({ severity: "info" }),
+    );
     const input = JSON.stringify(flags);
     const result = validateLLMResponse(input);
 
     expect(result.ok).toBe(true);
     if (result.ok) {
-      expect(result.flags).toHaveLength(20);
+      expect(result.flags).toHaveLength(50);
       expect(result.warnings.some((w) => w.includes("truncated"))).toBe(true);
+      expect(result.truncation).toBeDefined();
+      expect(result.truncation?.receivedCount).toBe(60);
+      expect(result.truncation?.keptCount).toBe(50);
+      expect(result.truncation?.droppedCount).toBe(10);
+      expect(result.truncation?.droppedBySeverity).toEqual({
+        critical: 0,
+        warning: 0,
+        info: 10,
+      });
+    }
+  });
+
+  it("preserves all critical flags under truncation", () => {
+    // 3 critical + 60 info (total 63, cap 50). Critical must all survive;
+    // info is dropped first.
+    const criticals = Array.from({ length: 3 }, (_, i) =>
+      makeFlag({ severity: "critical", summary: `crit-${i}` }),
+    );
+    const infos = Array.from({ length: 60 }, (_, i) =>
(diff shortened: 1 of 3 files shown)


blamechris:
## Review — PR #493

### Summary
Raises `MAX_FLAGS` 20→50, sorts LLM findings by severity (critical > warning > info) before truncation, adds structured `TruncationInfo` on the validator result, and emits an ALERT-level `console.error` in review-service when truncation occurs. Closes the signal-loss gap flagged in #260.

### Strengths
- **Severity-preserving truncation is the right fix.** Sort-then-slice guarantees criticals never get dropped until all warning + info are gone; tests cover the 3-criticals-survive and warning-dropped-before-critical cases explicitly.
- **Stable sort assumption is valid.** `Array.prototype.sort` is stable in V8/Node 12+, and severity values are pre-validated against `VALID_SEVERITIES` before the sort runs (llm-validator.ts:75-81), so `SEVERITY_RANK[f.severity]` can never be `undefined`. LLM's intra-bucket ordering is preserved.
- **No PHI in the ALERT log.** `patient_id` is routed through `redactPatientId`; `event.type` is an enum (`ClinicalEventType`), `event.id` is the event row UUID, `jobId` is internal. Matches the existing redaction pattern used for token-budget warnings earlier in the same function.
- **`TruncationInfo.droppedBySeverity` is type-safe.** The `Record<(typeof VALID_SEVERITIES)[number], number>` is initialized exhaustively to `{ critical: 0, warning: 0, info: 0 }` before the loop, so no key is ever `undefined`, and TS enforces adding a new severity literal there.
- **Cap raise 20→50 is deliberate.** Explicitly called out by #260; downstream per-finding `createFlag` + dedup has no rate-limit or cost coupling (LLM token spend is pre-truncation).
- **Test coverage is strong.** New tests cover cap, critical-preservation, warning-vs-critical overflow, and the no-truncation negative case.

### Issues
None blocking. Minor observations below — all deferrable.

### Deferred Items
1. **Unstructured alert.** `console.error` with a formatted string is hard for ops dashboards / log aggregators to parse as a counter or alert trigger. A structured log line (JSON) or a metrics counter (`ai_oversight.llm_findings.truncated{severity=critical}`) would make this actionable in Grafana / Datadog. Filed as a follow-up since no structured logger currently exists in this service.
2. **`SUSPICIOUS_FLAG_THRESHOLD = 15` is now stale.** Designed against the old 20-cap; at cap=50 it fires from 30% full, which is noisy for the new limit. Worth re-tuning (e.g. proportional: `Math.floor(MAX_FLAGS * 0.75)`).
3. **Conditional spread at the return.** `...(truncation ? { truncation } : {})` works but a plain `const result: ValidationSuccess = { ok: true, flags, warnings }; if (truncation) result.truncation = truncation; return result;` reads cleaner. Style nit only.

### Verdict
**Approve.** Clinically correct, type-safe, PHI-safe, and well-tested. Follow-up items filed to avoid scope creep.

PR #494, opened 2026-04-16
fix(ai): filter patient state to event-time snapshot in rule context

### LINKED ISSUE #258: TOCTOU race: deterministic rules read patient state at review time, not event time
**File:** services/ai-oversight/src/services/review-service.ts:76-81,255-291

If a med is discontinued between event emission and review, rules see stale state. False flags or missed flags.

**Fix:** Snapshot relevant patient state at event-emit time; embed snapshot_at in event payload.

blamechris:
## Summary
- Filter diagnoses/medications/allergies/labs in \`buildPatientContextForRules\` by \`event.timestamp\` instead of current DB state.
- Uses row-level timestamps (\`onset_date\`/\`resolved_date\`, \`started_at\`/\`ended_at\`, \`created_at\`) — not the denormalized \`status\` columns — so a status column lagging its timestamp column cannot flip a rule outcome.

## Why
The rule pipeline read current state but evaluated it against an event emitted seconds-to-minutes earlier. Races between emission and review caused:

- **Missed flags** — a med active at event time but discontinued before review runs was excluded by \`status='active'\`; drug-interaction/allergy-medication rules didn't fire.
- **False flags** — an allergy added after event emission appeared relevant to the triggering event.

## Test plan
- [x] 5 new unit tests covering: med discontinued before event (exclude), med discontinued after event (include — the TOCTOU case), allergy added after event (exclude), diagnosis resolved before event (exclude), future lab (exclude).
- [x] \`pnpm --filter @carebridge/ai-oversight test\` — 283/283 passing

Closes #258

### DIFF
--- a/services/ai-oversight/src/__tests__/build-patient-context.test.ts
+++ b/services/ai-oversight/src/__tests__/build-patient-context.test.ts
@@ -122,4 +122,113 @@ describe("buildPatientContextForRules — recent_labs wiring", () => {
 
     expect(ctx.recent_labs).toBeUndefined();
   });
+
+  it("excludes a medication discontinued before the event timestamp", async () => {
+    const eventAt = "2026-04-16T12:00:00.000Z";
+    const event: ClinicalEvent = { ...stubEvent, timestamp: eventAt };
+
+    diagnosesSelect.mockResolvedValue([]);
+    medicationsSelect.mockResolvedValue([
+      // Started 10 days before event, ended 1 day before event — inactive at event time.
+      {
+        name: "Warfarin",
+        status: "active", // status is stale (wasn't yet updated to discontinued)
+        started_at: "2026-04-06T00:00:00.000Z",
+        ended_at: "2026-04-15T00:00:00.000Z",
+        created_at: "2026-04-06T00:00:00.000Z",
+      },
+    ]);
+    labsSelect.mockResolvedValue([]);
+
+    const ctx = await buildPatientContextForRules("p-1", event);
+
+    expect(ctx.active_medications).not.toContain("Warfarin");
+  });
+
+  it("includes a medication that is discontinued AFTER the event timestamp", async () => {
+    const eventAt = "2026-04-16T12:00:00.000Z";
+    const event: ClinicalEvent = { ...stubEvent, timestamp: eventAt };
+
+    diagnosesSelect.mockResolvedValue([]);
+    medicationsSelect.mockResolvedValue([
+      // The TOCTOU scenario: med was active when event was emitted, then
+      // discontinued before the review ran. The rules MUST still see it.
+      {
+        name: "Apixaban",
+        status: "discontinued",
+        started_at: "2026-04-01T00:00:00.000Z",
+        ended_at: "2026-04-16T15:00:00.000Z",
+        created_at: "2026-04-01T00:00:00.000Z",
+      },
+    ]);
+    labsSelect.mockResolvedValue([]);
+
+    const ctx = await buildPatientContextForRules("p-1", event);
+
+    expect(ctx.active_medications).toContain("Apixaban");
+  });
+
+  it("excludes an allergy added after the event timestamp", async () => {
+    const eventAt = "2026-04-16T12:00:00.000Z";
+    const event: ClinicalEvent = { ...stubEvent, timestamp: eventAt };
+
+    diagnosesSelect.mockResolvedValue([]);
+    medicationsSelect.mockResolvedValue([]);
+    allergiesSelect.mockResolvedValue([
+      // Allergy recorded an hour AFTER the event — rule must not retroactively
+      // consider it when evaluating the triggering event.
+      {
+        allergen: "penicillin",
+        rxnorm_code: "7980",
+        severity: "severe",
+        reaction: "rash",
+        created_at: "2026-04-16T13:00:00.000Z",
+      },
+    ]);
+    labsSelect.mockResolvedValue([]);
+
+    const ctx = await buildPatientContextForRules("p-1", event);
+
+    expect(ctx.allergies).toHaveLength(0);
+  });
+
+  it("excludes a diagnosis resolved before the event timestamp", async () => {
+    const eventAt = "2026-04-16T12:00:00.000Z";
+    const event: ClinicalEvent = { ...stubEvent, timestamp: eventAt };
+
+    diagnosesSelect.mockResolvedValue([
+      {
+        description: "Strep throat (resolved)",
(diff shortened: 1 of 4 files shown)


blamechris:
## Agent Review — PR #494

### Summary
TOCTOU fix for `buildPatientContextForRules`: patient state now filtered to a snapshot as of `event.timestamp` using row-level timestamps (`onset_date`/`resolved_date`, `started_at`/`ended_at`, `created_at`) rather than the denormalized `status` columns. Closes #258. Rules see the state that existed when the event was emitted, preventing the two documented race failures (missed flags when a med was discontinued before review; false flags when an allergy was added after the event).

### Strengths
- Correct root-cause framing: `status` is a denormalized projection that can lag its authoritative timestamps. Dropping `status` in favor of `started_at`/`ended_at`/`resolved_date` is the right move and also fixes latent bugs where a status update was never applied.
- Five tests cover the important directions including the genuine TOCTOU case (`med discontinued AFTER event, rules still see it`) — this is the clinically load-bearing path.
- Future-lab exclusion (`row.created_at > eventAt`) preserves determinism for rules like CHEMO-FEVER-001 (ANC-aware).
- Scope is tight: only the rule-path context builder is touched; `checkCriticalValues` (which uses the raw event, not patient context) and the symptom extraction are unaffected.
- Filter output is internally consistent: `active_diagnoses` and `active_diagnosis_codes` share the same `activeDx` source, so parallel arrays remain index-aligned after filtering (same for meds / rxnorm codes).

### Issues

**1. LLM review path has NOT been snapshot-filtered — rules and LLM now see different patient states (clinical consistency risk).**
`buildPatientContext` in `services/ai-oversight/src/workers/context-builder.ts` (lines ~70–80) still queries with `eq(diagnoses.status, "active")` and has no `event.timestamp` filtering. For the same trigger event the deterministic rules see the as-of-event snapshot, while the LLM sees current state. The LLM can produce flags (or suppress flags) based on a newer snapshot than the rules, defeating the dedup/precedence logic in the review pipeline (recently reinforced in #490). The PR description frames this as a complete TOCTOU fix; in practice it is a half-fix. Either extend the snapshot to `buildPatientContext` in this PR or document and track the gap explicitly.

**2. String comparison assumes normalized Z-suffixed ISO-8601 — not guaranteed for non-native timestamp fields.**
The comparison `row.resolved_date <= at` is lexicographic. It is correct iff both operands are strict ISO-8601 UTC Z-form (`YYYY-MM-DDTHH:MM:SS.sssZ`). Internal writes via clinical-data repos always use `new Date().toISOString()` (OK). But:
  - `onset_date` and `resolved_date` are user/clinician-supplied dates — seed data and the FHIR generator pass bare `YYYY-MM-DD` (e.g. tooling/seed/index.ts:122 uses `onset_date: "2025-12-01"`). A bare-date string and a Z-datetime string compare lexically OK only because `""` < `"T"` — relying on that is subtle and should be asserted, not assumed. If any importer or manual edit ever lands a string with a `+HH:MM` offset (FHIR Condition.onsetDateTime permits offsets), the comparison silently becomes wrong (a post-event onset with `-05:00` will lex-compare less than a Z-time with the same instant).
  - Recommendation: parse both sides with `Date.parse()` (or compare `.getTime()`), or enforce at the schema/write layer that all timestamp text columns be Z-form, and add a regression test for the bare-date and offset-form edge cases.

**3. `wasActiveAt` fallback to `created_at` when `onset_date`/`started_at` is null — masks schema drift.**
The comment acknowledges this is defensive. But the `medications` schema (packages/db-schema/src/schema/clinical-data.ts) allows `started_at` to be nullable, whereas the top-level patients schema (packages/db-schema/src/schema/patients.ts) has a different medications table with `started_at NOT NULL`. That's a known schema duplication gap. Falling back silently to `created_at` means a NULL `started_at` produced by a partial import path will pass as active-at-event without any signal, hiding the data bug. Consider either logging a warning when the fallback fires, or failing closed for medication rows with null `started_at` (safer — a rule that sees an unknowable med is unhelpful).

**4. No validation or guard on `event.timestamp`.**
If `event.timestamp` is missing, empty, or malformed, the comparisons still execute against an arbitrary string. `undefined > "2026-04-01T..."` is `false` (safe-ish: everything falls through as started-before), but `"not-a-date"` sorts unpredictably. At minimum assert `Number.isFinite(Date.parse(event.timestamp))` at function entry and fail the job rather than silently computing a nonsense snapshot.

**5. Scalability: filter is in-memory over the full patient history.**
The three selects fetch *all* diagnoses, *all* medications, and *all* allergies for the patient unconditionally, then filter in JS. For long-history oncology patients this is already non-trivial (hundreds of rows) and grows monotonically with tenure. The prior implementation had the same issue — this PR does not make it worse — but since we are editing these three queries anyway, adding an SQL-side predicate (`WHERE (onset_date IS NULL OR onset_date <= $at) AND (resolved_date IS NULL OR resolved_date > $at)` etc.) is a cheap win and lets the `idx_diagnoses_patient` index do useful work. Defer-OK for this PR, but worth a follow-up.

**6. Active-diagnosis status handling loses `"chronic"`.**
Previous code filtered `status === "active"`. Diagnoses schema doc comments list `active`, `resolved`, `chronic`. The new `wasActiveAt` treats a row with null `resolved_date` as active regardless of `status`, so `chronic` is now correctly included (likely an improvement — a chronic diagnosis should fire rules). But `entered_in_error` / future statuses that indicate "not actually a real diagnosis" are also now included. Worth a brief allowlist: still include `chronic` but exclude `entered_in_error` and anything representing a logical retraction. Allergies have the same concern — `verification_status = 'entered_in_error'` or `'refuted'` rows are now included.

### Deferred Items
- Apply the same event-time snapshot to `buildPatientContext` (LLM review path) for consistency.
- Normalize timestamp comparison: parse with `Date.parse()` or enforce Z-form at the schema layer; add regression tests for bare-date and offset-form strings.
- Push the snapshot filter to SQL with an index-friendly predicate rather than in-memory filter.
- Exclude diagnoses/allergies whose status indicates retraction (`entered_in_error`, `refuted`) from the rule context.
- Decide fallback policy for NULL `started_at`: warn-and-include vs fail-closed.
- Guard `event.timestamp` validity at function entry.

### Verdict
**REQUEST-CHANGES-soft** — The rule-path fix is correct and the tests are well-targeted, but the inconsistency with the LLM path (item 1) is a clinical-safety concern that should be either addressed here or called out explicitly in the PR body and tracked with a linked issue before merge. The remaining items are reasonable to defer.

blamechris:
Addressed in 7292179: context-builder now filters by event.timestamp matching the rule path.

PR #495, opened 2026-04-16
fix(ai): explicit BullMQ lock tuning + trigger_event_id review idempotency

### LINKED ISSUE #257: BullMQ worker crash mid-job risks lost reviews; idempotency unclear
**Files:** services/ai-oversight/src/workers/review-worker.ts:32-69; services/clinical-notes/src/events.ts

Default BullMQ settings; explicit lockDuration/stalledInterval not configured. Job state isn't checkpointed.

**Fix:** Configure lockDuration + stalled retries explicitly; ensure trigger_event_id-based idempotency in flag dedup.

blamechris:
## Summary
- \`review-worker\`: set \`lockDuration=120s\`, \`stalledInterval=30s\`, \`maxStalledCount=1\` explicitly so a slow (30–90s) healthy review can't be reclaimed by another worker mid-flight.
- \`review-service\`: add a pre-insert idempotency probe — if a \`review_jobs\` row with this \`trigger_event_id\` and \`status='completed'\` already exists, skip. Pairs with #261's flag-level idempotency for partial completions.

## Why
BullMQ's default 30s lockDuration is shorter than the p99 review duration. A slow healthy job could be reclaimed and double-processed; a crashed job was redelivered without any app-level guard against re-executing the full pipeline, leading to duplicate \`review_jobs\` rows and noisy audit data.

## Test plan
- [x] New test file \`review-idempotency.test.ts\` — 2 tests: skips when prior completed row, proceeds when none
- [x] \`pnpm --filter @carebridge/ai-oversight test\` — 280/280 passing

Closes #257

### DIFF
--- a/services/ai-oversight/src/__tests__/review-idempotency.test.ts
+++ b/services/ai-oversight/src/__tests__/review-idempotency.test.ts
@@ -1,0 +1,161 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import type { ClinicalEvent } from "@carebridge/shared-types";
+
+// ─── Mocks ───────────────────────────────────────────────────────
+//
+// This test covers the narrow contract that processReviewJob SKIPS the
+// full review pipeline when a `completed` review_jobs row already exists
+// for the trigger_event_id. We mock the first select().from().where().limit()
+// to return a prior-completed row; if the idempotency check works, the
+// pipeline returns early without touching flag-service, LLM, or context.
+
+const insertValues = vi.fn().mockResolvedValue(undefined);
+const insertMock = vi.fn().mockReturnValue({ values: insertValues });
+
+const updateSetWhere = vi.fn().mockResolvedValue(undefined);
+const updateSet = vi.fn().mockReturnValue({ where: updateSetWhere });
+const updateMock = vi.fn().mockReturnValue({ set: updateSet });
+
+// Idempotency-first select chain: select().from().where().limit()
+const limitMock = vi.fn();
+const selectWhere = vi.fn().mockReturnValue({ limit: limitMock });
+const selectFrom = vi.fn().mockReturnValue({ where: selectWhere });
+const selectMock = vi.fn().mockImplementation(() => ({ from: selectFrom }));
+
+vi.mock("@carebridge/db-schema", () => ({
+  getDb: () => ({
+    insert: insertMock,
+    update: updateMock,
+    select: selectMock,
+    query: { patients: { findFirst: vi.fn() } },
+  }),
+  reviewJobs: { id: "id", trigger_event_id: "trigger_event_id", status: "status" },
+  diagnoses: {},
+  medications: {},
+  patients: {},
+  allergies: {},
+  messages: {},
+  patientObservations: {},
+  labPanels: {},
+  labResults: {},
+  clinicalFlags: {},
+  encounters: {},
+}));
+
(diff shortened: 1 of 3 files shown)


blamechris:
## Review — PR #495 (BullMQ lock tuning + trigger_event_id idempotency)

### Summary
This PR correctly addresses two real problems: BullMQ's 30s default lockDuration is shorter than the p99 review duration (so a slow-but-healthy review could be reclaimed mid-flight and double-executed), and a crash-reclaimed or outbox-replayed event had no app-level guard against re-running the full pipeline. The `lockDuration=120s / stalledInterval=30s / maxStalledCount=1` numbers are defensible and well-commented, and the probe is cheap and safe for the happy path.

### Strengths
- Constants are named and justified with an explicit rationale that survives future refactors — "a BullMQ default change can't silently shift our recovery window" is exactly the right framing for healthcare infra.
- Placing the probe before the `review_jobs` insert means skipped events don't pollute audit counts with `processing` rows that never complete.
- Test mock shape `{id, status}` matches the production `select({id, status})` — no hidden column mismatch.
- Pairs correctly with the flag-level dedup from #261 for partial-completion recovery.

### Issues

1. **[Real gap] Idempotency predicate is too narrow — misses `llm_timeout` / `llm_error` terminal states.** The production code writes three terminal statuses on successful completion (`completed`, `llm_timeout`, `llm_error` — see `llm-timeout-fallback.test.ts` and `review-service.ts:381`) plus `failed` on thrown errors. A redelivered job whose prior run ended at `llm_timeout` will re-execute the full pipeline even though the rule-tier flags already ran and were written. Consider: `inArray(reviewJobs.status, ['completed', 'llm_timeout', 'llm_error'])`. `failed` is the right one to exclude (it means the pipeline threw before flags landed, so retry is desired).

2. **[Real gap] In-flight race between concurrent workers is still open.** The check is `status='completed'` only. If worker A is 40s into processing an event (row written 40s ago, `status='processing'`), and BullMQ redelivers the same event to worker B (stalled-scan false positive, outbox replay, or manual requeue), worker B's probe sees zero completed rows and proceeds — both execute the full pipeline concurrently, hit Claude, and may create duplicate flags for rule IDs that rely on open-flag dedup within narrow windows. A stale-window check on `processing` (e.g., `OR (status='processing' AND created_at > now() - interval '150 seconds')`) would close this without blocking legitimate retries of truly-stuck jobs. Deferred unless you consider it pre-launch critical — the flag-level dedup (#261) likely catches most duplicates, but not all.

3. **[Perf] No index supports the probe query.** `review_jobs` has only `idx_review_jobs_patient(patient_id, status)`. The new probe filters by `trigger_event_id` + `status` with no supporting index. On a production table this becomes a seq scan on every review job — cheap today, linear-degrading as the table grows. An index on `(trigger_event_id, status)` (or just `trigger_event_id`) is a one-line migration. Deferring as its own issue.

### Deferred items
- Broaden idempotency predicate to cover `llm_timeout` / `llm_error` terminal states
- Add stale-window `processing` race check
- Add `review_jobs(trigger_event_id)` index to support the probe

### Redis / lock cost note
BullMQ extends the lock every `lockDuration/2` (~60s) via a single `PEXPIRE` per in-flight job. At expected concurrency this is negligible — the 120s value is fine.

### Verdict
**APPROVE with deferred follow-ups.** The core change is correct and ships real safety value; the gaps above are incremental improvements on the same invariant and better handled as separate, testable PRs.

PR #496, opened 2026-04-16
fix(portal): distinguish empty vs stale vs current clinical data

### LINKED ISSUE #256: Empty data renders identically to 'all clear' — clinician cannot tell
**Files:** services/ai-oversight/src/workers/context-builder.ts:210-211; services/api-gateway/src/routers/clinical-data.ts:87-88

Missing vitals returned as []. UI shows nothing — indistinguishable from 'all normal'. Clinician misses required assessment.

**Fix:** Distinguish 'no data ever recorded' / 'last recorded N ago' / 'sync failure' in API and UI.

blamechris:
## Summary
- Empty-state copy upgraded: vitals/labs tabs now say \"No vitals have ever been recorded\" with a reminder that absence of data is not evidence of normality.
- New \`StaleDataBanner\` surfaces above the data cards when the freshest recording is older than 7 days.
- Labs tab derives freshness from \`collected_at\` with \`created_at\` fallback.

## Why
A 30-day-old BP reading rendered identically to one taken an hour ago, and the terse empty state was easy to misread as \"all clear.\" Clinicians could miss required assessments.

## Test plan
- [x] \`pnpm typecheck\` / \`pnpm lint\` — clean (pre-existing clinician-portal warnings only)
- [ ] Manual: load patient chart with 0 vitals → explicit empty state
- [ ] Manual: load patient chart with only stale vitals → banner renders above cards
- [ ] Manual: stop API → ErrorState still renders

Closes #256

### DIFF
--- a/apps/clinician-portal/app/patients/[id]/page.tsx
+++ b/apps/clinician-portal/app/patients/[id]/page.tsx
@@ -56,9 +56,12 @@ function StaleDataBanner({
 }) {
   const ageMs = Date.now() - new Date(lastRecordedAt).getTime();
   const ageDays = Math.round(ageMs / (24 * 60 * 60 * 1000));
+  // role="status" (polite live region) rather than role="alert" (assertive):
+  // a chronic chart banner should not interrupt a screen-reader mid-sentence
+  // every time the clinician opens a patient with week-old data.
   return (
     <div
-      role="alert"
+      role="status"
       style={{
         background: "var(--color-warning-bg, #fff3cd)",
         border: "1px solid var(--color-warning-border, #ffc107)",
--- a/apps/clinician-portal/app/patients/[id]/page.tsx
+++ b/apps/clinician-portal/app/patients/[id]/page.tsx
@@ -38,6 +38,44 @@ function ErrorState({ label }: { label: string }) {
   );
 }
 
+/** Minimum age (ms) at which a latest vital/lab is flagged as stale. 7 days. */
+const STALE_THRESHOLD_MS = 7 * 24 * 60 * 60 * 1000;
+
+/**
+ * Distinguishes "data present but stale" from "data present and current".
+ * Without this banner, a 30-day-old BP reading renders the same way as one
+ * taken an hour ago — the clinician cannot tell at a glance that the
+ * displayed number is not a current assessment.
+ */
+function StaleDataBanner({
+  lastRecordedAt,
+  label,
+}: {
+  lastRecordedAt: string;
+  label: string;
+}) {
+  const ageMs = Date.now() - new Date(lastRecordedAt).getTime();
+  const ageDays = Math.round(ageMs / (24 * 60 * 60 * 1000));
+  return (
+    <div
+      role="alert"
+      style={{
+        background: "var(--color-warning-bg, #fff3cd)",
+        border: "1px solid var(--color-warning-border, #ffc107)",
+        color: "var(--color-warning-text, #856404)",
+        padding: "12px 16px",
+        borderRadius: 6,
+        fontSize: 14,
+      }}
+    >
+      <strong>Stale {label}:</strong> last recorded {ageDays} day
+      {ageDays === 1 ? "" : "s"} ago (
+      {new Date(lastRecordedAt).toLocaleString()}). Values shown may not
+      reflect the patient&rsquo;s current state.
+    </div>
+  );
+}
+
 function OverviewTab({ patientId }: { patientId: string }) {
   const patientQuery = trpc.patients.getById.useQuery({ id: patientId });
   const diagnosesQuery = trpc.patients.diagnoses.getByPatient.useQuery({ patientId });
@@ -146,14 +184,39 @@ function VitalsTab({ patientId }: { patientId: string }) {
 
   if (latest.length === 0) {
     return (
-      <div className="empty-state">
-        <div className="empty-state-text">No vitals recorded for this patient</div>
+      <div className="empty-state" role="status">
+        <div className="empty-state-text" style={{ fontWeight: 600 }}>
+          No vitals have ever been recorded for this patient.
+        </div>
+        <div
+          style={{
+            color: "var(--text-muted)",
+            fontSize: 13,
+            marginTop: 4,
+          }}
+        >
+          This does NOT mean vitals are normal. A clinical assessment is
+          required.
+        </div>
       </div>
     );
   }
 
+  // Freshest recording across types — drives the staleness banner.
+  const mostRecent = latest.reduce((a, b) =>
+    a.recorded_at > b.recorded_at ? a : b,
+  );
+  const isStale =
+    Date.now() - new Date(mostRecent.recorded_at).getTime() > STALE_THRESHOLD_MS;
+
   return (
     <div style={{ display: "flex", flexDirection: "column", gap: 24 }}>
+      {isStale ? (
+        <StaleDataBanner
+          lastRecordedAt={mostRecent.recorded_at}
+          label="vitals"
+        />
+      ) : null}
       <div className="detail-grid">
         {latest.map((vital, i) => (
           <div key={i} className="stat-card">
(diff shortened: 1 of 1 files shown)


blamechris:
## CareBridge Inspector Review — PR #496

### Summary
Upgrades empty-state copy in `VitalsTab`/`LabsTab` to explicitly reject "absence of data = all clear" misreadings, and adds a `StaleDataBanner` when the freshest reading is older than 7 days. All changes in `apps/clinician-portal/app/patients/[id]/page.tsx`. Clinically well-motivated; fixes a real risk-miscommunication in the chart.

### Strengths
- Empty-state copy explicitly warns that "no data" is not evidence of normality — correct clinical framing.
- `STALE_THRESHOLD_MS` is a named constant with a docstring, not a magic number.
- Labs tab correctly falls back from `collected_at` to `created_at` to avoid crashing on null.
- Pluralisation (`day`/`days`) handled, and empty states use `role="status"` (polite), appropriate for "no data yet."
- Keeps the change to a single file with clear blast radius.

### Issues

**1. `--color-warning-*` tokens do not exist in globals.css — hex fallbacks are always used.**
`apps/clinician-portal/app/globals.css` defines `--warning`, `--warning-bg`, `--warning-border` (lines 13-15) as part of the dark Apple-style palette: `rgba(255, 159, 10, 0.12)` / `rgba(255, 159, 10, 0.25)` / `#ff9f0a`. The PR references `--color-warning-bg`, `--color-warning-border`, `--color-warning-text` which are **not defined anywhere in the repo** — the banner will always render with the hardcoded Bootstrap-mustard fallbacks (`#fff3cd` / `#ffc107` / `#856404`), which visually clash with every other warning surface in the portal (`inbox/page.tsx`, `settings/page.tsx`, the vitals range chip at line 236 of this same file). Either switch to the existing tokens (`var(--warning-bg)` / `var(--warning-border)` / `var(--warning)`), or add the `--color-warning-*` family to `globals.css`. As written this silently diverges from the design system.

**2. `role=\"alert\"` is too assertive for a stale-data surface.**
`role=\"alert\"` is an ARIA live region with `aria-live=\"assertive\"` — it interrupts whatever the screen reader is announcing. Appropriate for errors or safety-critical warnings that demand immediate attention (e.g. the `ErrorState` on line 34). A 7-day stale indicator is informational context that the clinician should notice but doesn't need to interrupt their current reading. `role=\"status\"` (polite) is the correct choice, matching what the same PR already uses for the empty states. Recommend downgrading to `role=\"status\"`.

**3. ISO timestamp string comparison — same latent concern as #494.**
`latest.reduce((a, b) => a.recorded_at > b.recorded_at ? a : b)` and `panels.map(...).sort().slice(-1)[0]` both rely on lexicographic ordering of the ISO strings. Works correctly **only** if every timestamp is Z-suffixed UTC in the same format. Mixed timezone offsets (`2026-04-16T10:00:00-05:00` vs `2026-04-16T15:00:00Z` refer to the same instant but sort differently) or missing milliseconds produce wrong-most-recent results. Seed data is currently consistent, but nothing in the DB schema (both columns are `text`) enforces this. Safer: `.map(x => new Date(x.recorded_at).getTime())` before comparing. Low likelihood today, high cost if it breaks (wrong banner shown or hidden on stale data).

### Deferred Items
- D1: Define `--color-warning-*` token family in globals.css OR align the banner to existing `--warning-*` tokens — tracked as #XXX.
- D2: Switch `StaleDataBanner` to `role=\"status\"`; reserve `role=\"alert\"` for clinical errors — tracked as #XXX.
- D3: Normalize ISO-string timestamp comparisons across the clinician-portal to use `Date.getTime()` or a shared helper — tracked as #XXX.
- D4 (noted, not blocking): No component-test suite exists for `app/patients/[id]/page.tsx`; only `src/__tests__/next-config-headers.test.ts` in the portal. Consistent with current project convention — not a gate for this PR.
- D5 (noted, not blocking): Copy is English-only. No i18n framework installed in the clinician-portal (`next-intl`/`react-i18next` absent); out of scope here but worth considering before externalisation.

### Verdict
**Approve with follow-ups.** The clinical intent is right and the logic is correct. Ship it; address D1-D3 in follow-up PRs. The token mismatch (D1) is the most visible since the banner will look out-of-place in the dark theme — worth fast-following.

blamechris:
# Agent Review

## Summary
PR #496 replaces terse empty-state copy in `VitalsTab`/`LabsTab` with an explicit "No vitals have ever been recorded" + "does NOT mean vitals are normal" disclaimer, and introduces `StaleDataBanner` that surfaces when the freshest reading is older than 7 days. Labs freshness derives from `collected_at` with a `created_at` fallback. This directly addresses a real clinical-safety failure mode where a 30-day-old BP reading rendered identically to a current one.

## Strengths
- Clinically correct framing: absence of data is not evidence of normality. The copy explicitly says so.
- `role="status"` on empty states (polite live region) is the right a11y choice for non-urgent "nothing here yet" content.
- Safe `ageDays === 1` pluralization.
- Labs fallback chain `collected_at ?? created_at` handles the real schema where `collected_at` can be null.
- Threshold is a named constant with a clear comment.
- Minimal blast radius — one file, additive changes.

## Issues

### 1. `StaleDataBanner` references undefined CSS custom properties (cosmetic, not functional)
The inline styles use `var(--color-warning-bg, #fff3cd)`, `var(--color-warning-border, #ffc107)`, `var(--color-warning-text, #856404)`. None of these tokens exist in `apps/clinician-portal/app/globals.css` (verified — only `--text-primary`, `--text-secondary`, `--text-muted`, and a few others are defined). The banner will always render the hardcoded light-mode yellow fallbacks, which clashes with the portal's dark theme (`body { background: #0a0a0a }`, `--text-muted: #666666`). Not blocking — the fallbacks do render — but it's a visible styling mismatch. Deferred: https://github.com/blamechris/carebridge/issues/530

### 2. `role="alert"` on `StaleDataBanner` is too assertive
ARIA `role="alert"` is an assertive live region — screen readers interrupt the user and announce the content immediately. Stale data is informational, not urgent. Every time a clinician tabs to a patient with stale vitals the screen reader will interrupt whatever they were reading. `role="status"` (polite) matches the intent and is already used for the empty states in the same PR — the inconsistency is telling. Recommend changing `role="alert"` → `role="status"` (or drop the role and add `aria-live="polite"`).

### 3. ISO-string lexicographic compare for freshest record (minor)
```ts
const mostRecent = latest.reduce((a, b) =>
  a.recorded_at > b.recorded_at ? a : b,
);
```
This works **only** because `recorded_at` is validated by `z.string().datetime()` which requires the `Z` suffix — so all values are UTC and lexicographic compare is numeric-equivalent. If validation ever loosens (e.g. `{ offset: true }` to accept `+00:00`) this comparison silently gives wrong answers (a `+00:00` string sorts BEFORE a `Z` string at the same instant). Low-risk today, but `new Date(a.recorded_at).getTime() > new Date(b.recorded_at).getTime()` is cheap and future-proof. The labs version uses `.sort().slice(-1)` which has the same underlying assumption.

### 4. No component tests for empty/stale/current branches
The three states this PR distinguishes are exactly the regression surface the clinical-safety goal depends on. A future refactor that silently drops the banner or reverts the disclaimer copy would not be caught. Clinician-portal currently has zero component tests (only `next-config-headers.test.ts`). Deferred: https://github.com/blamechris/carebridge/issues/531

### 5. Inline component + magic threshold limit reuse
`StaleDataBanner` and `STALE_THRESHOLD_MS` are defined inside `page.tsx`. Medications, procedures, and the patient-portal mirror all have the same freshness-vs-currency ambiguity. Deferred: https://github.com/blamechris/carebridge/issues/532

## Deferred Items
- https://github.com/blamechris/carebridge/issues/530 — define `--color-warning-*` design tokens in globals.css
- https://github.com/blamechris/carebridge/issues/531 — component tests for empty / stale / current states
- https://github.com/blamechris/carebridge/issues/532 — extract `StaleDataBanner` + threshold to shared module

## Verdict
**Approve with one requested change** — flip `role="alert"` to `role="status"` (or `aria-live="polite"`) before merge. This is a ~1-character fix and avoids shipping a screen-reader regression on every patient chart with stale vitals. Everything else (undefined CSS tokens, tests, extraction) is legitimately deferrable and tracked.

PR #499, opened 2026-04-16
fix(clinical): baseline-aware trend computation

### LINKED ISSUE #249: Trend direction logic ignores baseline (creatinine rise from baseline 0.7→1.2 looks 'stable')
**File:** packages/medical-logic/src/trend-utils.ts:35-67

Hardcoded 'up is good/bad' ignores per-patient baseline.

**Fix:** Trend computed against baseline parameter; show delta from baseline.

blamechris:
## Summary
- Add \`calculateDeltaFromBaseline(values, baseline?)\` — compares current value against an explicit baseline (or \`values[0]\` when omitted).
- Add \`getTrendInfoFromBaseline(values, goodDirection, baseline?, isOutOfRange?)\` — same \`TrendInfo\` shape as \`getTrendInfo\`, so UI can swap in.
- Guard against divide-by-zero on \`pctChange\` when baseline is 0.

## Why
A slow creatinine drift of 0.7 → 0.8 → 0.9 → 1.1 → 1.2 reports as \"+0.1 (+9%) slightly rising\" under \`calculateDelta\` — nearly stable — when it is actually a ~70% rise from baseline and KDIGO stage-1 AKI territory. Baseline-aware delta preserves the true trajectory.

## Test plan
- [x] 6 new unit tests — explicit baseline, fallback-to-values[0], zero-baseline, neutral-at-baseline
- [x] \`pnpm --filter @carebridge/medical-logic test\` — 146/146 passing

Closes #249

### DIFF
--- a/packages/medical-logic/src/__tests__/trend-utils.test.ts
+++ b/packages/medical-logic/src/__tests__/trend-utils.test.ts
@@ -1,11 +1,13 @@
 import { describe, it, expect } from "vitest";
 import {
   calculateDelta,
+  calculateDeltaFromBaseline,
   classifyTrend,
   detectAKI,
   getGoodDirection,
   getTrendColor,
   getTrendInfo,
+  getTrendInfoFromBaseline,
   isOutOfRange,
   getRangeFlag,
   formatDelta,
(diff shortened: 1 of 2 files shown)


blamechris:
## Summary

Adds `calculateDeltaFromBaseline(values, baseline?)` and `getTrendInfoFromBaseline(values, goodDirection, baseline?, isOutOfRange?)` to `packages/medical-logic/src/trend-utils.ts`. Closes #249. Existing `calculateDelta` / `getTrendInfo` are untouched. 6 new unit tests, CI green (Build / Test / Typecheck).

The clinical motivation is sound: a 0.7 → 0.8 → 0.9 → 1.1 → 1.2 creatinine drift is ~70% from baseline (KDIGO stage-1 AKI territory) but `calculateDelta` only surfaces the last-two-point +9% step. A baseline-aware variant is the right tool.

## Strengths

- Clear JSDoc with concrete clinical example (the exact 5-point creatinine series the tests also exercise).
- Zero-baseline guard (`base !== 0 ? ... : 0`) avoids the `Infinity`/`NaN` that `calculateDelta` can produce when `previous === 0`.
- Tests cover the interesting cases: explicit baseline, `values[0]` fallback, single-value + no baseline, zero baseline, and a neutral-at-baseline `getTrendInfoFromBaseline` case.
- Additive change — no behavior shift for existing callers, low blast radius.

## Issues

**1. Asymmetric single-value semantics deserve a line in the JSDoc.** `calculateDeltaFromBaseline([1.2])` returns `null` but `calculateDeltaFromBaseline([1.2], 0.7)` returns a valid delta. That divergence from `calculateDelta` (which always requires length ≥ 2) is intentional and tested, but the JSDoc never states it — a caller reading only the docstring would assume both functions need ≥ 2 points. One line like "returns a delta even for a single-value series when an explicit baseline is supplied" would prevent a footgun.

**2. The length-check lives below the `base` assignment.** The body reads:
```ts
const current = values[values.length - 1]!;
const base = baseline ?? values[0]!;
if (values.length < 2 && baseline === undefined) return null;
```
This is correct because the short-circuit in `??` means `values[0]!` is only evaluated when `baseline === undefined` — but the non-null assertion on `values[0]!` when `values.length === 0` would be a lie if `values.length === 0 && baseline !== undefined`. Current code handles that via the `values.length === 0` early-return on line 1, so it's safe, but moving the length guard to the top would make the invariants obvious without relying on two separate guards to compose correctly.

**3. No call sites are migrated.** `services/ai-oversight/src/workers/context-builder.ts:143` is the canonical consumer of `calculateDelta` and computes vitals trends fed to the LLM review — exactly the surface where "slow drift from baseline" matters for AI oversight. The patient-portal labs/health-summary pages also don't consume any trend util yet. The PR closes #249 but ships a dormant helper; a follow-up to wire `context-builder` to `calculateDeltaFromBaseline` (using a per-patient baseline from prior admissions, or `values[0]` of the window) is what actually delivers the clinical win. Worth a tracked issue.

**4. Overlap with `classifyTrend` is worth acknowledging.** `classifyTrend` (line 173) is already baseline-aware — it compares `values[0]` to `values[values.length - 1]`, the same fallback the new function uses. `calculateDeltaFromBaseline` complements it (numeric delta + color/arrow vs. direction string), but the two should cross-reference in their JSDoc so a future contributor doesn't write a third overlapping helper. Consider: "See also `classifyTrend` for a direction-only classification over the same baseline semantics."

**5. Zero-baseline `pctChange = 0` silently hides an infinite rise.** `calculateDeltaFromBaseline([5], 0)` returns `pctChange: 0` (tested explicitly). For a metric where 0 is a real measurement (e.g. pain level, output volumes) this reads as "no change" when the absolute change is +5. The `change` field is still correct, so UI that shows the delta in units is fine, but anything thresholding on `pctChange` would under-fire. Consider returning `null` for `pctChange` (and updating the `Delta` type) or documenting that callers must consult `change` when `previous === 0`.

## Deferred Items

- Wiring `context-builder.ts` and any patient-portal trend displays to the baseline-aware variant — this is where clinical value lands.
- Deciding whether `pctChange` should be `number | null` to distinguish "0% change" from "divided by zero".

## Verdict

**Approve with follow-ups.** The helper is correct, tested, and backward-compatible. Ship it, then file the migration + zero-baseline-semantics tickets so the dormant code actually flows into AI oversight and UI.

PR #500, opened 2026-04-16
fix(clinical): lab-result unit validation against per-test allow-list

### LINKED ISSUE #248: No unit validation: glucose 200 mg/dL vs mmol/L confusion (fatal)
**Files:** packages/shared-types/src/clinical-data.ts:171-256; medical-validation.ts:99-127

System accepts numeric labs without enforcing the unit matches expected. mg/dL vs mmol/L glucose ambiguity is a known sentinel-event source.

**Fix:** Per-test allowed units enum; reject/warn on mismatch.

blamechris:
## Summary
- \`CommonLabTest\` reference entries gain optional \`allowed_units: string[]\`.
- Populated for **Glucose, BUN, Creatinine, Calcium** (canonical mg/dL only) and **Sodium, Potassium, Chloride, CO2** (accept both mEq/L and mmol/L).
- \`validateLabResult\`:
  - allow-listed test + unit not in list → **error** (reject)
  - non-allow-listed test + unit mismatch → **warning**
  - allow-listed test + no unit at all → **error** (ambiguous)

## Why
Glucose submitted as 200 mmol/L is fatal at mg/dL dosing — the 18:1 ratio for glucose and 88.4:1 for creatinine make unit confusion one of the classic sentinel-event sources in inpatient systems.

## Test plan
- [x] 6 new tests: reject glucose in mmol/L, accept canonical, reject no-unit glucose, accept both K units, warn-not-error on non-allow-listed mismatch
- [x] Existing WBC tests migrated from \"10^3/uL\" synonym to canonical \"K/uL\"
- [x] \`pnpm --filter @carebridge/medical-logic test\` — 144/144 passing

Closes #248

### DIFF
--- a/packages/medical-logic/src/__tests__/medical-validation.test.ts
+++ b/packages/medical-logic/src/__tests__/medical-validation.test.ts
@@ -131,21 +131,19 @@ describe("validateMedicationDose", () => {
 
 describe("validateLabResult", () => {
   it("returns valid for value in typical range", () => {
-    const result = validateLabResult("WBC", 7.5, "10^3/uL");
+    const result = validateLabResult("WBC", 7.5, "K/uL");
     expect(result.valid).toBe(true);
     expect(result.warnings).toHaveLength(0);
   });
 
   it("warns on value below typical range", () => {
-    const result = validateLabResult("WBC", 1.0, "10^3/uL");
-    expect(result.warnings.length).toBeGreaterThan(0);
-    expect(result.warnings[0]).toContain("below typical range");
+    const result = validateLabResult("WBC", 1.0, "K/uL");
+    expect(result.warnings.some((w) => w.includes("below typical range"))).toBe(true);
   });
 
   it("warns on value above typical range", () => {
-    const result = validateLabResult("WBC", 50.0, "10^3/uL");
-    expect(result.warnings.length).toBeGreaterThan(0);
-    expect(result.warnings[0]).toContain("above typical range");
+    const result = validateLabResult("WBC", 50.0, "K/uL");
+    expect(result.warnings.some((w) => w.includes("above typical range"))).toBe(true);
   });
 
   it("returns valid for unknown test name (no reference data)", () => {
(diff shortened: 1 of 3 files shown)


blamechris:
## Summary
This PR adds an optional `allowed_units: string[]` to `CommonLabTest` reference entries and teaches `validateLabResult` to reject unit mismatches on high-stakes tests (Glucose/BUN/Creatinine/Calcium: strict mg/dL; Na/K/Cl/CO2: mEq/L or mmol/L). Motivation is correctly framed around the ~18:1 glucose and ~88.4:1 creatinine conversion factors — this is a real class-of-event fix, not a stylistic one. Well-scoped change, clean typing, six new tests, existing WBC tests migrated off the `10^3/uL` synonym to the canonical `K/uL`. No AI/Claude attribution in diff.

## Strengths
- The two-tier behavior (strict allow-list → error, no allow-list → warning) is the right call and matches how a clinician would triage "unknown unit on a Potassium" versus "unexpected unit on an Albumin."
- Distinguishing *missing unit* from *wrong unit* in the error message is good UX; the ambiguity failure mode is exactly the sentinel-event shape.
- JSDoc on `CommonLabTest` spells out *when* to populate `allowed_units`, so future contributors have a principled rule for new entries.
- Scope discipline — only the lab reference map and validator changed, no shotgun refactor of FHIR/UCUM code.

## Issues
1. **Unit comparison is exact-string, case-sensitive, whitespace-sensitive.** `"MG/DL"`, `"mg/dl"`, `"mg / dL"`, or `" mg/dL"` all fail. FHIR inbound bundles and HL7 v2 feeds routinely normalize to lower-case UCUM (`"mg/dl"` is common in legacy data). A trimmed + case-insensitive compare, or a single `normalizeUnit()` helper, would eliminate a whole class of false-reject bugs with no loss of safety. This is a real issue — it will bite on the first real-world FHIR import.
2. **Allow-list strings are display forms, not UCUM codes** — and the codebase already has a UCUM mapping table (`services/fhir-gateway/src/generators/observation.ts:11` `UNIT_TO_UCUM`) that converts `"mEq/L" → "meq/L"`, `"K/uL" → "10*3/uL"`. A FHIR `Observation.valueQuantity.code = "meq/L"` (canonical UCUM) coming back through an ingest path would be rejected by this validator because the allow-list only contains `"mEq/L"`. The validator and the UCUM map should share a normalization layer — otherwise the two representations will drift.
3. **Missing-unit → hard error has no current upstream caller that legitimately omits unit** (`createLabPanelSchema` requires `unit.min(1)`), so the error path is defensive-only today — which is fine, but worth a code comment noting that if any future caller (e.g. a manual-entry UI with a separate unit dropdown) constructs a `LabResult` without a unit *pre-validation*, it will fail here. Not a blocker, just a landmine to flag.
4. **HbA1c falls into the warning path with no allow-list**, and the test explicitly asserts `"mmol/mol"` produces a warning. In reality, `mmol/mol` is the IFCC international standard and 5.5 mmol/mol is a legitimate value — warning on it is correct behavior *today* (it flags a unit the system doesn't interpret) but will become spurious once CareBridge supports IFCC. Consider adding `HbA1c: { ..., allowed_units: ["%", "mmol/mol"] }` now with a conversion note, or leaving a TODO — otherwise the warning will accumulate noise for international sites.

## Deferred Items
- Wire `validateLabResult` into `labRepo.createLabPanel` and `medlens-bridge.importLabs` as a pre-insert gate. Today the allow-list is advisory — nothing stops persistence of `Glucose + mmol/L`. (Tracking issue to be filed.)
- Centralize unit normalization: a `normalizeUnit(u: string): string` helper used by both the FHIR generator's `UNIT_TO_UCUM` map and `validateLabResult`, so display-form and UCUM-form of the same unit are interchangeable. (Tracking issue to be filed.)
- Broaden HbA1c to accept `mmol/mol` now, before international data lands. (Tracking issue to be filed.)

## Verdict
**Approve with follow-ups.** The safety logic is correct and the scope is tight. The case-sensitivity and UCUM-form gap are real and will produce false rejects on real FHIR data, but they're incremental fixes on top of this PR, not reasons to block the allow-list itself. Merge, then land the three follow-up issues before wiring the validator into any write path.

PR #501, opened 2026-04-16
feat(ai): persist full rule-evaluation output for audit

### LINKED ISSUE #241: Review job audit lacks rule outputs; only stores rule IDs
**File:** packages/db-schema/src/schema/ai-flags.ts:46-64

reviewJobs.rules_fired stores rule IDs but not the matched drugs, severity, or reasoning. Insufficient for forensic/regulatory audit.

**Fix:** Add rules_output JSONB column with full match details; persist (redacted) prompt.

blamechris:
## Summary
- Migration **0032** adds \`review_jobs.rules_output jsonb DEFAULT '[]'\`.
- Drizzle schema exposes it as an array of \`RuleFlag\`-shaped records.
- \`review-service\` populates it with \`allRuleFlags\` in the same update that closes out the \`review_jobs\` row.

## Why
\`rules_fired\` stored only rule IDs. For regulatory review / adverse-event investigation / liability defense we need to reproduce the exact rule output — severity, category, rationale, matched-drug context, notify_specialties — at decision time without replaying the rule engine against (possibly-mutated) patient state. HIPAA §164.308(a)(1)(ii)(D) / §164.312(b) expects that.

## Test plan
- [x] \`pnpm --filter @carebridge/db-schema build\` — clean
- [x] \`pnpm --filter @carebridge/ai-oversight test\` — 278/278 passing
- [ ] Manual: apply migration 0032 to dev DB, confirm column exists with default '[]'

Closes #241

### DIFF
--- a/packages/db-schema/drizzle/0032_review_jobs_rules_output.sql
+++ b/packages/db-schema/drizzle/0032_review_jobs_rules_output.sql
@@ -1,0 +1,14 @@
+-- Migration: Persist full rule-evaluation output on review_jobs.
+--
+-- Previously review_jobs only stored rules_fired as an array of rule IDs.
+-- For forensic / regulatory audit (HIPAA §164.308(a)(1)(ii)(D), §164.312(b))
+-- that is not enough: when a decision is challenged we need the severity,
+-- category, matched-drug context, rationale, and notify_specialties that
+-- each rule produced — not just the rule's static ID.
+--
+-- rules_output is a jsonb array of RuleFlag entries (see packages/shared-
+-- types for the canonical shape). Nullable default '[]' keeps pre-existing
+-- rows valid; new writes always populate.
+
+ALTER TABLE review_jobs
+  ADD COLUMN IF NOT EXISTS rules_output jsonb NOT NULL DEFAULT '[]';
--- a/packages/db-schema/src/schema/ai-flags.ts
+++ b/packages/db-schema/src/schema/ai-flags.ts
@@ -69,6 +69,12 @@ export const reviewJobs = pgTable("review_jobs", {
   context_hash: text("context_hash"),
   rules_evaluated: jsonb("rules_evaluated").$type<string[]>().default([]),
   rules_fired: jsonb("rules_fired").$type<string[]>().default([]),
+  // Full rule-evaluation output: for each fired rule we persist the exact
+  // RuleFlag (severity, category, summary, rationale, suggested_action,
+  // notify_specialties, rule_id) so regulatory and forensic audits can
+  // reconstruct the decision without replaying the rule engine against
+  // (possibly-mutated) patient state. See migration 0032.
+  rules_output: jsonb("rules_output").$type<Array<Record<string, unknown>>>().default([]),
   llm_request_tokens: integer("llm_request_tokens"),
   llm_response_tokens: integer("llm_response_tokens"),
   redacted_prompt: text("redacted_prompt"),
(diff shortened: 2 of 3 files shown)


blamechris:
## Agent Review — PR #501

### Summary
Adds migration `0032_review_jobs_rules_output.sql` introducing `rules_output jsonb NOT NULL DEFAULT '[]'` on `review_jobs`, updates the Drizzle schema, and persists the full `RuleFlag[]` on job completion in `review-service.processReviewJob`. Closes #241.

### Strengths
- Migration is cleanly backward-compatible: `NOT NULL DEFAULT '[]'` means no backfill is needed and existing rows remain queryable.
- Good regulatory framing in the migration header (HIPAA §164.308(a)(1)(ii)(D) / §164.312(b)) — matches the reasoning style already used in `docs/hipaa-retention.md` and `docs/cds-exemption.md`.
- Stored alongside `redacted_prompt` / `redaction_audit` on the same row, so forensic reconstruction is one query.
- Tiny surface area (24 additions, 0 deletions), localized to one service write site.

### Issues
1. **PHI retention/encryption story not updated for the new column.** `rationale` and `summary` on `RuleFlag` are free-form strings produced by rules like `checkAllergyMedication` and `checkDrugInteractions`, which embed matched drug names and allergy substrates. These are PHI-adjacent. The existing `review_jobs.redacted_prompt` is called out in `docs/anthropic-baa.md` and `docs/cds-exemption.md`, but `rules_output` is not. The table has no retention trigger (unlike `audit_log` per migration 0012) and no column-level encryption (unlike `patient_diagnosis_notes` per 0028). Either (a) document that `rules_output` falls under the same 7-year retention as audit, (b) run it through PHI redaction on write, or (c) state explicitly that rule-flag text is de-identified by construction. Right now the PR sidesteps this.
2. **Audit immutability is NOT inherited.** The migration adds a column to `review_jobs`, which does NOT have the append-only triggers from migration 0012 (those are scoped to `audit_log` only). That's a defensible design choice (review jobs are operational, not the audit trail), but the PR claims this is for "forensic / regulatory audit" — forensic evidence should be tamper-evident. Worth a one-line note in the migration header clarifying that the authoritative audit trail remains `audit_log` and `rules_output` is supplementary.
3. **Type safety regression via `$type<Array<Record<string, unknown>>>()`.** `RuleFlag` already exists in `services/ai-oversight/src/rules/critical-values.ts:326` with a precise shape. Using `Record<string, unknown>` loses all type-checking on reads — downstream code that queries `review_jobs.rules_output` gets no IDE help and no structural guarantees. The correct fix is to lift `RuleFlag` into `packages/shared-types` (the migration header even says "see packages/shared-types for the canonical shape" — but it isn't there) and use `$type<RuleFlag[]>()`.
4. **Double cast `allRuleFlags as unknown as Array<Record<string, unknown>>` in review-service.ts.** Symptom of #3. Once `RuleFlag` lives in shared-types and the schema uses `$type<RuleFlag[]>()`, this cast becomes a no-op assignment.
5. **Migration header references "see packages/shared-types for the canonical shape"** but `RuleFlag` does not exist there. Either move the type or update the comment — documentation drift on day one.

### Deferred Items
- Consider a companion migration to add a retention/archival policy for `review_jobs` mirroring the `audit_log` plan in `docs/hipaa-retention.md`.
- Evaluate whether `rules_output` should be indexed for specific queries (e.g., severity filter for retrospective safety analytics). Not needed for merge.

### Verdict
**Approve with follow-ups.** The migration and write path are correct and safe. The type-safety issues and retention/encryption documentation gap should be addressed in follow-up PRs before the column accumulates volume. Filing `from-review` issues.

PR #502, opened 2026-04-16
fix(gateway): minimum-necessary projection on patient record reads

### LINKED ISSUE #283: API responses return entire patient record when UI needs only name/MRN
**File:** services/api-gateway/src/routers/patient-records.ts:115-119,131

Endpoints return insurance_id, emergency contact, clinical notes etc. by default. Violates minimum necessary.

**Fix:** Field-level Drizzle selects per use case; create thin DTOs.

blamechris:
## Summary
- \`patients.getById\` switches from \`SELECT *\` to an explicit field projection that excludes \`insurance_id\`, \`emergency_contact_name\`, \`emergency_contact_phone\`, and \`notes\`.
- New \`patients.getSummary(id)\` returns only \`{ id, name, mrn }\` for banner / pick-list / lookup use cases.

## Why
HIPAA §164.502(b) requires minimum-necessary access. The UI consumes name, MRN, DOB, sex, diagnosis, and allergy_status — billing and emergency-contact PHI were leaking to every caller regardless of use case.

## Breaking change risk
No UI currently reads the dropped fields (verified via grep). Dedicated billing/front-desk endpoints should be added when those use cases land.

## Test plan
- [x] \`pnpm --filter @carebridge/api-gateway test\` — 207/207 passing
- [ ] Manual: load clinician chart and patient portal home → confirm banner renders

Closes #283

### DIFF
--- a/services/api-gateway/src/routers/patient-records.ts
+++ b/services/api-gateway/src/routers/patient-records.ts
@@ -124,13 +124,54 @@ export const patientRecordsRbacRouter = t.router({
     }),
 
   // Reading a specific patient record is patient-scoped: input.id is the patientId.
+  //
+  // Explicit field selection (HIPAA §164.502(b) minimum necessary): insurance_id,
+  // emergency_contact_*, and patient-level free-text notes are intentionally
+  // NOT returned here. Callers that need them must use a dedicated endpoint —
+  // that scopes the PHI exposure to the use case rather than every getById
+  // implicitly leaking billing and contact data to every consumer.
   getById: protectedProcedure
     .input(z.object({ id: z.string() }))
     .query(async ({ ctx, input }) => {
       await enforcePatientAccess(ctx.user, input.id);
       const db = getDb();
       const [patient] = await db
-        .select()
+        .select({
+          id: patients.id,
+          name: patients.name,
+          date_of_birth: patients.date_of_birth,
+          biological_sex: patients.biological_sex,
+          diagnosis: patients.diagnosis,
+          mrn: patients.mrn,
+          mrn_hmac: patients.mrn_hmac,
+          primary_provider_id: patients.primary_provider_id,
+          allergy_status: patients.allergy_status,
+          weight_kg: patients.weight_kg,
+          created_at: patients.created_at,
+          updated_at: patients.updated_at,
+        })
+        .from(patients)
+        .where(eq(patients.id, input.id));
+      return patient ?? null;
+    }),
+
+  // Minimum-necessary summary for banner / lookup use cases.
+  //
+  // Returns only the fields needed to identify a patient — no date-of-birth,
+  // no diagnosis, no weight, no billing data. Callers that need richer data
+  // should prefer `getById`; callers that only need a name/MRN tile should
+  // use this endpoint to avoid incidentally fetching PHI.
+  getSummary: protectedProcedure
+    .input(z.object({ id: z.string() }))
+    .query(async ({ ctx, input }) => {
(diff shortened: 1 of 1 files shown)


blamechris:
# Agent Review — PR #502

## Summary
Switches `patients.getById` from `SELECT *` to an explicit field projection (drops `insurance_id`, `emergency_contact_name`, `emergency_contact_phone`, `notes`) and introduces a minimum-necessary `patients.getSummary(id)` returning `{ id, name, mrn }`. Single-file change in `services/api-gateway/src/routers/patient-records.ts` (+42/-1).

## Strengths
- Correct HIPAA §164.502(b) framing — minimum-necessary projection on a high-traffic read path is a real PHI-exposure reduction.
- Projection list matches actual UI consumption (verified: `apps/clinician-portal/app/patients/[id]/page.tsx`, `apps/clinician-portal/app/notes/[id]/page.tsx`, `apps/patient-portal/src/lib/use-my-patient.ts`).
- Grep claim holds: only non-dist references to the dropped fields are `packages/db-schema/schema/patients.ts`, validators, seed, and `README.md` — no UI/service reader.
- `getSummary` reuses `enforcePatientAccess`, so it is not an unauthenticated enumeration oracle.
- Clear comments explaining *why* each field is (or isn't) returned, which will deter future drift back to `SELECT *`.

## Issues

### 1. Inconsistent projection on `list` procedures (real, medium)
`getById` now returns minimum-necessary, but three `list` paths in the same file still do `db.select().from(patients)` and therefore return `insurance_id`, `emergency_contact_*`, and `notes` on every row:
- L152-155 (patient self-lookup)
- L160 (admin)
- L193 (family_caregiver)

Only the clinician path (L198-219) is already projected. The admin path is the worst offender — a bulk list that ships every patient's insurance/emergency/notes PHI to the client. This arguably defeats the purpose of the getById narrowing, since anything a clinician would fetch via getById is already in the admin/caregiver list response. Suggest adding the same projection to all four list branches (or a shared `patientSummaryColumns`/`patientChartColumns` const).

### 2. `getSummary` has no rate limit (real, low)
`enforcePatientAccess` blocks cross-patient access for non-admins, but an admin or a compromised clinician session can hit `getSummary` in a tight loop to enumerate names/MRNs against guessed IDs. The repo already has an IP-bucket rate-limit pattern (`services/api-gateway/src/__tests__/accept-invite-rate-limit.test.ts`). Worth applying a per-user limiter to `getSummary` (and arguably `getById`) before launch.

### 3. `primary_provider_id` in getById (minor, discuss)
Included in the new projection. It is a plaintext internal user UUID (`packages/db-schema/src/schema/patients.ts:24`), not a §164.514 direct identifier, and it's needed for the chart header. Not a blocker, but worth noting — if it is ever used to join on a `users` table that returns a provider name to a patient-role caller, you'd want a RBAC check there.

### 4. No test coverage for the projection change (real, low)
No test in `services/api-gateway/src/__tests__/` asserts that the dropped fields are absent from `getById`/`getSummary` responses. A regression to `SELECT *` would pass CI silently. Trivial to add: call the procedure in a test and `expect(result).not.toHaveProperty("insurance_id")` etc.

## Deferred Items
- Dedicated billing and emergency-contact endpoints (called out in the PR body — deferred until those use cases land, which is the right call).

## Verdict
Approve with follow-ups. The change is correct, narrow, and improves the privacy posture. The `list`-projection inconsistency is the one finding I'd want tracked before this ships to production, but it is out-of-scope-appropriate to land this PR first.

PR #503, opened 2026-04-16
fix(fhir): audit every patient bundle export + bundle expiry metadata

### LINKED ISSUE #290: FHIR exportPatient bundle returned without expiry, encryption, or signed URL
**File:** services/fhir-gateway/src/router.ts:83-178

Complete patient bundle returned inline as JSON. No signed URL, no expiry, no password protection, cacheable by intermediaries/browsers.

**Fix:** Stream to signed, short-TTL download URL; Cache-Control: no-store; Content-Disposition: attachment; audit every export.

blamechris:
## Summary
- \`exportPatient\` writes an \`audit_log\` row (\`action: \"export\", resource_type: \"fhir_bundle\"\`) **before** assembling the bundle, so a crash between generation and delivery still leaves a trail.
- Response gains a non-FHIR \`meta\` block: \`export_id\`, \`exported_at\`, \`exported_by\`, and \`recommended_purge_at\` (15 min from generation).
- Inline docstring directs callers to treat the response as ephemeral and never cache it.

## Scope
This is the narrow audit-side fix. Signed short-TTL URL delivery, \`Cache-Control: no-store\`, and \`Content-Disposition: attachment\` remain follow-up work — they require a new HTTP delivery layer outside the tRPC surface.

## Test plan
- [x] \`pnpm --filter @carebridge/fhir-gateway test\` — 67/67 passing
- [ ] Manual: trigger exportPatient, confirm audit_log row exists with \`action='export'\` and matching export_id

Closes #290

### DIFF
--- a/services/fhir-gateway/src/__tests__/export-bundle-audit.test.ts
+++ b/services/fhir-gateway/src/__tests__/export-bundle-audit.test.ts
@@ -1,0 +1,237 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+
+// ── Mock DB ──────────────────────────────────────────────────────
+// Capture every `db.insert(table).values(row)` call so we can assert the
+// exportPatient procedure writes an audit_log row with the right action,
+// success flag, and http_status_code on both the happy path and the
+// patient-not-found failure path.
+type InsertCall = { table: unknown; row: Record<string, unknown> };
+const insertCalls: InsertCall[] = [];
+
+const insertMock = vi.fn((table: unknown) => ({
+  values: vi.fn(async (row: Record<string, unknown>) => {
+    insertCalls.push({ table, row });
+  }),
+}));
+
+// Control what `db.select().from(table).where(...)` resolves to. The
+// exportPatient procedure issues 6 distinct reads: one `patients` lookup,
+// then vitals/labPanels/medications/diagnoses/allergies in parallel, then
+// zero-or-more labResults reads per panel. We route by table reference.
+const patientsTable = { __name: "patients" };
+const vitalsTable = { __name: "vitals" };
+const labPanelsTable = { __name: "lab_panels" };
+const labResultsTable = { __name: "lab_results" };
+const medicationsTable = { __name: "medications" };
+const diagnosesTable = { __name: "diagnoses" };
+const allergiesTable = { __name: "allergies" };
+const fhirResourcesTable = { __name: "fhir_resources" };
+const auditLogTable = { __name: "audit_log" };
+
+let patientRows: Record<string, unknown>[] = [];
+// When set, reading this table throws the given error — used to simulate
+// a mid-bundle DB failure so we can assert the failure audit path.
+let throwOnTable: { table: unknown; error: Error } | null = null;
+
+function selectMock() {
+  return {
+    from: (table: unknown) => ({
+      where: async () => {
+        if (throwOnTable && table === throwOnTable.table) {
+          throw throwOnTable.error;
+        }
+        if (table === patientsTable) return patientRows;
+        // Every other table returns empty — a minimal patient export.
+        return [];
+      },
+    }),
+  };
+}
+
+const transactionMock = vi.fn(
+  async (cb: (tx: { insert: typeof insertMock }) => Promise<unknown>) => {
+    return cb({ insert: insertMock });
+  },
+);
+
+vi.mock("@carebridge/db-schema", () => ({
+  getDb: () => ({
+    insert: insertMock,
+    transaction: transactionMock,
+    select: selectMock,
+  }),
+  fhirResources: fhirResourcesTable,
+  auditLog: auditLogTable,
+  patients: patientsTable,
+  vitals: vitalsTable,
+  labPanels: labPanelsTable,
+  labResults: labResultsTable,
+  medications: medicationsTable,
+  diagnoses: diagnosesTable,
+  allergies: allergiesTable,
+}));
+
+// drizzle-orm's `eq` is only used to build where-clauses; our select mock
+// ignores the predicate and returns rows based solely on the table, so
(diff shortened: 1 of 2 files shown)


blamechris:
## Summary
PR #503 adds a pre-flight `audit_log` insert to `exportPatient` and decorates the returned FHIR Bundle with a non-standard `meta` block (`export_id`, `exported_at`, `exported_by`, `recommended_purge_at` = +15 min). Intent — guaranteeing an audit trail even if delivery fails — is correct and aligned with HIPAA § 164.312(b). Execution has a few sharp edges worth addressing before merge.

## Strengths
- Audit-before-return ordering closes the real gap: even a crash mid-bundle assembly now leaves a paper trail. This matches the pattern already used for `fhir_import` in the same file (PR #378).
- Inline docstring is unusually honest — it names the limitations (no signed URL, no `Cache-Control`, advisory purge) and points at #290 as the real fix.
- Self-contained change, 44 additions, single file, no schema churn.

## Issues

### 1. `meta` block violates FHIR R4 Bundle schema (high)
`services/fhir-gateway/src/types/fhir-r4.ts:48-52` defines `Meta` as `{ profile?, lastUpdated?, versionId? }`. The new fields (`export_id`, `exported_at`, `recommended_purge_at`, `exported_by`) are non-spec. Strict FHIR consumers (HAPI validator, IBM FHIR Server, Smile CDR) will either reject the bundle or strip the fields silently. Two safer options:
- Move the metadata outside `meta` — e.g., `bundle.extension[]` with a CareBridge-owned URL (the FHIR escape hatch for custom data), or a sibling `_carebridge` field at the response root (outside the Bundle contract).
- Keep `meta` but restrict it to spec fields (`lastUpdated: exportedAt`), and move the custom keys to an extension.

### 2. `success: true` recorded before the bundle is actually built (high)
The audit insert at lines 229-248 hardcodes `http_status_code: 200, success: true` and runs before the patient-existence check (line 233) and all the data fetches / FHIR conversions (lines 240-309). If any of those throw — NOT_FOUND on a bad patientId, a DB error in the labs loop, a generator crash — the audit row still says "success: 200" and no bundle was delivered. Two fixes worth considering:
- Split into two rows: one `action: "fhir_export_started"` at entry (no success field, or `success: null`), and a second `action: "fhir_export_completed"` after assembly succeeds. Auditors can diff them to find attempts that never delivered.
- Or wrap the whole handler in try/catch and only set `success: true` on the happy path, with a `success: false` row on the failure branch.

The current shape is strictly worse than the pre-PR state for forensic accuracy: an attacker who triggers an error mid-fetch gets a falsely-green audit row.

### 3. `action: "export"` is not in the existing vocabulary (medium)
Grepping `action:` across services shows the existing vocabulary: `read / create / update / delete` (middleware/audit.ts), `fhir_import`, `session_evicted / session_logout / session_refreshed / session_idle_expired / session_rejected_inactive`, `password_changed`, `emergency_access / emergency_access_used`, `access_denied`, `family_invite_*`, `family_access_revoked`. The bare token `"export"` is new and collides with the generic verb space. `fhir_export` would mirror the existing `fhir_import` and makes `resource_type: "fhir_bundle"` redundant-but-consistent.

### 4. No test coverage for the audit insert (medium)
`router-auth.test.ts` covers authz on `exportPatient` but nothing asserts the audit row is written, contains the right `action` / `resource_type` / `details`, or that it precedes the bundle return. The import-bundle path has `import-bundle-audit.test.ts` as precedent. Adding a mirror `export-bundle-audit.test.ts` would prevent regressions and pin down the `action` name before it leaks into dashboards.

### 5. `recommended_purge_at` is advisory only (known, low)
No TTL enforcement, no `Cache-Control: no-store`, no `Content-Disposition: attachment`, and tRPC responses are JSON-over-HTTP so any intermediary can cache them. The docstring says this and points at #290. Acceptable as scoped for this PR, but worth an explicit follow-up issue so #290 doesn't swallow it.

## Deferred Items
- Signed short-TTL URL delivery layer (#290) — out of scope, requires new HTTP surface outside tRPC.
- `Cache-Control: no-store` / `Content-Disposition: attachment` response headers — depends on the delivery layer change.

## Verdict
**Request changes.** The audit-before-build ordering is the right instinct but the hardcoded `success:true` makes the audit trail less trustworthy than before the PR, not more. Combined with the non-standard `meta` fields silently breaking FHIR validators, both issues deserve a fix before merge. Issues 3 and 4 are quick, mechanical cleanups. Filing follow-up issues for each.

blamechris:
Addressed in 5efb8ac1e75b35f5d66e061f30d260c40a9e5b23:

- **#545 (meta violates FHIR R4)**: moved `export_id`/`exported_at`/`recommended_purge_at`/`exported_by` off primitive `Bundle.meta` keys and onto `Bundle.meta.extension[]` with URL `https://carebridge.dev/fhir/StructureDefinition/export-meta` (valueString JSON). Meta now also carries spec-compliant `lastUpdated`. Strict FHIR validators ignore unknown extensions but were rejecting the old primitive keys.
- **#547 (success:true before build)**: audit insert moved after bundle assembly; whole handler wrapped in try/catch. NOT_FOUND writes `success=false`/404 before throwing; mid-fetch errors write `success=false`/500 via the outer catch with `error_message` populated. Happy path writes `success=true`/200 after the last entry is pushed.
- **#555 (action verb)**: `action: "export"` renamed to `action: "fhir_export"` to mirror `fhir_import`. Also set `procedure_name: "fhir.exportPatient"` for consistency.
- **#556 (test coverage)**: added `services/fhir-gateway/src/__tests__/export-bundle-audit.test.ts` mirroring `import-bundle-audit.test.ts` — three cases covering happy path, patient-not-found (404), and mid-fetch DB error (500). All 70 fhir-gateway tests pass; typecheck clean.

PR #504, opened 2026-04-16
fix(notifications): strip PHI from push-delivery notification body

### LINKED ISSUE #289: Notification payloads risk lock-screen PHI disclosure
**Files:** services/ai-oversight/src/rules/critical-values.ts:40,123; services/notifications/src/router.ts:40

Generated notification text contains specific values ('Critically high heart_rate: 145 bpm', 'Potassium = 7.2 mmol/L'). When push notifications are wired up these will appear on locked devices.

**Fix:** Two-tier payload: short generic title for OS-level display, detail behind authenticated fetch; never include patient name or numeric values in payload.

blamechris:
## Summary
Two-tier notification delivery to prevent HIPAA-disclosive lock-screen previews:

- **Persisted row** (\`notifications.body\`): full flag summary, rendered only inside the authenticated portal.
- **Pushed payload** (\`publishNotification(body)\`): generic category-only cue from \`buildSafeSummary\` — no numeric values, no patient identifiers.

Title stays as the already-safe severity + category form. Full clinical detail is keyed off \`related_flag_id\` for post-unlock fetch.

## Why
Flag summaries like \"Potassium = 7.2 mmol/L\" and \"Critically high heart_rate: 145 bpm\" were flowing straight into the SSE publish payload — and into any future APNs/FCM push integration. Locked-device preview of that text is a HIPAA disclosure.

## Test plan
- [x] \`pnpm --filter @carebridge/notifications test\` — 43/43 passing
- [ ] Manual: trigger critical flag, confirm SSE payload \`body\` is generic cue, confirm portal still shows full summary

Closes #289

### DIFF
--- a/services/notifications/src/__tests__/dispatch-worker.test.ts
+++ b/services/notifications/src/__tests__/dispatch-worker.test.ts
@@ -244,4 +244,97 @@ describe("dispatch-worker", () => {
     expect(insertedRecords).toHaveLength(2);
     expect(mockPublishNotification).toHaveBeenCalledTimes(2);
   });
+
+  // ── PHI lock-screen safety (issue #289) ────────────────────────────
+  //
+  // These tests lock in the two-tier split: the Redis pub/sub payload
+  // (what any future APNs/FCM integration hands to the OS for lock-screen
+  // render) must never contain PHI, while the persisted notification row
+  // must still carry the full summary for the authenticated portal fetch.
+
+  describe("PHI lock-screen safety", () => {
+    const phiEvent = (): NotificationEvent =>
+      makeEvent({
+        severity: "critical",
+        category: "critical-value",
+        summary: "Potassium = 7.2 mmol/L for MRN 123456, BP 145/95",
+      });
+
+    it("published body contains no numeric values", async () => {
+      mockSelectResult.push({ user_id: "user-neuro" });
+      mockSelectResult2.push({
+        id: "user-neuro",
+        specialty: "Neurology",
+        role: "physician",
+      });
+
+      const event = phiEvent();
+      await processorFn({ data: event, id: "job-phi-1" });
+
+      expect(mockPublishNotification).toHaveBeenCalledTimes(1);
+      const [, payload] = mockPublishNotification.mock.calls[0];
+      expect(payload.body).not.toMatch(/\d/);
+    });
+
+    it("published body does not contain the raw event.summary text", async () => {
+      mockSelectResult.push({ user_id: "user-neuro" });
+      mockSelectResult2.push({
+        id: "user-neuro",
+        specialty: "Neurology",
+        role: "physician",
+      });
+
+      const event = phiEvent();
+      await processorFn({ data: event, id: "job-phi-2" });
+
+      const [, payload] = mockPublishNotification.mock.calls[0];
+      expect(payload.body).not.toContain(event.summary);
+      // And individual PHI tokens from the summary must not leak.
+      expect(payload.body).not.toContain("Potassium");
+      expect(payload.body).not.toContain("MRN");
+      expect(payload.body).not.toContain("7.2");
+      expect(payload.body).not.toContain("145/95");
+    });
+
+    it("persisted record.summary_safe equals the buildSafeSummary output", async () => {
+      mockSelectResult.push({ user_id: "user-neuro" });
+      mockSelectResult2.push({
+        id: "user-neuro",
+        specialty: "Neurology",
+        role: "physician",
+      });
+
+      const event = phiEvent();
+      await processorFn({ data: event, id: "job-phi-3" });
+
+      // Reproduce buildSafeSummary locally (mirror of the private helper):
+      // a category-only template, PHI-free by construction.
+      const expectedSafe =
+        `Clinical flag — ${event.category.replace(/-/g, " ")}. ` +
+        `Open the portal to view details.`;
+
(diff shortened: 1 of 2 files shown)


blamechris:
# /agent-review — PR #504

## Summary
Two-tier split looks correct in shape: persisted `notifications.body` keeps the full summary for authenticated portal render, and the Redis pub/sub payload now gets a generic category-only cue from `buildSafeSummary`. Scope is tight (one file, ~40 lines), and addresses issue #289 directly. Confirmed: `notifications.body` and `notifications.title` are both `encryptedText` in schema (`packages/db-schema/src/schema/notifications.ts`) and migration `0016_notification_phi_encryption.sql`, so persisted PHI is encrypted at rest via AES-256-GCM. DB side is sound.

## Strengths
- Correct split of trust tiers: push layer = generic, authenticated fetch = full detail.
- Both the immediate and delayed (quiet-hours) dispatch paths were updated — easy to miss the second one.
- Docstrings explicitly flag the HIPAA rationale so the next editor won't silently re-introduce PHI into the pushed body.
- Doesn't touch the persisted row shape, so no migration coupling.
- 43/43 existing tests still pass — no regression in existing behavior.

## Issues

### 1. `summary_safe` column already exists in schema but is unused (correctness / architectural drift)
The DB schema already ships a dedicated `summary_safe TEXT` column (schema line 11, migration `0016_notification_phi_encryption.sql` line 12) expressly created for this purpose. The router's `create` procedure even accepts it (`services/notifications/src/router.ts:40`). This PR computes a safe summary but throws it away instead of persisting it in `summary_safe`. Two consequences:
- `summary_safe` remains permanently NULL in practice, which will confuse future readers (why does the column exist?).
- A future push integration (APNs/FCM) that reads from the DB row (e.g., re-dispatching a missed notification) has to re-derive the safe summary instead of reading the canonical column.
Recommend: set `summary_safe: safeSummary` on both `immediateRecords` and the delayed `record`, and publish `record.summary_safe` rather than re-computing.

### 2. No tests lock in the PHI-safety behavior (test coverage gap)
`buildSafeSummary` is introduced without any direct assertion. The existing `dispatch-worker.test.ts` only asserts `is_urgent`, `related_flag_id`, and `type` on the published payload (`src/__tests__/dispatch-worker.test.ts:223-228`). There is no test that:
- The published `body` does NOT contain the event.summary text (e.g., "Potassium = 7.2").
- The published `body` contains no digits (`/\d/` should not match).
- Title is also digit-free.
Without these, a future "make the push richer" refactor will silently re-introduce PHI without breaking any red test.
Recommend: add an assertion that `mockPublishNotification.mock.calls[0][1].body` does not include `event.summary`, does not match `/\d/`, and the title likewise does not match `/\d/`.

### 3. Title safety is claimed but not enforced (defense in depth)
The PR description says "Title stays as the already-safe severity + category form." The docstring added to `buildNotificationTitle` asserts this as an invariant — but nothing enforces it. `event.category` is interpolated directly; today all categories are known-safe enum values (`cross-specialty`, `medication-safety`, `critical-value`, `patient-reported`, `care-gap`, `drug-interaction`), but nothing prevents a future rule author from passing a free-form category like `psychiatric-symptoms-self-harm`. Low likelihood, non-zero impact on a lock screen.
Recommend (small): either type `NotificationEvent.category` as the narrowed union from `@carebridge/types`, or add a whitelist check inside `buildNotificationTitle` that falls back to `"clinical alert"` for unknown categories.

### 4. Minor: `buildSafeSummary` docstring describes logic that isn't implemented
The docstring advertises "We strip: every number (integers, decimals, ratios) → "…", common clinical identifiers (MRN-like strings)" but the function body doesn't do that — it just ignores `event.summary` entirely and returns a template string. The comment-code mismatch will mislead the next reader. Either implement the strip logic (overkill if the template approach is sufficient) or trim the docstring to match what the code actually does: "returns a template string built from category alone; event.summary is intentionally discarded."

## Deferred Items
- APNs/FCM integration itself is not in scope here; the safety contract is prepared for it but not exercised end-to-end.
- `summary_safe` backfill for existing notification rows (if/when #1 is fixed) would be a separate migration.

## Verdict
**Request changes (minor)** — shape is right and ships a real HIPAA fix, but:
- persisting `summary_safe` is low-cost and uses infrastructure already in place,
- at minimum a regression test locking in "body contains no PHI and no digits" should land with this PR.

Issue #1 and Issue #2 are the load-bearing ones. Issue #3 and #4 are polish.

blamechris:
Addressed in 1c7a498:

- **#548 (persist summary_safe)** — `processNotificationJob` and `processDelayedNotification` now set `summary_safe: safeSummary` on the inserted row; the publish call reads `record.summary_safe` rather than re-computing.
- **#549 (PHI-safety tests)** — Added a `PHI lock-screen safety` describe block with 4 assertions: (a) published `body` matches no `/\d/`, (b) published `body` does not contain `event.summary` or its PHI tokens (`Potassium`, `MRN`, `7.2`, `145/95`), (c) persisted `record.summary_safe` equals the `buildSafeSummary` output (and cross-checks the payload body), (d) persisted `record.body` still equals `event.summary` for authenticated fetch. 47/47 tests pass.
- **#553 (docstring alignment)** — `buildSafeSummary` docstring simplified to describe the actual category-only template behavior, calling out that `event.summary` is intentionally discarded and that `summary_safe` is persisted for later fetch.

**#551 (category whitelist)** left open — it is a separate defense-in-depth hardening and not load-bearing for this PR's scope.

PR #505, opened 2026-04-16
fix(portal): render lab reference range + infer H/L for unflagged out-of-range

### LINKED ISSUE #262: clinical-safety(p3): lab reference range not displayed in UI — borderline abnormal values appear normal
## Problem
`apps/clinician-portal/app/patients/[id]/page.tsx` (lines 196-235), `LabsTab`:

The table attempts to display `r.reference_range as string`, but `LabResult` objects have `reference_low?: number` and `reference_high?: number` fields — not a pre-formatted `reference_range` string. The display renders `undefined` for every lab reference range.

Additionally, values that are outside the reference range but NOT flagged as "critical" are displayed in default text color — indistinguishable from normal values visually.

## Clinical Consequence
- Clinician reviews lab results
- No reference ranges visible (undefined in every row)
- K+ 3.2 mEq/L (low-normal, below reference 3.5-5.0) displayed identically to K+ 4.0 mEq/L
- Clinician misses a trend of gradually declining potassium
- No visual cue for "H" (high) or "L" (low) flags on borderline values

### 1. Format reference range from separate fields:
```tsx
<td>
  {r.reference_low != null && r.reference_high != null
    ? `${r.reference_low}–${r.reference_high}`
    : "—"}
</td>
```

### 2. Add visual distinction for any out-of-range value:
```tsx
const isAbnormal = 
  (r.reference_low != null && r.value < r.reference_low) ||
  (r.reference_high != null && r.value > r.reference_high);

<td style={{ 
  color: r.flag === "critical" ? "var(--critical)" 
       : r.flag === "high" ? "var(--warning)"
       : isAbnormal ? "var(--warning)"
       : "inherit"
}}>
```

blamechris:
## Summary
- Reference-range cell now formats from \`reference_low\` / \`reference_high\` (the actual LabResult shape). Previously read \`r.reference_range\` which is undefined.
- Value text color reflects any out-of-range condition, not just the server \`flag\` field — catches borderline-abnormal values the instrument didn't mark.
- Client-inferred L/H badge surfaces for out-of-range values with no server flag, clearly labeled as client-derived.

## Why
A K+ of 3.2 against a 3.5–5.0 range rendered in the default color with \`undefined\` for the reference range. Clinicians could miss gradually declining potassium because there was no visual cue for \"borderline abnormal.\"

## Test plan
- [x] \`pnpm typecheck\` / \`pnpm lint\` — clean
- [ ] Manual: load chart with lab panel → reference range renders (e.g. \`3.5–5.0\`), value in warning color when out-of-range, inferred badge shown

Closes #262

### DIFF
--- a/apps/clinician-portal/app/patients/[id]/page.tsx
+++ b/apps/clinician-portal/app/patients/[id]/page.tsx
@@ -223,18 +223,56 @@ function LabsTab({ patientId }: { patientId: string }) {
               <tbody>
                 {panel.results.map((r: Record<string, unknown>, ri: number) => {
                   const flag = (r.flag as string) ?? "";
+                  const value = r.value as number | null | undefined;
+                  const refLow = r.reference_low as number | null | undefined;
+                  const refHigh = r.reference_high as number | null | undefined;
+
+                  // A value is out-of-range when it falls below reference_low or
+                  // above reference_high. This catches "borderline abnormal"
+                  // cases (e.g. K+ 3.2 with ref 3.5–5.0) that the server-side
+                  // `flag` field often leaves unset — the lab instrument may
+                  // only assign "critical" for extreme values, leaving
+                  // clinically meaningful drift unmarked.
+                  const isOutOfRange =
+                    typeof value === "number" &&
+                    ((typeof refLow === "number" && value < refLow) ||
+                      (typeof refHigh === "number" && value > refHigh));
+
+                  const valueColor =
+                    flag === "critical"
+                      ? "var(--critical)"
+                      : flag === "high" || flag === "low"
+                      ? "var(--warning)"
+                      : isOutOfRange
+                      ? "var(--warning)"
+                      : "var(--text-primary)";
+
+                  const referenceRange =
+                    typeof refLow === "number" && typeof refHigh === "number"
+                      ? `${refLow}\u2013${refHigh}`
+                      : typeof refLow === "number"
+                      ? `> ${refLow}`
+                      : typeof refHigh === "number"
+                      ? `< ${refHigh}`
+                      : "\u2014";
+
+                  // Out-of-range without a server flag — surface an inferred
+                  // H/L badge so the clinician has the same visual cue they
+                  // would for an instrument-flagged value.
+                  const inferredFlag =
+                    !flag && isOutOfRange && typeof value === "number"
+                      ? typeof refLow === "number" && value < refLow
+                        ? "low"
+                        : "high"
+                      : "";
+
                   return (
                     <tr key={ri}>
                       <td>{r.test_name as string}</td>
                       <td
                         style={{
                           fontWeight: 600,
-                          color:
-                            flag === "critical"
-                              ? "var(--critical)"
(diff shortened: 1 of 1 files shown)


blamechris:
## Agent Review — PR #505

### Summary
Fixes a real clinical-safety bug: the Labs tab previously read `r.reference_range` (which doesn't exist on `LabResult`) so the reference column rendered `undefined`, and out-of-range values that the instrument didn't flag rendered in default color. The change formats from `reference_low`/`reference_high`, colors value text on any out-of-range condition, and surfaces an inferred H/L badge (clearly labeled as client-derived) for unflagged out-of-range values.

### Strengths
- Addresses the exact failure mode in the issue: borderline-abnormal K+ 3.2 against a 3.5–5.0 reference is now both formatted and visually surfaced.
- Inferred badge uses `role="status"` and a descriptive `aria-label` explicitly naming it as "Inferred" and "out of reference range" — good accessibility baseline.
- Reference-range formatting handles all four cases (both bounds, low-only, high-only, neither) with sensible fallbacks.
- Color logic is layered defensively: server `critical` still wins over `high`/`low` which still wins over inferred range.
- Narrow, surgical diff to a single file; no API or schema changes.

### Issues (non-blocking)
1. **Type tightening (minor).** The row is cast `r as Record<string, unknown>` and then each field is cast individually (`r.value as number | null | undefined`, etc.). A shared `LabResult` type imported from `@carebridge/types` (or the tRPC router's inferred output) would eliminate the per-field casts and catch future schema drift at compile time — which is how this bug slipped in originally.
2. **Inferred-vs-server visual parity.** The inferred badge reuses `badge-warning` and differs from a server `low`/`high` badge only via the `title` attribute (hover) and the `aria-label`. On mobile/touch there is no hover, so a sighted touch user sees no distinction between instrument-flagged and client-derived flags. Consider a subtle visual marker (asterisk, dashed border, or an explicit "(inferred)" suffix for screen width >= some breakpoint).
3. **No component test.** The new branching (range formatting, inferred-flag derivation, color precedence) is pure render logic and highly testable. A component/unit test asserting the K+ 3.2 scenario would lock in the fix against regression. The PR body also lists the manual test checkbox as unchecked.
4. **Future conflict risk with server flags.** If/when the server starts emitting `flag="low"` for borderline values (e.g. after the instrument-config work in the AI oversight pipeline), the client inference becomes redundant but still fires because it only short-circuits on *truthy* `flag`. That's currently correct, but worth a comment noting the precedence assumption so a future contributor doesn't "simplify" by removing the inference.
5. **En-dash as literal escape.** `\u2013` and `\u2014` are inlined. Fine for now, but a `REFERENCE_RANGE_SEPARATOR` constant (or an `Intl`-aware formatter) would centralize i18n later — non-blocking.

### Deferred Items
All five issues above are non-blocking and tracked as follow-up issues with the "from-review" label (see below). The PR fixes a live UI bug and the follow-ups are quality/hardening, not correctness.

### Verdict
**APPROVE.** Ship it — the fix is correct, narrowly scoped, and directly resolves a clinical-safety display bug. Address follow-ups in separate PRs.

PR #560, opened 2026-04-16
perf(db): add btree index on review_jobs.trigger_event_id

### LINKED ISSUE #527: perf(db): add review_jobs(trigger_event_id) index for idempotency probe
Identified during review of PR #495.

## Problem
PR #495 added a pre-insert idempotency probe on `review_jobs`:

\`\`\`ts
db.select({ id, status })
  .from(reviewJobs)
  .where(and(
    eq(reviewJobs.trigger_event_id, event.id),
    eq(reviewJobs.status, 'completed'),
  ))
  .limit(1);
\`\`\`

The only existing index on the table is `idx_review_jobs_patient(patient_id, status)` (see `packages/db-schema/src/schema/ai-flags.ts:82`). Nothing supports a lookup by `trigger_event_id`, so Postgres falls back to a sequential scan on every single review job — a cost that grows linearly with the table.

Today `review_jobs` is small and the cost is invisible. At our target throughput and the 7-year retention window implied by the HIPAA audit policy, this table is append-only and unbounded; the probe's latency will degrade linearly forever without an index.

## Proposed fix
Add a Drizzle migration creating a btree index on `(trigger_event_id, status)` (or just `trigger_event_id` — both satisfy the probe; the two-column form also serves the partial index pattern if we ever add more status filters). Example:

\`\`\`sql
CREATE INDEX IF NOT EXISTS idx_review_jobs_trigger_event
  ON review_jobs (trigger_event_id, status);
\`\`\`

Update `ai-flags.ts` schema accordingly:

\`\`\`ts
index('idx_review_jobs_trigger_event').on(table.trigger_event_id, table.status),
\`\`\`

## Acceptance criteria
- New migration file created and applied
- Drizzle schema updated to match
- `EXPLAIN` on the probe query shows `Index Scan` (or `Index Only Scan`), not `Seq Scan`

blamechris:
Closes #527

## Problem
PR #495 added a pre-insert idempotency probe in `services/ai-oversight/src/services/review-service.ts`:

```ts
db.select({ id: reviewJobs.id, status: reviewJobs.status })
  .from(reviewJobs)
  .where(and(
    eq(reviewJobs.trigger_event_id, event.id),
    eq(reviewJobs.status, "completed"),
  ))
  .limit(1);
```

The only existing index on `review_jobs` is `idx_review_jobs_patient(patient_id, status)` — nothing supports a lookup by `trigger_event_id`, so Postgres falls back to a sequential scan on every probe. `review_jobs` is append-only and bounded by the 7-year HIPAA retention window (`docs/hipaa-retention.md`), so the cost grows linearly forever.

## Fix
Composite btree on `(trigger_event_id, status)`:

```sql
CREATE INDEX IF NOT EXISTS idx_review_jobs_trigger_event
  ON review_jobs (trigger_event_id, status);
```

- Covers the equality-equality probe directly.
- Satisfies any future `WHERE trigger_event_id = $1`-only queries via the btree leftmost-prefix rule.
- Column is singular `text` (not a jsonb array), so a btree applies; a GIN index would not.

## Files
- `packages/db-schema/drizzle/0033_review_jobs_trigger_event_index.sql` — migration.
- `packages/db-schema/src/schema/ai-flags.ts` — declares `idx_review_jobs_trigger_event` alongside the existing `idx_review_jobs_patient` so the Drizzle generator stays in sync.
- `packages/db-schema/drizzle/meta/_journal.json` — registers the new entry.

## Note on `CREATE INDEX CONCURRENTLY`
Drizzle's postgres-js migrator wraps each migration in a transaction, and `CREATE INDEX CONCURRENTLY` is not transaction-safe. No existing migration in this repo uses `CONCURRENTLY` — we match that convention and use a plain `CREATE INDEX IF NOT EXISTS`. The `review_jobs` table is small today; if we ever need to add this against a large live table, it should be done out-of-band (psql session, outside migrations) and then registered as a no-op `IF NOT EXISTS` here.

## Verification
- `pnpm --filter @carebridge/db-schema typecheck` — pass.
- `pnpm typecheck` across the monorepo — pass.
- `pnpm lint` — pass (no new warnings).
- `pnpm --filter @carebridge/ai-oversight test -- review-idempotency` — both idempotency tests still pass.
- `EXPLAIN` verification on a populated dev DB to confirm `Index Scan` over `Seq Scan` is deferred to post-merge (requires DB with rows).

### DIFF
--- a/packages/db-schema/drizzle/0033_review_jobs_trigger_event_index.sql
+++ b/packages/db-schema/drizzle/0033_review_jobs_trigger_event_index.sql
@@ -1,0 +1,25 @@
+-- Migration: Index review_jobs.(trigger_event_id, status) for the idempotency probe.
+--
+-- PR #495 added a pre-insert check in services/ai-oversight/src/services/
+-- review-service.ts that asks "does a completed review_jobs row already
+-- exist for this trigger_event_id?" to suppress duplicate reviews when
+-- BullMQ redelivers a job (worker crash / stalled-scan reclaim) or the
+-- outbox reconciler replays the same event:
+--
+--   SELECT id, status FROM review_jobs
+--   WHERE trigger_event_id = $1 AND status = 'completed'
+--   LIMIT 1;
+--
+-- The only pre-existing index on review_jobs is
+-- idx_review_jobs_patient(patient_id, status), which does not support
+-- this lookup — so every probe falls back to a sequential scan. The
+-- table is append-only and bound by the 7-year HIPAA retention window
+-- (see docs/hipaa-retention.md), so the per-probe cost grows linearly
+-- forever without an index.
+--
+-- A composite btree on (trigger_event_id, status) supports both the
+-- equality-then-equality probe and any future queries that filter by
+-- trigger_event_id alone (btree index leftmost-prefix rule).
+
+CREATE INDEX IF NOT EXISTS idx_review_jobs_trigger_event
+  ON review_jobs (trigger_event_id, status);
--- a/packages/db-schema/drizzle/meta/_journal.json
+++ b/packages/db-schema/drizzle/meta/_journal.json
@@ -239,6 +239,13 @@
       "when": 1746547200000,
       "tag": "0031_audit_log_actor_relationship",
       "breakpoints": true
+    },
+    {
+      "idx": 34,
+      "version": "7",
+      "when": 1746720000000,
+      "tag": "0033_review_jobs_trigger_event_index",
+      "breakpoints": true
     }
   ]
 }
--- a/packages/db-schema/drizzle/meta/_journal.json
+++ b/packages/db-schema/drizzle/meta/_journal.json
@@ -243,6 +243,13 @@
     {
       "idx": 34,
       "version": "7",
+      "when": 1746633600000,
+      "tag": "0032_review_jobs_rules_output",
+      "breakpoints": true
+    },
+    {
+      "idx": 35,
+      "version": "7",
       "when": 1746720000000,
       "tag": "0033_review_jobs_trigger_event_index",
       "breakpoints": true
--- a/packages/db-schema/src/schema/ai-flags.ts
+++ b/packages/db-schema/src/schema/ai-flags.ts
@@ -86,4 +86,11 @@ export const reviewJobs = pgTable("review_jobs", {
   created_at: text("created_at").notNull(),
 }, (table) => [
   index("idx_review_jobs_patient").on(table.patient_id, table.status),
+  // Supports the pre-insert idempotency probe in review-service.ts which
+  // filters by (trigger_event_id, status='completed') to suppress duplicate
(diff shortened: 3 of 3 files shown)


blamechris:
## Review — PR #560: `perf(db): add btree index on review_jobs.trigger_event_id`

Clean, well-scoped performance fix. The composite `(trigger_event_id, status)` index directly covers the idempotency probe added in PR #495 and the PR description is thorough. One sequencing issue in the migration journal needs to be addressed before merge.

---

### Strengths
- Index column order `(trigger_event_id, status)` matches the query predicate exactly — optimal for the equality-equality lookup and satisfies future `trigger_event_id`-only queries via the leftmost-prefix rule.
- Excellent SQL migration comments linking back to the originating query and explaining the growth model (append-only, 7-year HIPAA retention).
- Thoughtful `CONCURRENTLY` note documenting why the repo avoids it inside Drizzle-managed transactions and what to do if a live backfill is ever needed.
- Schema declaration kept in sync with the SQL migration.

---

### Issues Found

#### Critical

| # | File | Issue | Suggested Fix |
|---|------|-------|---------------|
| 1 | `drizzle/meta/_journal.json` | **Journal sequence gap**: the new entry uses `idx: 34`, but the current last entry is `idx: 33` (`0031_audit_log_actor_relationship`). Migration `0032_review_jobs_rules_output.sql` exists on disk (landed in PR #501) but has **no journal entry** — Drizzle's migrator will never execute it, and this PR would widen the gap to two missing entries. | Add the missing `0032` journal entry at `idx: 34` and bump this migration to `idx: 35`. Better yet, open a prerequisite fix PR that registers `0032` in the journal (see follow-up issue below). |

#### Suggestions

| # | File | Suggestion |
|---|------|------------|
| 1 | `ai-flags.ts` | Consider adding a brief `@see` or inline note on the `reviewJobs` table pointing to the HIPAA retention window doc, since the PR description notes the table is retention-bounded — would help future readers evaluating index cost. |

#### Nitpicks

| # | File | Note |
|---|------|------|
| 1 | `0033_*.sql` | Minor: migration filename says `0033` but the journal `idx` is `34`. The existing repo already has tag/idx drift (e.g., `idx=7, tag=0008`), so this is consistent with convention — just noting it. |

---

### Architecture Notes
- The index is narrowly targeted and adds negligible write overhead (two `text` columns, one of which is a short enum-like status). Net positive for the hot idempotency path.
- The `IF NOT EXISTS` guard is a good defensive pattern for idempotent re-runs.
- No HIPAA concerns — this is a structural change with no data access or PHI exposure.

---

The index itself is correct and well-justified. The blocking issue is the orphaned `0032` journal entry — merging this PR would leave two migrations on disk that Drizzle's migrator will silently skip, which is a data-integrity risk for fresh deployments and CI `db:migrate` runs. Once the journal is repaired (either in this PR or a predecessor), this is a clean approve.

PR #561, opened 2026-04-16
fix(ai): ONCO-VTE-NEURO-001 ignores resolved or stale DVTs

### LINKED ISSUE #215: clinical-safety(p2): ONCO-VTE-NEURO-001 false positive on old resolved DVT — unnecessary CT/neuroimaging
## Problem
`services/ai-oversight/src/rules/cross-specialty.ts` (lines 46-90): the `hasVTE` check matches ANY diagnosis containing "dvt" or "thrombosis" without checking if it is active or historical.

```typescript
const hasVTE = ctx.active_diagnoses.some((d) =>
  /dvt|deep vein thrombosis|pulmonary embolism|vte|thrombosis|clot/i.test(d),
);
```

This fires for patients with a **resolved DVT from years ago** that remains listed as a historical diagnosis.

**Scenario:**
- 65-year-old cancer patient with history of DVT in 2018 (resolved, off anticoagulation since 2019)
- Patient presents with tension headache
- Rule fires: "Cancer + VTE + neuro symptom → urgent CT head / CT angiography"
- Patient receives CT with IV contrast → contrast nephropathy risk in an elderly patient
- Negative scan; workup was unnecessary

At scale (many cancer patients with old VTE history), this generates numerous unnecessary neuro workups.

## Fix
1. Cross-reference `active_diagnosis_codes` against ICD-10 VTE codes (I26, I80, I82) AND check recency
2. Require active anticoagulation medication alongside the VTE diagnosis as a proxy for active disease
3. Require more specific neuro symptoms alongside headache (vision changes, focal weakness, speech difficulty) for maximum specificity

```typescript
const hasActiveAnticoag = ctx.active_medications.some(m => ANTICOAGULANT_PATTERN.test(m));
const hasVTE = (hasVTEByCode || hasVTEByDescription) && hasActiveAnticoag;
```

blamechris:
Closes #215

## Problem
`ONCO-VTE-NEURO-001` (cancer + VTE + new neuro symptom → urgent CT head / CTA) matched any diagnosis containing "dvt" / "thrombosis" regardless of whether the VTE was active, years-old, or resolved. A 65-year-old oncology patient with a 2018 DVT (resolved, off anticoagulation since 2019) presenting with a tension headache got an IV-contrast CT head / CTA workup — contrast nephropathy risk with zero clinical benefit. At scale this drives a lot of unnecessary neuroimaging.

## Fix — recency gate on the VTE diagnosis
New helper `isActiveVTEDiagnosis` in `services/ai-oversight/src/rules/cross-specialty.ts`. A VTE diagnosis qualifies the rule only when:

1. It matches VTE by ICD-10 (`I26` / `I80` / `I82.*`) or description, **and**
2. It is **not** explicitly resolved — `status === "resolved"` OR any past-dated `resolved_date` disqualifies it (EHR problem lists often leave `status="active"` on resolved problems, so `resolved_date` wins), **and**
3. **Either** onset is within the last **6 months** (`VTE_RECENCY_WINDOW_MONTHS`) **or** the patient is on active anticoagulation (proxy for ongoing VTE treatment).

### Why 6 months
Cancer-associated VTE is typically anticoagulated for 3–6 months per CHEST and ASCO guidance. Beyond 6 months without ongoing anticoagulation the acute pro-thrombotic contribution of that specific clot has largely resolved; patients with persistent risk either remain on anticoagulants (caught by the anticoag branch) or have a newer VTE episode (caught by the onset branch).

## Shape change
- `PatientContext` gains an optional `active_diagnoses_detail?: PatientDiagnosis[]` carrying `{description, icd10_code, status, onset_date, resolved_date}` per diagnosis. The rule uses this when present and falls back to the legacy flat-list behavior when it's absent, so no existing caller or test breaks.
- `buildPatientContextForRules` now populates `active_diagnoses_detail` from the `diagnoses` table, so production callers get the recency gate automatically.

## Test coverage (TDD)
Added 5 new cases to `cross-specialty-rules.test.ts` (all started red, now green):
- Resolved DVT from 5 years ago → no flag
- `status=active` but onset >6 months ago, off anticoag → no flag
- Recent onset (<6 months), no anticoag → fires (critical)
- Old onset, patient on apixaban → fires (anticoag proxy keeps it active)
- Mixed signal: `status=active` but a past `resolved_date` → no flag (resolved_date wins over stale status)

All 69 tests in `cross-specialty-rules.test.ts` pass; all 307 tests in `@carebridge/ai-oversight` pass. `pnpm typecheck` and `pnpm lint` clean at repo root.

### DIFF
--- a/services/ai-oversight/src/__tests__/cross-specialty-rules.test.ts
+++ b/services/ai-oversight/src/__tests__/cross-specialty-rules.test.ts
@@ -189,6 +189,191 @@ describe("ONCO-VTE-NEURO-001 — Cancer + VTE + neurological symptom", () => {
     expect(flag).toBeDefined();
     expect(flag!.suggested_action).toContain("NOT on anticoagulation");
   });
+
+  // --- Recency gate (issue #215) ---
+  //
+  // When structured diagnosis detail is provided, the rule must treat stale /
+  // resolved DVTs as non-qualifying. Without this gate, cancer patients with
+  // a years-old resolved DVT still listed in their chart trigger unnecessary
+  // CT head / CT angiography with IV contrast whenever they present with any
+  // common neurological complaint (e.g. tension headache).
+
+  it("does NOT fire when the VTE is resolved years ago (status=resolved)", () => {
+    const today = new Date();
+    const sixYearsAgo = new Date(today);
+    sixYearsAgo.setFullYear(today.getFullYear() - 6);
+    const fiveYearsAgo = new Date(today);
+    fiveYearsAgo.setFullYear(today.getFullYear() - 5);
+
+    const ctx: PatientContext = {
+      active_diagnoses: ["Pancreatic cancer", "Deep vein thrombosis"],
+      active_diagnosis_codes: ["C25.9", "I82.401"],
+      active_medications: [],
+      new_symptoms: ["tension headache"],
+      care_team_specialties: ["oncology"],
+      active_diagnoses_detail: [
+        {
+          description: "Pancreatic cancer",
+          icd10_code: "C25.9",
+          status: "active",
+          onset_date: null,
+          resolved_date: null,
+        },
+        {
+          description: "Deep vein thrombosis",
+          icd10_code: "I82.401",
+          status: "resolved",
+          onset_date: sixYearsAgo.toISOString(),
+          resolved_date: fiveYearsAgo.toISOString(),
+        },
+      ],
+    };
+
+    const flags = checkCrossSpecialtyPatterns(ctx);
+    const flag = flags.find((f) => f.rule_id === "ONCO-VTE-NEURO-001");
+    expect(flag).toBeUndefined();
+  });
+
+  it("does NOT fire when VTE onset is >6 months ago and patient is off anticoagulation", () => {
+    const today = new Date();
+    const twoYearsAgo = new Date(today);
+    twoYearsAgo.setFullYear(today.getFullYear() - 2);
+
+    const ctx: PatientContext = {
+      active_diagnoses: ["Pancreatic cancer", "History of DVT"],
+      active_diagnosis_codes: ["C25.9", "I82.401"],
+      active_medications: [], // off anticoagulation
+      new_symptoms: ["headache"],
+      care_team_specialties: ["oncology"],
+      active_diagnoses_detail: [
+        {
+          description: "Pancreatic cancer",
+          icd10_code: "C25.9",
+          status: "active",
+          onset_date: null,
+          resolved_date: null,
+        },
+        {
+          description: "History of DVT",
+          icd10_code: "I82.401",
+          status: "active", // still listed as active in the problem list
+          onset_date: twoYearsAgo.toISOString(),
+          resolved_date: null,
+        },
+      ],
+    };
+
+    const flags = checkCrossSpecialtyPatterns(ctx);
+    const flag = flags.find((f) => f.rule_id === "ONCO-VTE-NEURO-001");
+    expect(flag).toBeUndefined();
+  });
+
+  it("fires when VTE onset is recent (<6 months) even without anticoagulation", () => {
+    const today = new Date();
+    const threeMonthsAgo = new Date(today);
+    threeMonthsAgo.setMonth(today.getMonth() - 3);
+
+    const ctx: PatientContext = {
+      active_diagnoses: ["Pancreatic cancer", "Deep vein thrombosis"],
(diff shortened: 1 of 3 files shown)


blamechris:
## Code Review — PR #561

**fix(ai): ONCO-VTE-NEURO-001 ignores resolved or stale DVTs**

This PR adds a clinically sound recency gate to the DVT/stroke rule, preventing unnecessary CT head/CTA workups for cancer patients whose VTE resolved years ago. The approach—structured diagnosis detail with onset/resolved metadata plus an anticoagulation proxy—is well-reasoned and the fallback to legacy flat-list behavior is clean.

---

### Strengths

- **Clinical reasoning is solid.** The 6-month window aligns with CHEST/ASCO guidance, and using active anticoagulation as a proxy for ongoing VTE treatment is the right call.
- **Backward-compatible shape extension.** `active_diagnoses_detail` is optional with a clean fallback, so no existing callers break.
- **Test coverage is thorough.** 5 new test cases covering resolved, stale, recent, anticoag-proxy, and mixed-signal scenarios. Good use of TDD.
- **`resolved_date` trumps `status`.** Correctly handles the EHR data quality problem where problem lists leave `status=active` on resolved conditions.
- **Well-documented constants.** `VTE_RECENCY_WINDOW_MONTHS` has clear clinical rationale in comments.

---

### Issues Found

#### Critical

None.

#### Suggestions

| S1 | `cross-specialty.ts` `isActiveVTEDiagnosis` | The `now` parameter defaults to `new Date()` (wall-clock time), but the rule context is built as an event-time snapshot via `wasActiveAt(d, eventAt)`. In production, `now` could be minutes/hours after the event. For a VTE onset exactly at the 6-month boundary, clock skew between event time and processing time could produce inconsistent results. | Accept an optional `eventTimestamp` from the caller and use it instead of `new Date()`. The `check` function doesn't currently have access to event time, so this could be deferred — but file an issue to thread event time through `PatientContext` for time-sensitive rules. |
| S2 | `cross-specialty.ts` lines with `VTE_DESCRIPTION_PATTERN` | The new `VTE_DESCRIPTION_PATTERN` constant and the old inline regex in the legacy fallback are now semantically identical (the constant adds `venous thromboembolism`). However, the extraction means a subtle behavioral change in the legacy path: previously the regex was inline, now it uses the shared constant. This is fine, but the PR description doesn't call out the broadening. | No code change needed — just noting for reviewer awareness that the legacy path now also matches "venous thromboembolism" where it didn't before. |
| S3 | `review-service.ts` `buildPatientContextForRules` | `wasActiveAt` already filters out diagnoses with `resolved_date <= eventAt`, so in the production path `active_diagnoses_detail` will never contain a resolved diagnosis. The `resolved_date` guard in `isActiveVTEDiagnosis` is effectively dead code in production (though exercised by direct-call tests). | Consider adding an integration-level test that goes through `buildPatientContextForRules` with a resolved VTE in the DB to confirm the two filtering layers compose correctly, or add a comment explaining the defense-in-depth intent. |

#### Nitpicks

| # | Location | Issue |
|---|----------|-------|
| N1 | `cross-specialty.ts` `isActiveVTEDiagnosis` | `setMonth(cutoff.getMonth() - VTE_RECENCY_WINDOW_MONTHS)` can produce unexpected results for edge-case months (e.g., subtracting 6 months from March 31 gives September 31 → October 1). Consider using a date library or `setMonth` with day clamping if precision at month boundaries matters clinically. |
| N2 | Test file | Test dates are computed relative to `new Date()` which makes tests time-dependent. If the test suite ever runs near a month boundary, the 3-months-ago test could behave differently than expected. Consider using a fixed reference date and injecting `now` into the rule. |

---

### Architecture Notes

- **`PatientContext` extension pattern is good.** Adding optional fields with fallback logic is the right way to enrich rule inputs without breaking existing consumers. This pattern should be documented as the standard approach for future rule enrichments.
- **Rule registry integration:** The change is self-contained within the `CROSS_SPECIALTY_RULES` array entry — no rule-registry wiring changes needed since the rule ID is unchanged. Clean.
- **LLM path alignment:** The `buildPatientContext` in `context-builder.ts` (LLM path) does NOT populate `active_diagnoses_detail`. This is acceptable since the LLM doesn't use the deterministic rule interface, but if future LLM prompt assembly needs diagnosis recency metadata, it should be added there too.

---

### Verdict: **Approve**

Well-structured fix for a real clinical safety issue. The recency gate logic is sound, tests are comprehensive, and backward compatibility is preserved. The suggestions above (especially S1 — event-time threading) are worth tracking but not blocking.

PR #562, opened 2026-04-16
fix(medical-logic): case/whitespace-tolerant lab unit validation

### LINKED ISSUE #538: lab-unit validator: normalize case/whitespace before allow-list compare
## Context
Identified during review of PR #500.

PR #500 added `allowed_units: string[]` on `CommonLabTest` entries and made `validateLabResult` reject unit mismatches as errors for high-stakes tests. The comparison is a literal `Array.prototype.includes(unit)` — exact string, case-sensitive, whitespace-sensitive.

## Problem
Real-world FHIR and HL7 v2 feeds normalize units inconsistently:
- `"mg/dL"` vs `"mg/dl"` vs `"MG/DL"`
- `"mEq/L"` vs `"meq/L"` (UCUM canonical is lowercase — see `services/fhir-gateway/src/generators/observation.ts:11` `UNIT_TO_UCUM` which maps `"mEq/L" → "meq/L"`)
- `" mg/dL"` / `"mg/dL "` / `"mg / dL"` from manual entry or flaky transforms

Under the current exact-string compare, any of these variants against `allowed_units: ["mg/dL"]` produces a validation error — rejecting a legitimate Glucose reading that happens to arrive in lowercase UCUM. This is the inverse of the sentinel event the allow-list was added to prevent.

## Proposed fix
Introduce a single `normalizeUnit(u: string): string` helper in `@carebridge/medical-logic`:
- Trim whitespace
- Collapse internal whitespace around `/`
- Lowercase for the compare only (preserve original in error messages)
- Apply equivalence mappings (`"meq/L" ↔ "mEq/L"`, `"10*3/uL" ↔ "K/uL"`) — ideally shared with the FHIR generator's UCUM map

Then in `validateLabResult`:
```ts
const normalizedInput = normalizeUnit(unit);
const normalizedAllowed = ref.allowed_units.map(normalizeUnit);
if (!normalizedAllowed.includes(normalizedInput)) { ... }
```

## Acceptance
- Test: `validateLabResult("Glucose", 95, "mg/dl")` → valid
- Test: `validateLabResult("Potassium", 4.1, "meq/L")` → valid
- Test: `validateLabResult("Glucose", 200, " mmol/L ")` → still rejected (normalization doesn't change the unit class, only the format)
- Error messages still print the caller's original string for debugging.

blamechris:
Closes #538

## Summary
`validateLabResult` compared the caller's `unit` against `allowed_units` by exact string match, so `"mg/dl"`, `"MG/DL"`, `" mg/dL"`, and `"mg / dL"` all rejected as invalid even though they are the same unit. Real-world FHIR/HL7 feeds routinely send these variants — the allow-list added in #500 was producing the inverse of the sentinel event it was designed to prevent.

Fix: a small `normalizeUnit` helper is applied to both sides of the compare.

```ts
function normalizeUnit(u: string): string {
  return u.trim().toLowerCase().replace(/\s+/g, "");
}
```

Normalisation rules:
- `trim()` + collapse internal whitespace (so `"mg / dL"` collapses to `"mg/dL"` before lowercasing).
- `toLowerCase()` for compare only — error messages still quote the caller's original string verbatim so clinicians see what they typed.
- `/` and digits preserved, so `"10*3/uL"` and `"K/uL"` remain distinguishable.

Class-mismatch rejections (e.g. `mmol/L` against Glucose) still fire as errors — normalisation changes format only, not unit class.

## New accepted-variant tests
- `validateLabResult("Glucose", 95, "MG/DL")` accepted (upper case)
- `validateLabResult("Glucose", 95, "mg/dL ")` accepted (trailing space)
- `validateLabResult("Glucose", 95, " mg/dl")` accepted (leading space + lower)
- `validateLabResult("Glucose", 95, "mg/dl")` accepted (UCUM lowercase)
- `validateLabResult("Glucose", 95, "mg / dL")` accepted (internal whitespace)
- `validateLabResult("Potassium", 4.1, "meq/L")` accepted (lowercase UCUM equivalent)
- `validateLabResult("Glucose", 95, "dL")` still rejected (genuinely wrong unit)
- Error message quotes caller's original `"mmol/L"` verbatim on class-mismatch

## Test plan
- [x] `pnpm --filter @carebridge/medical-logic test` — 163 passed (8 new cases)
- [x] `pnpm lint` — green (only unrelated pre-existing portal warnings)

### DIFF
--- a/packages/medical-logic/src/__tests__/medical-validation.test.ts
+++ b/packages/medical-logic/src/__tests__/medical-validation.test.ts
@@ -213,6 +213,60 @@ describe("validateLabResult", () => {
     expect(result.valid).toBe(true);
     expect(result.warnings.some((w) => w.includes("does not match expected"))).toBe(true);
   });
+
+  // Unit comparison tolerates case/whitespace variants. Real-world FHIR/HL7
+  // feeds frequently send "mg/dl" (UCUM canonical) or "MG/DL" (lab vendor
+  // shorthand); normalising both sides prevents false-reject errors on
+  // otherwise legitimate readings. See issue #538.
+  it("accepts glucose unit in upper case (MG/DL)", () => {
+    const result = validateLabResult("Glucose", 95, "MG/DL");
+    expect(result.valid).toBe(true);
+    expect(result.errors).toHaveLength(0);
+  });
+
+  it("accepts glucose unit with trailing whitespace (mg/dL )", () => {
+    const result = validateLabResult("Glucose", 95, "mg/dL ");
+    expect(result.valid).toBe(true);
+    expect(result.errors).toHaveLength(0);
+  });
+
+  it("accepts glucose unit with leading whitespace + lowercase ( mg/dl)", () => {
+    const result = validateLabResult("Glucose", 95, " mg/dl");
+    expect(result.valid).toBe(true);
+    expect(result.errors).toHaveLength(0);
+  });
+
+  it("accepts glucose unit in canonical UCUM lowercase (mg/dl)", () => {
+    const result = validateLabResult("Glucose", 95, "mg/dl");
+    expect(result.valid).toBe(true);
+    expect(result.errors).toHaveLength(0);
+  });
+
+  it("accepts glucose unit with internal whitespace (mg / dL)", () => {
+    const result = validateLabResult("Glucose", 95, "mg / dL");
+    expect(result.valid).toBe(true);
+    expect(result.errors).toHaveLength(0);
+  });
+
+  it("accepts potassium submitted as lowercase meq/L", () => {
+    const result = validateLabResult("Potassium", 4.1, "meq/L");
+    expect(result.valid).toBe(true);
+    expect(result.errors).toHaveLength(0);
+  });
+
+  it("still rejects a genuinely wrong unit (dL alone)", () => {
+    const result = validateLabResult("Glucose", 95, "dL");
+    expect(result.valid).toBe(false);
+    expect(result.errors.some((e) => e.includes("unit"))).toBe(true);
+  });
+
+  it("error message quotes the caller's original (non-normalized) unit", () => {
+    // Normalisation is only used for comparison; the operator's typed
+    // string should surface verbatim so they can see what they entered.
+    const result = validateLabResult("Glucose", 200, "mmol/L");
+    expect(result.valid).toBe(false);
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review — PR #562

**Summary:** Clean, well-scoped fix for a real-world FHIR/HL7 interoperability gap where case and whitespace variants of valid units were being falsely rejected. The `normalizeUnit` helper is minimal and correct for the current allow-list, and the test suite covers the important variants thoroughly.

### Strengths
- **Correct layering**: normalisation is applied to *both* sides of the comparison, so allow-list entries themselves don't need to be maintained in multiple casings.
- **Error message fidelity**: quoting the caller's original (non-normalised) unit in error messages is the right call — clinicians need to see what the system received.
- **Good test coverage**: 8 new cases covering upper, lower, leading/trailing/internal whitespace, still-rejected wrong unit, and verbatim error message. The "wrong unit still rejected" case (`"dL"`) is especially important as a regression guard.
- **Tight scope**: no unnecessary refactors, no unrelated changes.

### Issues Found

#### Critical
*None.*

#### Suggestions

| S1 | `normalizeUnit` | Unicode micro sign (`µ` U+00B5, `μ` U+03BC) is sent by some European lab/FHIR systems instead of ASCII `u`. Current normalisation won't match `µg/dL` → `mcg/dL`. Low risk today but latent. | Add `.replace(/[\u00b5\u03bc]/g, "u")` step. Filed as #574. |
| S2 | `normalizeUnit` | Function is private but potentially useful in `fhir-gateway` and `clinical-data` services that also handle unit strings. Consider exporting it from `@carebridge/medical-logic`. | Not blocking — can be done when the next consumer appears. |

#### Nitpicks

| # | Location | Note |
|---|----------|------|
| N1 | `ref.allowed_units.map(normalizeUnit)` | This allocates a new array on every validation call. Negligible for small arrays but if this ever becomes a hot path, the normalised allow-list could be memoised at module load time. Not worth changing now. |

### Architecture Notes
The normalisation approach (format-only, no semantic aliasing) is correct for this layer. Semantic unit equivalence (e.g., `µmol/L` ↔ `mg/dL` with a conversion factor) belongs in a dedicated unit-conversion module, not here. This function's job is strictly "same unit, different formatting" — and it does exactly that.

### Verdict: **Approve**

Solid fix. S1 (Unicode mu) is filed as a follow-up (#574) and is not blocking. Ship it.

PR #563, opened 2026-04-16
fix(portal): distinguish allergy error, loading, and unknown states in overview tab

### LINKED ISSUE #218: clinical-safety(p2): overview tab silently swallows allergy query errors — shows NKDA when fetch fails
## Problem
`apps/clinician-portal/app/patients/[id]/page.tsx` (lines 32-122), `OverviewTab`:

```typescript
const diagnoses = diagnosesQuery.data ?? [];   // line 39
const allergies = allergiesQuery.data ?? [];   // line 40
const careTeam  = careTeamQuery.data  ?? [];   // line 41

// Error check only looks at patientQuery:
if (diagnosesQuery.isLoading) return <LoadingState label="overview" />;
if (diagnosesQuery.isError || !patient) return <ErrorState label="overview" />;
// allergiesQuery.isError is NEVER checked
```

When the allergies API returns a 500 error:
- `allergiesQuery.isError = true`
- `allergiesQuery.data = undefined`
- `allergies` defaults to `[]`
- The component renders the allergy section as "NKDA" (no known drug allergies)

## Clinical Consequence
**This is the most dangerous silent failure in the codebase.**

- Patient has documented penicillin + cephalosporin allergy in the database
- The allergies endpoint returns a 500 error (database timeout, connection pool exhausted)
- Clinician sees "NKDA" in the patient overview
- Clinician prescribes amoxicillin
- Patient has an anaphylactic reaction

## Fix
Check `allergiesQuery.isError` explicitly:
```typescript
if (diagnosesQuery.isError || allergiesQuery.isError || careTeamQuery.isError || !patient) {
  return <ErrorState label="overview" />;
}
```

Or, display per-section error states so the allergy section shows "Unable to load allergies — do not prescribe until confirmed" rather than silently showing NKDA.

blamechris:
## Summary

Closes #218.

The patient overview tab rendered "NKDA" whenever the allergies tRPC query returned an empty list, including the `isError` path. A network error or backend outage therefore silently misrepresented documented drug allergies as "no known drug allergies" — a clinical-safety hazard that could lead to prescribing a contraindicated drug (e.g., amoxicillin to a penicillin-allergic patient).

## Change

Introduces a pure `deriveAllergyDisplayState()` helper in `apps/clinician-portal/src/lib/allergy-display.ts` that resolves the query + `patient.allergy_status` into six explicit render kinds. Order of checks is load-bearing: `isError` wins over cached data, and `isLoading` wins over the empty-list branch so "NKDA" never flashes on mount.

## Rendered states

- **loading** — spinner/placeholder ("Loading...").
- **error** — critical-colored advisory "`⚠` Allergy data unavailable — refresh before prescribing" with raw error tucked behind a "Show details" toggle. Never "NKDA".
- **populated** — existing list rendering (`allergen (reaction)`).
- **empty + `allergy_status === "nkda"`** — "NKDA (confirmed)" in success color.
- **empty + `allergy_status === "unknown"`** — "Allergy status unknown — not assessed" in warning color. Distinct from NKDA.
- **empty + `allergy_status === "has_allergies"`** — "Has allergies — none documented (data gap)" in warning color.

Absent `allergy_status` defaults to `unknown` (the safer assumption), not NKDA.

## Semantics alignment

The empty-list semantics mirror `formatAllergies()` in `packages/ai-prompts/src/clinical-review.ts` (lines 103-124), so the clinician UI and the LLM reviewer agree on what `allergies: []` means — neither should treat an empty list as NKDA unless `allergy_status === "nkda"`.

## Test plan

- [x] 10 new unit tests in `apps/clinician-portal/src/__tests__/allergy-display.test.ts` covering: error precedence over stale data, loading, populated, all three empty variants, missing/undefined `allergy_status` defaulting to "unknown", and error-message fallback.
- [x] `pnpm --filter @carebridge/clinician-portal test` — 12/12 passing.
- [x] `tsc --noEmit` clean.
- [x] `pnpm lint` — no new warnings.
- [ ] Manual: open a patient chart, kill the allergies endpoint, confirm the overview shows the red unavailable advisory (not NKDA).
- [ ] Manual: patient with `allergy_status = 'unknown'` and empty allergy list renders the amber "not assessed" state.

### DIFF
--- a/apps/clinician-portal/app/patients/[id]/page.tsx
+++ b/apps/clinician-portal/app/patients/[id]/page.tsx
@@ -11,6 +11,10 @@ import {
   type FlagActionModalFlag,
 } from "@/components/flag-action-modal";
 import { VitalsTrendChart } from "@/components/vitals-trend-chart";
+import {
+  deriveAllergyDisplayState,
+  type AllergyStatus,
+} from "@/lib/allergy-display";
 
 const tabs = [
   { key: "overview", label: "Overview" },
@@ -146,13 +150,24 @@ function OverviewTab({ patientId }: { patientId: string }) {
   const allergiesQuery = trpc.patients.allergies.getByPatient.useQuery({ patientId });
   const careTeamQuery = trpc.patients.careTeam.getByPatient.useQuery({ patientId });
 
+  const [showAllergyError, setShowAllergyError] = useState(false);
+
   const patient = patientQuery.data;
   const diagnoses = diagnosesQuery.data ?? [];
-  const allergies = allergiesQuery.data ?? [];
   const careTeam = careTeamQuery.data ?? [];
 
   if (patientQuery.isLoading) return <LoadingState label="overview" />;
   if (patientQuery.isError || !patient) return <ErrorState label="overview" />;
+
+  // Issue #218: allergies must never silently fall back to "NKDA" on fetch
+  // failure. deriveAllergyDisplayState() distinguishes loading / error /
+  // the three empty variants (nkda, unknown, has_allergies) from a
+  // populated list, using the same allergy_status semantics as
+  // formatAllergies() in @carebridge/ai-prompts.
+  const allergyState = deriveAllergyDisplayState(
+    allergiesQuery,
+    (patient.allergy_status as AllergyStatus | null | undefined) ?? null,
+  );
 
   return (
     <div className="detail-grid">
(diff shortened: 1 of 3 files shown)


blamechris:
## Code Review — PR #563

**fix(portal): distinguish allergy error, loading, and unknown states in overview tab**

This PR fixes a genuine clinical-safety bug where a failed allergies fetch silently rendered "NKDA", potentially masking real drug allergies. The new `deriveAllergyDisplayState()` helper encodes six distinct render states with a deliberate, load-bearing check order, backed by 10 unit tests.

---

### Strengths

- **Clinical-safety reasoning is excellent.** The error-before-stale-data precedence, loading-before-empty guard, and "unknown" default are all correct and well-commented.
- **Semantic alignment with `formatAllergies()`** in `@carebridge/ai-prompts` means the clinician UI and the LLM reviewer agree on what `allergies: []` means. This is exactly right.
- **Pure function + discriminated union** — easy to test, impossible to render the wrong card arm without a type error. The test file covers every branch including the stale-data edge case.
- **`role="alert"` on the error state** ensures screen readers announce the unavailable indicator immediately.

---

### Issues Found

#### Critical

*None.*

#### Suggestions

| # | File | Line(s) | Issue | Suggested fix |
|---|------|---------|-------|---------------|
| S1 | `allergy-display.ts` | 22 | `AllergyStatus` is duplicated — `ai-prompts` already defines the same union at `clinical-review.ts:60`. If a fourth status is added to the DB, only one of the two will be updated. | Extract the type to `@carebridge/types` (e.g., `packages/types/src/patient.ts`) and import in both places. |
| S2 | `page.tsx` | 169 | `patient.allergy_status as AllergyStatus` is an unsafe cast. If the DB returns an unexpected string (e.g., a new status value), it silently passes through. | Use a runtime guard: `const raw = patient.allergy_status; const status: AllergyStatus \| null = (raw === "nkda" \|\| raw === "unknown" \|\| raw === "has_allergies") ? raw : null;` — then an unrecognized value falls through to the safe "unknown" default. |
| S3 | `allergy-display.ts` | 1-20 | Doc comment says "five distinct states" but then lists six bullet points. | Fix the count to "six". |
| S4 | `page.tsx` | 208-260 | The error card lacks `aria-live="assertive"` — if the query transitions from loading to error after mount, screen readers won't announce the change despite `role="alert"`. Dynamic insertion into the DOM after initial render needs `aria-live` on a wrapper or the alert region itself. | Add `aria-live="assertive"` to the `role="alert"` div, or wrap the entire allergy card in an `aria-live="polite"` region so any state transition is announced. |

#### Nitpicks

| # | File | Line(s) | Note |
|---|------|---------|------|
| N1 | `page.tsx` | 264 | `key={i}` on allergy list items uses array index. Pre-existing, but if allergies are ever reordered this causes stale DOM. Prefer `allergy.id` or a composite key. |
| N2 | `allergy-display.test.ts` | — | Consider adding a test for an unrecognized `allergy_status` string (e.g., `"reviewed"`) to document the fallback behavior if S2 is addressed. |

---

### Architecture Notes

The discriminated union approach (`AllergyDisplayState`) is a strong pattern — it makes illegal states unrepresentable. The one architectural gap is the duplicated `AllergyStatus` type between `ai-prompts` and the new `allergy-display.ts` module. Since both must agree on the semantics, a single source of truth in `@carebridge/types` would prevent divergence. This is a low-risk follow-up, not a blocker.

---

### Verdict: **Approve**

The clinical-safety fix is correct and well-tested. S2 (unsafe cast) and S4 (aria-live) are worth addressing before merge but are not blockers given the net improvement over the previous "NKDA on error" behavior. S1 (shared type) is best handled as a follow-up.

PR #567, opened 2026-04-16
fix(notifications): whitelist notification categories for lock-screen safety

### LINKED ISSUE #551: notifications: enforce category whitelist in buildNotificationTitle
Identified during review of PR #504.

**Context**
PR #504 adds a docstring to `buildNotificationTitle` (`services/notifications/src/workers/dispatch-worker.ts:107`) asserting that titles are PHI-safe. Today the invariant holds because all known categories are generic enum values (`cross-specialty`, `medication-safety`, `critical-value`, `patient-reported`, `care-gap`, `drug-interaction`).

**Problem**
`NotificationEvent.category` is currently typed as `string` (`services/notifications/src/queue.ts:19`). Nothing prevents a future rule author (human or LLM) from passing a free-form category like `psychiatric-symptoms-self-harm` or `hiv-status-change` that, while not containing numbers or patient ID, meaningfully reveals PHI by association on a lock screen.

**Fix**
Option A (preferred): narrow the `category` field on `NotificationEvent` to the `ClinicalFlagCategory` union from `@carebridge/types` / the rules layer. This catches violations at compile time.

Option B: inside `buildNotificationTitle`, check `event.category` against a whitelist of safe-to-display categories; fall back to `"clinical alert"` for unknown values. Log a warning so we can detect and triage new categories.

**Acceptance criteria**
- [ ] Categories not in the approved set cannot surface in the lock-screen title.
- [ ] Test asserts that an unknown category (e.g., `"hiv-status"`) does not appear in the published title.

blamechris:
Closes #551

## Why
`NotificationEvent.category` is typed as free-form `string`, so a
future rule author or LLM response could set it to a PHI-adjacent
value like `psychiatric-symptoms-self-harm` or `hiv-status-change`
that would render verbatim in a push-notification title surfaced on a
device lock screen. The existing dash-to-space transform in
`buildNotificationTitle` / `buildSafeSummary` offered no guard.

## What changed
- New `CATEGORY_LABELS` whitelist in `services/notifications/src/workers/dispatch-worker.ts`
  mapping every canonical `FlagCategory` value to a human-readable
  label:
    - `cross-specialty` -> Cross-specialty concern
    - `drug-interaction` -> Drug interaction
    - `medication-safety` -> Medication safety
    - `care-gap` -> Care gap
    - `critical-value` -> Critical value
    - `trend-concern` -> Trend concern
    - `documentation-discrepancy` -> Documentation discrepancy
    - `patient-reported` -> Patient-reported concern
- `buildNotificationTitle` and `buildSafeSummary` now route the
  category through `resolveCategoryLabel` instead of the raw
  dash-to-space transform.
- Unknown categories fall back to the generic label `"Clinical alert"`
  and emit a structured `console.warn` with `{ category, fallback }`
  context so operators can triage and either admit the new taxonomy to
  the whitelist or fix the upstream producer.
- Lock-screen PHI invariant is preserved: the full `event.summary` is
  still persisted on `notifications.body` (encrypted at rest) and only
  the category-derived `summary_safe` is handed to the push layer.

## Tests
- 8 parameterised cases assert each whitelisted category renders to
  its expected label and does not trigger the warn log.
- 1 case asserts an unknown category (`hiv-status-change`) falls back
  to `"Clinical alert"`, never leaks tokens from the raw category, and
  emits a structured warn with the expected context fields.
- 1 case re-runs the same unknown-category event with a PHI-laden
  `summary` and asserts the published lock-screen payload carries no
  digits, no `psychiatric`, and no `MRN`.
- Existing PHI lock-screen-safety tests (issue #289) updated to the
  new label capitalisation; all 57 notifications tests pass.

## Test plan
- [x] `pnpm --filter @carebridge/notifications test` -> 57 passed
- [x] `pnpm typecheck`
- [x] `pnpm lint`

### DIFF
--- a/services/notifications/src/__tests__/dispatch-worker.test.ts
+++ b/services/notifications/src/__tests__/dispatch-worker.test.ts
@@ -308,10 +308,11 @@ describe("dispatch-worker", () => {
       await processorFn({ data: event, id: "job-phi-3" });
 
       // Reproduce buildSafeSummary locally (mirror of the private helper):
-      // a category-only template, PHI-free by construction.
+      // a category-only template, PHI-free by construction. The category
+      // is rendered through the whitelist label (`critical-value` →
+      // "Critical value") rather than a raw dash→space transform.
       const expectedSafe =
-        `Clinical flag — ${event.category.replace(/-/g, " ")}. ` +
-        `Open the portal to view details.`;
+        "Clinical flag — Critical value. Open the portal to view details.";
 
       const insertedRecords = mockInsertValues.mock.calls[0][0];
       expect(insertedRecords).toHaveLength(1);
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review — PR #567

**fix(notifications): whitelist notification categories for lock-screen safety**

### Summary
This PR closes a real HIPAA lock-screen exposure vector: free-form `category` strings on `NotificationEvent` could surface PHI-adjacent labels (e.g. `psychiatric-symptoms-self-harm`) verbatim in push-notification titles on locked devices. The fix introduces a `CATEGORY_LABELS` whitelist with a safe generic fallback and structured logging for unknown values. Test coverage is thorough, including parameterised whitelisted-category checks, unknown-category fallback, and PHI-leakage assertions.

### Strengths
- Whitelist exactly mirrors the `FlagCategory` union in `@carebridge/shared-types/ai-flags` (all 8 values present).
- Deliberate decoupling: local constant rather than deriving from the type, so new categories cannot bypass review.
- Fallback label (`"Clinical alert"`) is appropriately generic — no PHI leakage by construction.
- Good observability: structured `console.warn` with `{ category, fallback }` context for operator triage.
- Test matrix is comprehensive — 10 new test cases cover the full whitelist, unknown-category fallback, warn-log emission, and a dedicated PHI-laden-summary lock-screen assertion.

### Issues Found

#### Suggestions

| # | Location | Issue | Suggested Fix |
| S1 | `dispatch-worker.ts` L182+L186 | `resolveCategoryLabel()` is called **twice** per event — once in `buildNotificationTitle` and once in `buildSafeSummary`. For unknown categories this emits `console.warn` **twice** per notification, which will double-count in alerting dashboards. | Resolve the label once in `processNotificationJob` and pass it into both helpers, e.g. `const categoryLabel = resolveCategoryLabel(event.category); const title = buildNotificationTitle(event, categoryLabel); const safeSummary = buildSafeSummary(categoryLabel);` |
| S2 | `dispatch-worker.test.ts` unknown-category test | The test asserts `warnSpy.toHaveBeenCalled()` but does not assert the call count. After S1 is fixed this should assert `toHaveBeenCalledTimes(1)` to lock in single-warn semantics. | Add `expect(warnSpy).toHaveBeenCalledTimes(1);` (or `2` if the double-call is intentional). |
| S3 | `dispatch-worker.ts` L30 | `as const satisfies Record<string, string>` loses the key constraint — a typo like `"cros-specialty"` would silently compile. Consider `satisfies Record<FlagCategory, string>` to get a compile error when the upstream union gains a new member. This re-introduces the coupling but makes drift a build failure rather than a runtime warn. | `satisfies Record<FlagCategory, string>` (import `FlagCategory` from `@carebridge/shared-types/ai-flags`). Alternatively, add a static exhaustiveness test. |

#### Nitpicks

| # | Location | Note |
|---|----------|------|
| N1 | `dispatch-worker.ts` `isSafeCategory` | `Object.prototype.hasOwnProperty.call` is correct but verbose. Since `CATEGORY_LABELS` is `as const` with no prototype pollution risk, `category in CATEGORY_LABELS` suffices (guarded by the `satisfies`). |

### Architecture Notes
- The `NotificationEvent.category` field in `services/notifications/src/queue.ts` is typed as `string` (not `FlagCategory`). This is the root cause of the exposure, and the whitelist is the right mitigation layer. A longer-term follow-up could narrow the type to `FlagCategory` at the queue boundary with a Zod parse, which would shift the guard left to event emission time.
- No structured logger (pino/winston) exists in the notifications service today, so `console.warn` is consistent. When a logger is introduced, this callsite should migrate.

### Verdict: **Approve**

Clean, well-tested HIPAA safety fix. The double-warn (S1/S2) is the only functional concern and is low severity — worth fixing in this PR or a fast follow-up before merge.

PR #568, opened 2026-04-16
fix(ai): validate event.timestamp before using in context-builder filter

### LINKED ISSUE #517: ai-oversight: validate event.timestamp in rule context builder

`buildPatientContextForRules` (`services/ai-oversight/src/services/review-service.ts`) assigns `const eventAt = event.timestamp` and uses it as the right-hand side of every snapshot comparison without validating that it is a well-formed ISO-8601 string.

### Problem
- If `event.timestamp` is `undefined`, every `row.x > eventAt` becomes `row.x > undefined` = `false`; every `row.x <= eventAt` = `false`. Net effect: nothing is filtered out AND nothing is filtered in — the returned context is nonsense.
- If `event.timestamp` is a malformed string (`\"not-a-date\"`), lex comparisons still execute and produce unpredictable inclusion/exclusion.
- The `ClinicalEvent` type declares `timestamp: string` but nothing at the type boundary enforces ISO-8601 validity. Events are deserialized from the BullMQ queue and could carry whatever upstream wrote.

### Fix
At function entry:
```ts
const eventMs = Date.parse(event.timestamp);
if (!Number.isFinite(eventMs)) {
  throw new Error(`buildPatientContextForRules: invalid event.timestamp: ${event.timestamp}`);
}
```
This fails the review job (and pushes it to the BullMQ failure path where it is logged and retried) rather than silently producing a bad context.

### Acceptance
- Unit test: `buildPatientContextForRules` throws on `timestamp: \"\"`, `undefined`, `\"not-a-date\"`.
- Failure surfaces in the review-worker logs with the event id (no PHI).

blamechris:
Closes #517.

## Problem

PR #494 introduced `wasActiveAt(row, eventAt)` filtering inside both context builders, using `event.timestamp` as the right-hand side of every snapshot comparison. If the timestamp is missing or malformed, every `row.x <= eventAt` / `row.x > eventAt` silently returns the wrong value and the patient snapshot becomes nonsense — a clinical-safety hazard because rules then fire against an incorrect record.

## Fix

New helper `validateEventTimestamp(ts, { eventId, caller, now? })` at `services/ai-oversight/src/utils/validate-event-timestamp.ts`. It rejects four classes of bad input and falls back to `new Date().toISOString()` with a structured warning (event id only, no PHI):

| Case | Example | Detection |
| --- | --- | --- |
| Missing / empty / whitespace / non-string | `undefined`, `""`, `"   "`, `123` | typeof + trim check |
| Unparseable string | `"not-a-date"` | `Number.isFinite(Date.parse(ts))` |
| Implausibly old | `"1970-01-01T00:00:00Z"` | < year 2000 (epoch-leak guard) |
| Future beyond grace | `now + 5min` | > `now + 1min` clock-skew grace |

Applied at both call sites:
- `services/review-service.ts::buildPatientContextForRules` (deterministic rule path)
- `workers/context-builder.ts::buildPatientContext` (LLM path)

## Fallback strategy: fall-back-to-now vs throw

The original issue proposed throwing on invalid input and letting BullMQ retry. I went with fall-back-to-now-with-warning instead so a malformed event that would always fail (e.g. upstream publisher bug) still gets a best-effort review rather than looping in the BullMQ retry queue. The 1-minute clock-skew grace keeps `now` a reasonable proxy for the true event time in the overwhelming majority of cases (worker runs seconds after emit). If the underlying value is a genuinely-unknowable time (not just delayed), deterministic rules still run but against the current snapshot — safer than silently returning an empty or inclusive-of-everything context. Happy to flip to throw-on-invalid on request.

## Test plan

- [x] New unit tests: `services/ai-oversight/src/__tests__/validate-event-timestamp.test.ts` (10 cases)
  - Valid ISO passes through unchanged (no warning)
  - `undefined`, `""`, `"   "`, non-string → fallback + warn
  - `"not-a-date"` → fallback + warn
  - `> now + 5min` → fallback + warn
  - `now + 30s` (inside grace) → passes through
  - Year-1970 epoch → fallback + warn
  - Custom `caller` label appears in log prefix
- [x] `pnpm --filter @carebridge/ai-oversight test` — 312/312 pass
- [x] `pnpm lint` — clean (only pre-existing unrelated portal warnings)

### DIFF
--- a/services/ai-oversight/src/__tests__/validate-event-timestamp.test.ts
+++ b/services/ai-oversight/src/__tests__/validate-event-timestamp.test.ts
@@ -1,0 +1,109 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+
+import { validateEventTimestamp } from "../utils/validate-event-timestamp.js";
+
+// Fixed reference time so tests are deterministic across timezones and CI
+// clock drift. 2026-04-16T12:00:00Z aligns with the event timestamp used in
+// the context-builder tests in this service.
+const NOW_MS = Date.UTC(2026, 3, 16, 12, 0, 0);
+const now = () => NOW_MS;
+
+describe("validateEventTimestamp", () => {
+  let warnSpy: ReturnType<typeof vi.spyOn>;
+
+  beforeEach(() => {
+    warnSpy = vi.spyOn(console, "warn").mockImplementation(() => {});
+  });
+
+  afterEach(() => {
+    warnSpy.mockRestore();
+  });
+
+  it("returns a valid ISO-8601 timestamp unchanged and logs nothing", () => {
+    const ts = "2026-04-16T11:59:00.000Z";
+    const result = validateEventTimestamp(ts, { now });
+    expect(result).toBe(ts);
+    expect(warnSpy).not.toHaveBeenCalled();
+  });
+
+  it("falls back to now when timestamp is undefined, and warns", () => {
+    const result = validateEventTimestamp(undefined, { now, eventId: "evt-1" });
+    expect(result).toBe(new Date(NOW_MS).toISOString());
+    expect(warnSpy).toHaveBeenCalledTimes(1);
+    expect(warnSpy.mock.calls[0][0]).toContain("missing");
+    expect(warnSpy.mock.calls[0][0]).toContain("evt-1");
+  });
+
+  it("falls back to now when timestamp is an empty string, and warns", () => {
+    const result = validateEventTimestamp("", { now, eventId: "evt-2" });
+    expect(result).toBe(new Date(NOW_MS).toISOString());
+    expect(warnSpy).toHaveBeenCalledTimes(1);
+    expect(warnSpy.mock.calls[0][0]).toContain("empty");
+    expect(warnSpy.mock.calls[0][0]).toContain("evt-2");
+  });
+
+  it("falls back to now when timestamp is whitespace-only", () => {
+    const result = validateEventTimestamp("   ", { now });
+    expect(result).toBe(new Date(NOW_MS).toISOString());
+    expect(warnSpy).toHaveBeenCalledTimes(1);
+    expect(warnSpy.mock.calls[0][0]).toContain("empty");
+  });
+
+  it("falls back to now when timestamp is unparseable, and warns", () => {
+    const result = validateEventTimestamp("not-a-date", {
+      now,
+      eventId: "evt-3",
+    });
+    expect(result).toBe(new Date(NOW_MS).toISOString());
+    expect(warnSpy).toHaveBeenCalledTimes(1);
+    expect(warnSpy.mock.calls[0][0]).toContain("unparseable");
+    expect(warnSpy.mock.calls[0][0]).toContain("evt-3");
+  });
+
+  it("falls back to now when timestamp is more than 1 minute in the future", () => {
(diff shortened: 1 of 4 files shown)


blamechris:
## Review: fix(ai): validate event.timestamp before using in context-builder filter

### Summary
Solid defensive-programming fix for a real clinical-safety gap. The new `validateEventTimestamp` helper catches four classes of bad input (missing, unparseable, epoch-leak, future) and falls back to `now` with a structured warning. Applied symmetrically at both context-builder call sites with good test coverage.

### Strengths
- Thorough classification of failure modes (missing, unparseable, epoch-leak, clock-skew) — well thought out
- Both call sites (review-service + context-builder) updated in lockstep — dual parity maintained
- Fallback-to-now with warning is the right trade-off for a safety system: best-effort review beats silent retry loop
- Test suite covers all branches including the runtime type boundary (`number as unknown as string`)
- `now` injection via options keeps tests deterministic without real-time coupling
- PHI-safe logging (event id only) — HIPAA-conscious

### Issues Found

#### Critical
*None*

#### Suggestions

| S1 | `validate-event-timestamp.ts:85` | `now()` is called twice on the fallback path — once in the guard (`parsed > now() + ...`) and again inside `fallback()`. In production the drift is negligible, but it violates single-source-of-truth for the "current time" within one validation call. | Capture `const currentMs = now()` once at the top of the function and use it in both the guard and the fallback ISO conversion. |
| S2 | `validate-event-timestamp.test.ts` | No test for the exact boundary: `now + 60_000ms` (exactly 1 minute). The `>` vs `>=` semantics matter here — is "exactly 1 minute future" accepted or rejected? Current code uses `>` so exactly-60s passes, but this isn't documented or tested. | Add a boundary test: `new Date(NOW_MS + 60 * 1000).toISOString()` — assert it passes through (or not) and document the choice. |
| S3 | `validate-event-timestamp.ts:73` | `ts === null` is checked but the parameter type is `string \| undefined`. The null check is good defensive code (BullMQ JSON can produce null), but TypeScript will flag it if `strict` + `strictNullChecks` without widening the type signature. | Widen the parameter type to `string \| undefined \| null` to make the null branch visible to the type checker. |

#### Nitpicks

| # | Location | Note |
|---|----------|------|
| N1 | `validate-event-timestamp.ts` | The `MINUTE_MS` constant is defined but only used to derive `CLOCK_SKEW_GRACE_MS`. Since they're equal, consider just using one named constant (`CLOCK_SKEW_GRACE_MS = 60_000`) for clarity. |
| N2 | PR description | Table says "Future beyond grace" example is `now + 5min` but the clock-skew guard is 1 minute, not 5 — not wrong, but "now + 2min" would be a tighter illustration. |

### Architecture Notes
- The fallback-to-now strategy is well-reasoned for this context (safety-critical review pipeline). Worth adding a metric counter (e.g. `timestamp_fallback_total`) in a follow-up so ops can alert on a sudden spike of bad timestamps — that would indicate an upstream publisher regression.
- Long-term, the `ClinicalEvent` type marks `timestamp` as required `string`. If upstream validators (e.g. Zod schema at queue publish time) enforced ISO-8601 format, this helper would become a pure safety net rather than a primary guard. Worth tracking as a hardening item.

### Verdict: **Approve**

Clean, well-tested, safety-conscious fix. The suggestions above (S1-S3) are minor and can be addressed in this PR or as a fast follow-up — none block merge.

PR #570, opened 2026-04-16
fix(ai): harden outbox reconciler with atomic claim-then-process

### LINKED ISSUE #507: outbox reconciler: idempotency hardening against duplicate event re-emit
Identified during review of PR #491 (outbox reconciler for `failed_clinical_events`).

## Problem
The reconciler's drain loop is:

```ts
await queue.add(row.event_type, row.event_payload);     // side-effect A
await db.update(failedClinicalEvents).set({ status: "processed", ... });  // side-effect B
```

If A succeeds but B fails (DB hiccup, connection drop, process crash between the two), the row remains `status='pending'`. On the next tick, the reconciler re-emits the same `event_payload`, so the review-worker sees the same `event.id` twice.

`services/ai-oversight/src/services/review-service.ts` uses `event.id` as `trigger_event_id` on flag creation. There is no dedup on that column, so a duplicate emission will create a duplicate clinical flag row — a clinician-visible artifact.

A second, related failure mode is horizontal scaling. If ai-oversight runs on N pods, every pod's reconciler SELECTs the same `status='pending'` rows (no `FOR UPDATE SKIP LOCKED`, no claim step that writes `status='processing'` first). Every pending row gets enqueued N times per tick.

## Proposed fix
Pick one of:

1. **Job-level idempotency:** pass `jobId: row.id` on `queue.add`. BullMQ dedupes by jobId, so a re-emit after a crashed UPDATE is a no-op. Cheap, local, no DB change.
2. **Claim-then-process:** first UPDATE `status='pending' -> 'processing'` with `retry_count` guard, only enqueue rows whose UPDATE RETURNING confirms we got the claim. Survives multi-pod.
3. **Processed-at dedup on consumer:** review-service refuses to create a flag if `trigger_event_id` already has a flag row.

Option (1) + (2) together gives full protection. Option (3) is defense in depth.

## Acceptance
- Two concurrent reconciler instances draining the same outbox do not double-enqueue.
- A process crash between `queue.add` and the status UPDATE does not cause duplicate flag rows on the next tick.
- Tests cover both scenarios (can fake concurrency with overlapping SELECTs in the test harness).

blamechris:
Closes #507.

## Summary
Follow-up to PR #491. The outbox drain loop had two remaining gaps that BullMQ `jobId` dedup alone did not fully close.

### 1. Atomic claim (`pending -> processing`)
Replace the prior `SELECT FOR UPDATE SKIP LOCKED` + loose `queue.add` + `UPDATE processed` with a single statement:

```sql
UPDATE failed_clinical_events
   SET status = 'processing'
 WHERE id IN (
   SELECT id FROM failed_clinical_events
    WHERE status = 'pending' AND retry_count < 5
    ORDER BY created_at ASC
    FOR UPDATE SKIP LOCKED
    LIMIT 100
 )
 RETURNING *;
```

Why: outside an explicit transaction, `FOR UPDATE SKIP LOCKED` in the prior SELECT held the row lock only for the SELECT statement's lifetime. The lock was gone before `queue.add` ran, so a second pod could still see the same row as `status='pending'`. The atomic UPDATE auto-commits the claim, so concurrent pods observe disjoint batches and the `pending -> processing` transition is persisted before enqueue.

### 2. Stale-`processing` recovery
At the top of every tick, reset any row stuck in `status='processing'` back to `pending`. The 5-minute reconcile interval is orders of magnitude longer than the per-row work (ms), so any row still in `processing` at tick start is orphaned by a prior crashed claim. `retry_count` is intentionally not incremented — a crashed attempt shouldn't consume a retry.

This closes the "crash between `queue.add` and terminal UPDATE" window: the row stays in `processing`, gets recovered on the next tick, re-claimed, and re-emitted — with BullMQ `jobId` dedup making the re-emit a no-op if the prior `queue.add` had already succeeded.

### 3. `jobId: row.id` retained
Belt-and-braces with the recovery pass — unchanged from PR #491.

## Not in scope (from the issue body's option list)
- **Consumer-side dedup on `trigger_event_id`** (issue option 3): already implemented in `review-service.ts` for `status='completed'` review_jobs. Extending to in-flight jobs is outside the outbox reconciler's concern and would be a follow-up against `review-service`.
- **Retry cap dead-letter transition** was already in the prior code (`status='failed'` at `retry_count >= MAX_RECONCILE_RETRIES`).

## Tests
10 tests (up from 9). New/updated coverage:
- `recovers rows stuck in status='processing' at the start of each tick`
- `atomic claim is a single UPDATE ... RETURNING — the row is persisted as 'processing' before queue.add runs`
- `concurrent reconcilers claim disjoint batches`
- Updated the claim-chain mocks for all prior tests to match the new UPDATE RETURNING shape.

## Test plan
- [x] `pnpm --filter @carebridge/ai-oversight test` — 303 tests pass across 20 files
- [x] `pnpm typecheck` — 34 tasks pass

### DIFF
--- a/packages/db-schema/drizzle/0033_failed_events_updated_at.sql
+++ b/packages/db-schema/drizzle/0033_failed_events_updated_at.sql
@@ -1,0 +1,11 @@
+-- Migration: Add updated_at column to failed_clinical_events.
+--
+-- The outbox reconciler recovery pass resets all status='processing' rows
+-- back to 'pending' on each tick. Without a time guard, a concurrent pod's
+-- in-flight rows (legitimately 'processing' right now) get stolen. The
+-- updated_at column lets the recovery pass only reset rows whose last
+-- status change is older than a stale-processing threshold (5 minutes),
+-- ensuring in-flight rows from other pods are not disturbed.
+
+ALTER TABLE failed_clinical_events
+  ADD COLUMN IF NOT EXISTS updated_at text;
--- a/packages/db-schema/src/schema/clinical-data.ts
+++ b/packages/db-schema/src/schema/clinical-data.ts
@@ -117,6 +117,7 @@ export const failedClinicalEvents = pgTable("failed_clinical_events", {
   status: text("status").notNull().default("pending"), // pending, retried, failed
   retry_count: real("retry_count").notNull().default(0),
   created_at: text("created_at").notNull(),
+  updated_at: text("updated_at"),
   processed_at: text("processed_at"),
 }, (table) => [
   index("idx_failed_events_status").on(table.status, table.created_at),
--- a/services/ai-oversight/src/__tests__/outbox-reconciler.test.ts
+++ b/services/ai-oversight/src/__tests__/outbox-reconciler.test.ts
@@ -34,6 +34,7 @@ vi.mock("@carebridge/db-schema", () => ({
     status: "status",
     retry_count: "retry_count",
     created_at: "created_at",
+    updated_at: "updated_at",
     processed_at: "processed_at",
     error_message: "error_message",
   },
@@ -66,6 +67,7 @@ import {
   reconcileFailedEvents,
   MAX_RECONCILE_RETRIES,
   RECONCILE_BATCH_SIZE,
+  STALE_PROCESSING_THRESHOLD_MS,
 } from "../workers/outbox-reconciler.js";
 
 type ClaimedRow = {
@@ -76,6 +78,7 @@ type ClaimedRow = {
   status: string;
   retry_count: number;
   created_at: string;
+  updated_at: string | null;
 };
 
 function makeRow(overrides: Partial<ClaimedRow> = {}): ClaimedRow {
@@ -93,6 +96,7 @@ function makeRow(overrides: Partial<ClaimedRow> = {}): ClaimedRow {
     status: "processing",
     retry_count: 0,
     created_at: "2026-04-16T00:00:00.000Z",
+    updated_at: null,
     ...overrides,
   };
 }
@@ -185,6 +189,7 @@ describe("reconcileFailedEvents", () => {
     const terminalPayload = mockSet.mock.calls[2][0];
     expect(terminalPayload.status).toBe("processed");
     expect(terminalPayload.processed_at).toEqual(expect.any(String));
+    expect(terminalPayload.updated_at).toEqual(expect.any(String));
   });
 
   it("increments retry_count and keeps status='pending' when re-emit fails below the cap", async () => {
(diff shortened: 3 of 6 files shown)


blamechris:
## Review: fix(ai): harden outbox reconciler with atomic claim-then-process

The atomic `UPDATE...WHERE id IN (SELECT ... FOR UPDATE SKIP LOCKED) RETURNING *` is a solid correctness improvement over the prior SELECT-then-UPDATE two-step. The stale-processing recovery pass and the preserved `jobId` dedup are the right layered defense. The test suite is well-structured and the PR description is exemplary.

---

### Strengths

- The atomic claim eliminates the lock-lifetime gap from the prior `SELECT FOR UPDATE` approach — the status flip to `processing` is committed before `queue.add` runs.
- Excellent PR description: clearly explains the two gaps, the fix, and what is intentionally out of scope.
- `retry_count` not incremented on crash recovery is the correct design — a crashed attempt should not consume a retry budget.
- Test coverage for the ordering invariant (set:processing before queue.add) is a great addition.

---

### Issues Found

#### Critical

| # | Issue | Details |
|---|-------|---------|
| C1 | **Recovery pass has no time guard — can reset rows actively being processed by a concurrent pod** | The blanket \`UPDATE SET status='pending' WHERE status='processing'\` at tick start will also reset rows that a *concurrent* pod claimed seconds ago and is actively processing. The 5-minute interval argument assumes ticks never overlap, but the code is explicitly designed for horizontal scaling and the concurrent test uses \`Promise.all\`. Without a time guard (e.g., \`AND updated_at < now() - interval '5 minutes'\`), Pod B's recovery pass can steal Pod A's in-flight row, leading to double-emission that bypasses BullMQ dedup if the first job has already completed. **Fix**: add an \`updated_at\` column (or use the existing \`created_at\`) and filter the recovery query to rows older than the reconcile interval. |
| C2 | **Concurrent reconciler test does not wire recovery-pass mock responses** | The concurrent test manually sets up \`mockReturning\` for the two claim calls but does not call \`queueUpdateChain\` and does not queue \`mockUpdateWhere\` responses for the recovery passes or terminal updates. The recovery \`.where()\` calls will resolve from the default mock queue, which may be exhausted or return wrong values under microtask interleave. This test may pass by accident today but is fragile. **Fix**: use \`queueUpdateChain\` for both invocations or manually queue the full chain for each pod. |

#### Suggestions

| # | Issue | Details |
|---|-------|---------|
| S1 | **Partial-failure test removed but replacement scenario not covered** | The old "on partial failure (add resolves, UPDATE rejects) a second tick re-adds with same jobId" test is deleted. The new flow changes the failure mode (row stays \`processing\`, not \`pending\`), but there is no test covering: queue.add succeeds -> terminal UPDATE throws -> row stays processing -> next tick recovery resets to pending -> re-claim -> re-emit is BullMQ no-op. This is the exact scenario the recovery pass exists for. |
| S2 | **Batch-size test is fragile** | The new test asserts batch size by \`JSON.stringify\`-ing a Drizzle \`sql\` template object and checking it contains \`"100"\`. This couples the test to Drizzle's internal serialization format, which is not part of its public API. A Drizzle upgrade could break this silently. Consider using \`sql.toQuery()\` or a snapshot-based assertion instead. |
| S3 | **BullMQ job retention eviction can defeat dedup** | If the BullMQ completed-job retention window (\`removeOnComplete: { count: 1000 }\`) evicts a job before the next reconciler tick (5 min), a recovered row's re-emit will create a duplicate job. This is unlikely at normal throughput but becomes possible during a burst. Consider using \`removeOnComplete: { age: 600 }\` (10 min, 2x the reconcile interval) to ensure dedup coverage spans at least two ticks. |

#### Nitpicks

| # | Issue | Details |
|---|-------|---------|
| N1 | \`ClaimedRow\` type duplicates the Drizzle infer type | Consider using \`InferSelectModel<typeof failedClinicalEvents>\` from Drizzle to keep the type in sync with schema changes. |
| N2 | Unused import: \`and, lt\` removed but file header comment still references "Responsibilities" numbering that was updated in the diff — the on-disk file (line 26) still has the old import. Confirm the branch file is clean after rebase. |

---

### Architecture Notes

The atomic claim pattern is the right call for an outbox drain without an explicit transaction. The layered defense (atomic claim + recovery pass + BullMQ jobId dedup + review-service trigger_event_id dedup) is solid. The one architectural gap is the recovery pass's inability to distinguish "orphaned by crash" from "actively in-flight on another pod" — this is the C1 finding above and should be resolved before merge to avoid a subtle production race under horizontal scaling.

---

Two critical findings (C1: recovery race, C2: test wiring) should be addressed before merge. S1 (partial-failure test replacement) is strongly recommended as it covers the exact crash-recovery scenario this PR introduces.

PR #498, opened 2026-04-16
fix(portal): per-vital staleness tiers with age-at-a-glance rendering

blamechris:
## Summary
- Classify each latest vital card into \`current\` (≤4h), \`overdue\` (4h–24h), or \`stale\` (>24h).
- Overdue cards get an amber tint + \"recheck due\" note; stale cards go gray with a \"stale\" note.
- Every card now shows a relative-age line (\"3h ago\", \"2d ago\") plus the absolute timestamp.

## Why
A vital from 3 days ago rendered identically to one from 30 minutes ago — clinicians could not tell a current assessment from stale data. The 4h/24h thresholds mirror acute-care inpatient vital-sign expectations.

## Relation to #256
Composes with #256's top-of-tab banner (fires when freshest reading > 7d): banner flags the dataset as aged overall; these per-card tiers show which specific readings are still trustworthy. Small merge conflict expected if both land — straightforward to resolve.

## Test plan
- [x] \`pnpm typecheck\` / \`pnpm lint\` — clean
- [ ] Manual: load chart with 1h-old, 6h-old, and 36h-old vitals → see three distinct visual tiers
- [ ] Manual: accessibility — colors paired with text notes, not color-only signal

Closes #246

PR #571, opened 2026-04-16
fix(portal): staleness banner tokens, a11y role, and ISO timestamp compare

### LINKED ISSUE #523: clinician-portal: StaleDataBanner uses undefined --color-warning-* tokens
Identified during review of PR #496.

## Problem
`StaleDataBanner` in `apps/clinician-portal/app/patients/[id]/page.tsx` references three CSS custom properties that are **not defined anywhere in the repo**:

- `var(--color-warning-bg, #fff3cd)`
- `var(--color-warning-border, #ffc107)`
- `var(--color-warning-text, #856404)`

Because none of these tokens exist, the hardcoded hex fallbacks are always used. Those fallbacks are Bootstrap-mustard and clash with the existing Apple-dark warning palette in `apps/clinician-portal/app/globals.css` (lines 13-15):

```css
--warning: #ff9f0a;
--warning-bg: rgba(255, 159, 10, 0.12);
--warning-border: rgba(255, 159, 10, 0.25);
```

The existing tokens are used by `inbox/page.tsx`, `settings/page.tsx`, and (ironically) the vitals range chip one tab over in the **same** patient-detail page.

## Fix options
1. **Preferred:** replace the `--color-warning-*` references in `StaleDataBanner` with the already-defined `--warning-bg` / `--warning-border` / `--warning`.
2. Alternative: add a `--color-warning-*` family to `globals.css` if there is a deliberate plan to migrate token names.

## Acceptance
- StaleDataBanner visually matches other warning surfaces in the clinician portal.
- No hardcoded hex fallbacks in the banner's inline styles.

### LINKED ISSUE #525: clinician-portal: StaleDataBanner should use role="status" not role="alert"
Identified during review of PR #496.

## Problem
`StaleDataBanner` in `apps/clinician-portal/app/patients/[id]/page.tsx` uses `role="alert"`, which maps to an ARIA live region with `aria-live="assertive"`. Assertive live regions interrupt whatever the screen reader is currently announcing.

Stale-data context is informational — the clinician should notice it but does **not** need their current reading interrupted. This is the same distinction the PR already makes correctly for the empty-state containers (which use `role="status"`).

`role="alert"` should be reserved for clinical errors (see `ErrorState` on line 34 of the same file) and safety-critical warnings that demand immediate attention.

## Fix
Change `role="alert"` to `role="status"` inside `StaleDataBanner`. No visual change; screen-reader behavior becomes polite instead of assertive.

## Acceptance
- Screen reader announces the stale banner without interrupting the current reading.
- `role="alert"` in the file is limited to genuine error/critical-warning components.

### LINKED ISSUE #526: Define missing theme tokens for staleness tiers (--color-warning-*, --color-muted-*)

PR #498 references four CSS custom properties that are not defined anywhere in the repo:

- `--color-warning-bg`
- `--color-warning-border`
- `--color-muted-bg`
- `--color-muted-border`

Grep across `apps/clinician-portal` returns no definitions, so the inline hex fallbacks (`#fff8e1`, `#f0b429`, `#f4f4f5`, `#d4d4d8`) always win. This means theme overrides (e.g. dark mode) silently have no effect on overdue/stale vital cards.

### Fix
Either:
1. Define these tokens in the global theme stylesheet (likely `apps/clinician-portal/app/globals.css` or the design-system package) alongside the existing `--text-muted` / `--text-primary` tokens, or
2. Drop the `var(...)` wrapper and commit to hex-only styling until the token system is expanded.

Option 1 is preferred — it aligns with the token usage already present in the same file.

### LINKED ISSUE #529: clinician-portal: normalize ISO timestamp comparisons (freshest-reading)
Identified during review of PR #496. Same class of latent concern as PR #494.

## Problem
The freshest-timestamp computation added in PR #496 uses lexicographic string comparison on ISO timestamps:

```ts
// VitalsTab
const mostRecent = latest.reduce((a, b) =>
  a.recorded_at > b.recorded_at ? a : b,
);

// LabsTab
const mostRecentPanelAt =
  panels
    .map((p) => p.panel.collected_at ?? p.panel.created_at)
    .filter((v): v is string => Boolean(v))
    .sort()
    .slice(-1)[0] ?? null;
```

This works **only** when every timestamp is Z-suffixed UTC in an identical format. The underlying columns (`vitals.recorded_at`, `lab_panels.collected_at`) are `text` in Drizzle — nothing at the DB level prevents timezone offsets, missing milliseconds, or mixed formats.

Examples that break lexicographic ordering:
- `2026-04-16T10:00:00-05:00` vs `2026-04-16T15:00:00Z` (same instant, different sort order)
- `2026-04-16T10:00:00Z` vs `2026-04-16T10:00:00.000Z` (same instant, different strings)

Consequence: wrong "most recent" reading selected, staleness banner shown when fresh data exists, or hidden when data is actually stale.

## Fix options
1. Normalize at comparison site: `.map(x => new Date(x.recorded_at).getTime())` before reducing/sorting.
2. Introduce a shared `mostRecentByDate(items, key)` helper in `packages/portal-shared` and use it everywhere.
3. Enforce the Z-suffixed canonical form at the API boundary (Zod validator) so text comparison remains safe.

Option 2 or 3 (or both) scales better — there are likely other sites in the clinician portal doing the same thing.

## Acceptance
- Freshest-timestamp selection is correct for mixed-format ISO inputs.
- New pattern applied to both `VitalsTab` and `LabsTab` in `apps/clinician-portal/app/patients/[id]/page.tsx`.
- Ideally: sweep other call sites using `>`/`<` on timestamp strings in `apps/clinician-portal`.

### LINKED ISSUE #530: StaleDataBanner: define --color-warning-* design tokens in globals.css
Identified during review of PR #496.

`StaleDataBanner` in `apps/clinician-portal/app/patients/[id]/page.tsx` references CSS custom properties that are not defined anywhere in the portal:

- `--color-warning-bg`
- `--color-warning-border`
- `--color-warning-text`

The inline styles supply hardcoded fallbacks (`#fff3cd`, `#ffc107`, `#856404`), so the banner renders correctly today — but only the fallbacks are ever used. The clinician-portal is a dark theme (`--text-muted: #666666`, body bg `#0a0a0a`) and the light-mode yellow fallback is visually jarring against that palette.

## Tasks
- [ ] Define `--color-warning-bg`, `--color-warning-border`, `--color-warning-text` in `apps/clinician-portal/app/globals.css` `:root {}`
- [ ] Pick values that read well on the existing dark background (e.g. desaturated amber border + muted warm-tinted surface)
- [ ] Consider adding `--color-danger-*` and `--color-info-*` tokens while there for future banners (flag notifications, ErrorState, etc.)
- [ ] Audit other inline `var(--color-*, fallback)` usages to confirm no other undefined tokens are shipping

## Why deferred
PR #496 is a targeted copy/UX fix; design-token work is a separate styling concern that should be reviewed holistically with the design system rather than bolted onto a stale-data patch.

blamechris:
Bundles three small-but-load-bearing fixes to the patient chart's staleness signals in `apps/clinician-portal/app/patients/[id]/page.tsx`. All three landed in the same code path (issue-#496 StaleDataBanner + the per-vital staleness tiers from #498), so they ship together to avoid rebase conflicts and to let the tests exercise the same helpers.

## Summary

- **Design tokens** — `StaleDataBanner` and `stalenessStyles` reference five CSS custom properties (`--color-warning-bg/-border/-text`, `--color-muted-bg/-border`) that were undefined anywhere in the portal. The inline hex fallbacks (`#fff3cd` / `#ffc107` / `#856404` / `#f4f4f5` / `#d4d4d8`) therefore always won, producing a Bootstrap-mustard banner against the Apple-dark palette and silently defeating any future theme override. Define them in `globals.css` `:root` aliasing the existing `--warning-*` palette, plus a neutral translucent muted pair for the >24h stale tier.
- **A11y role** — `StaleDataBanner` now declares `aria-live="polite"` explicitly alongside `role="status"`. Stale-data context is informational; `role="alert"` would preempt the screen reader mid-sentence every time a clinician opens a patient chart with week-old data. Adds a source-level regression test so the role can't drift back to `"alert"`.
- **Timestamp compare** — `VitalsTab` and `LabsTab` both selected the freshest reading via lexicographic string compare, which is wrong whenever any input timestamps differ in format for the same instant (e.g. `Z` vs `+00:00`, `10:00:00Z` vs `10:00:00.000Z`). Extracted a reusable `pickFreshest` / `mostRecentIso` helper in `src/lib/freshest.ts` that parses via `Date.parse` to epoch-ms first, and pointed both tabs at it.

## Closes
- Closes #523 (undefined `--color-warning-*` tokens)
- Closes #525 (`role="status"` for non-urgent staleness banner)
- Closes #529 (normalize ISO timestamp comparisons)
- Closes #526 (staleness-tier tokens — addressed by same `globals.css` additions)
- Closes #530 (define `--color-warning-*` tokens in globals.css)

## Test plan
- [x] `pnpm --filter @carebridge/clinician-portal test` — 20/20 pass (10 new freshest cases covering Z vs offset vs sub-second precision vs unparseable; 5 token existence assertions; 3 banner role/aria-live assertions)
- [x] `pnpm --filter @carebridge/clinician-portal lint` — no new warnings
- [x] `tsc --noEmit` scoped to clinician-portal — removes 3 implicit-any errors that existed on `main`, introduces none
- [ ] Manual: open a patient with >7d old vitals and confirm the banner renders with the themed amber palette (not the mustard fallback)
- [ ] Manual: screen-reader smoke test that opening a chart with stale data does not interrupt an in-progress announcement

### DIFF
--- a/apps/clinician-portal/app/globals.css
+++ b/apps/clinician-portal/app/globals.css
@@ -21,6 +21,18 @@
   --success-border: rgba(48, 209, 88, 0.25);
   --accent: #0a84ff;
   --accent-hover: #409cff;
+  /* Semantic surface tokens for staleness / warning / muted banners and
+     cards. These alias the Apple-dark warning palette above and add a
+     neutral "muted" surface for the >24h "stale" tier. Defined explicitly
+     so inline styles like `var(--color-warning-bg, #fff3cd)` resolve to
+     the themed value instead of the light-mode hex fallback, and so
+     dark-mode overrides take effect. Keep in sync with --warning /
+     --warning-bg / --warning-border. (Closes #523, #526, #530.) */
+  --color-warning-bg: var(--warning-bg);
+  --color-warning-border: var(--warning);
+  --color-warning-text: var(--warning);
+  --color-muted-bg: rgba(255, 255, 255, 0.04);
+  --color-muted-border: var(--border-light);
   --sidebar-width: 240px;
   --radius-sm: 6px;
   --radius-md: 8px;
--- a/apps/clinician-portal/app/patients/[id]/page.tsx
+++ b/apps/clinician-portal/app/patients/[id]/page.tsx
@@ -5,6 +5,7 @@ import Link from "next/link";
 import { useState } from "react";
 import { trpc } from "@/lib/trpc";
 import { AuthGuard } from "@/lib/auth-guard";
+import { pickFreshest, mostRecentIso } from "@/lib/freshest";
 import {
   FlagActionModal,
   type FlagAction,
@@ -270,9 +271,15 @@ function VitalsTab({ patientId }: { patientId: string }) {
   }
 
   // Freshest recording across types — drives the staleness banner.
-  const mostRecent = latest.reduce((a, b) =>
-    a.recorded_at > b.recorded_at ? a : b,
-  );
+  // Use epoch-ms normalization (issue #529) rather than lexicographic
+  // string compare: ISO strings that represent the same instant can
+  // differ in suffix (`Z` vs `+00:00`) or sub-second precision and
+  // sort wrong as plain strings.
+  const mostRecent =
+    pickFreshest<{ recorded_at: string }>(
+      latest as Array<{ recorded_at: string }>,
+      (v) => v.recorded_at,
+    ) ?? (latest[0] as { recorded_at: string });
   const isStale =
     Date.now() - new Date(mostRecent.recorded_at).getTime() > STALE_THRESHOLD_MS;
 
@@ -363,13 +370,13 @@ function LabsTab({ patientId }: { patientId: string }) {
     );
   }
 
-  // Panels come back ordered desc by collected_at; pick the freshest.
-  const mostRecentPanelAt =
-    panels
-      .map((p) => p.panel.collected_at ?? p.panel.created_at)
-      .filter((v): v is string => Boolean(v))
-      .sort()
-      .slice(-1)[0] ?? null;
+  // Panels are returned ordered desc by collected_at, but we must still
+  // normalize via Date.parse before comparing (issue #529): lab panels
+  // from ingestion paths use mixed ISO suffixes (`Z` vs `+00:00`) and
+  // lexicographic sort can pick the wrong panel as "most recent".
+  const mostRecentPanelAt = mostRecentIso(
+    panels.map((p) => p.panel.collected_at ?? p.panel.created_at),
+  );
   const labsStale =
     mostRecentPanelAt !== null &&
     Date.now() - new Date(mostRecentPanelAt).getTime() > STALE_THRESHOLD_MS;
--- a/apps/clinician-portal/app/patients/[id]/page.tsx
+++ b/apps/clinician-portal/app/patients/[id]/page.tsx
@@ -58,10 +58,13 @@ function StaleDataBanner({
   const ageDays = Math.round(ageMs / (24 * 60 * 60 * 1000));
   // role="status" (polite live region) rather than role="alert" (assertive):
   // a chronic chart banner should not interrupt a screen-reader mid-sentence
-  // every time the clinician opens a patient with week-old data.
+  // every time the clinician opens a patient with week-old data. Explicit
+  // aria-live="polite" belt-and-suspenders in case any AT doesn't map
+  // role="status" to a polite region by default. (Issue #525.)
   return (
     <div
       role="status"
+      aria-live="polite"
       style={{
         background: "var(--color-warning-bg, #fff3cd)",
         border: "1px solid var(--color-warning-border, #ffc107)",
--- a/apps/clinician-portal/app/patients/[id]/page.tsx
+++ b/apps/clinician-portal/app/patients/[id]/page.tsx
@@ -275,17 +275,40 @@ function VitalsTab({ patientId }: { patientId: string }) {
   // string compare: ISO strings that represent the same instant can
   // differ in suffix (`Z` vs `+00:00`) or sub-second precision and
   // sort wrong as plain strings.
-  const mostRecent =
-    pickFreshest<{ recorded_at: string }>(
-      latest as Array<{ recorded_at: string }>,
-      (v) => v.recorded_at,
-    ) ?? (latest[0] as { recorded_at: string });
-  const isStale =
-    Date.now() - new Date(mostRecent.recorded_at).getTime() > STALE_THRESHOLD_MS;
+  const mostRecent = pickFreshest<{ recorded_at: string }>(
+    latest as Array<{ recorded_at: string }>,
+    (v) => v.recorded_at,
+  );
+  // Fail-safe: when all timestamps are unparseable, pickFreshest returns
+  // null and we cannot determine freshness. In that case we must still
+  // surface a staleness warning — hiding it would be clinically unsafe
+  // because the clinician would have no cue that displayed values may be
+  // outdated. (PR #571 review finding.)
+  const freshnessUnknown = mostRecent === null && latest.length > 0;
+  const isStale = freshnessUnknown
+    || (mostRecent !== null &&
+        Date.now() - new Date(mostRecent.recorded_at).getTime() > STALE_THRESHOLD_MS);
 
   return (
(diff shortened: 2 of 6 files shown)


blamechris:
## Review — PR #571: staleness banner tokens, a11y role, and ISO timestamp compare

Good, focused PR that bundles three interrelated fixes to the patient chart staleness path. The CSS token aliasing closes a real theming gap, the `role="status"` + explicit `aria-live="polite"` is the correct ARIA pattern for informational banners, and normalizing timestamps through `Date.parse` eliminates a class of silent comparison bugs.

---

### Strengths

- **Correct a11y semantics** — `role="status"` with belt-and-suspenders `aria-live="polite"` is exactly right for a non-urgent informational banner. Good judgment keeping `role="alert"` reserved for `ErrorState` and critical flags.
- **Well-designed helper API** — `pickFreshest` is generic (`<T>` + accessor), handles nullish/unparseable gracefully, and `mostRecentIso` is a clean convenience wrapper. Single-pass `O(n)` instead of sort-then-slice.
- **Thorough test coverage** — 10 freshest cases (offset, sub-second, junk), 5 token existence checks, and 3 banner a11y assertions. The tie-break stability tests (same instant, different format, input order reversed) are a nice touch.
- **CSS tokens alias existing palette** — `--color-warning-bg: var(--warning-bg)` avoids value duplication and ensures dark-mode overrides cascade correctly.

---

### Issues Found

#### Critical

| # | Location | Issue | Suggested Fix |
| 1 | `page.tsx` VitalsTab, `pickFreshest` fallback | If every timestamp is unparseable, `pickFreshest` returns `null` and the `?? latest[0]` fallback feeds a bad string into `new Date()`, producing `NaN`. `isStale` silently evaluates to `false`, **hiding the banner when data is most suspect**. Clinical fail-unsafe. | Guard the null path: if `pickFreshest` returns null and `latest` is non-empty, show the banner unconditionally (fail-safe) or render an "unknown freshness" warning. Tracked in #576. |

#### Suggestions

| # | Location | Issue | Suggested Fix |
| 1 | `page.tsx` VitalsTab L274–277 | `latest as Array<{ recorded_at: string }>` — the double type assertion discards the inferred tRPC return type. If the upstream schema ever renames `recorded_at`, this cast silently masks the error instead of surfacing a type-check failure. | Remove the `as` cast; if the tRPC type is too wide, narrow it with a type guard or satisfies clause so the compiler still checks the field name. |
| 2 | `stale-data-banner-a11y.test.ts` | Source-level regex test (reads `.tsx` as a string, strips comments, matches `role="status"`) is fragile — any innocent refactor (template literals, prop reordering, multi-line attributes) can break or false-pass it. | Extract `StaleDataBanner` to `src/components/` and use render-based assertions (`@testing-library/react`). Tracked in #577. |
| 3 | `freshest.ts` | Missing `.js` extension on the import in `page.tsx` (`from "@/lib/freshest"`). Project style mandates ESM with `.js` extensions in imports per CLAUDE.md. | Add `.js` extension: `from "@/lib/freshest.js"`. (Next.js path aliases may resolve it today, but it violates the stated convention and could break under stricter ESM loaders.) |

#### Nitpicks

| # | Location | Note |
|---|----------|------|
| 1 | `globals.css` L24–33 | The block comment is 7 lines for 5 one-liner declarations. Consider trimming to 2–3 lines — the `Closes #` references belong in the commit message, not the stylesheet. |
| 2 | `freshest.ts` L1–14 | The JSDoc duplicates the issue narrative already in the PR description and commit. A 2-line summary + `@see #529` would suffice for in-code docs. |

---

### Architecture Notes

- The `pickFreshest` / `mostRecentIso` helpers are scoped to `clinician-portal/src/lib/`. If other apps (patient-portal) or services need the same normalization, consider promoting to `packages/validators` or a shared `packages/date-utils` workspace package.
- The staleness-tier tokens (`--color-warning-*`, `--color-muted-*`) are defined only in `:root`. If a light-mode theme is added later, these will need overrides in a `@media (prefers-color-scheme: light)` block or a `.theme-light` class — the current fallback hex values (Bootstrap-light palette) would clash with a proper light theme. Not blocking, but worth a TODO comment.

---

One critical issue (fail-unsafe null fallback on `pickFreshest`) must be addressed before merge. The type assertion and ESM import extension are strong suggestions. Everything else is solid work.

PR #572, opened 2026-04-16
fix(ai): normalize ISO timestamps + exclude retracted clinical records

### LINKED ISSUE #513: ai-oversight: normalize timestamp comparison in event-time snapshot

The snapshot filter in `buildPatientContextForRules` (`services/ai-oversight/src/services/review-service.ts`) compares timestamps using raw string lexicographic comparison (`row.resolved_date <= at`, `row.ended_at <= eventAt`, `row.created_at > eventAt`).

### Problem
String comparison on ISO-8601 is only correct when **all operands are strict UTC Z-form** (`YYYY-MM-DDTHH:MM:SS.sssZ`). Current behavior:

- Internal writes via clinical-data repos always use `new Date().toISOString()` — safe.
- Seed data passes bare dates: `tooling/seed/index.ts:122` uses `onset_date: \"2025-12-01\"`. Bare-date vs Z-datetime lex-compares correctly only because `\"\"` < `\"T\"` — fragile, not asserted.
- FHIR Condition.onsetDateTime permits timezone offsets (`2026-04-16T12:00:00-05:00`). If any importer lands an offset-form string, comparison against a Z-form event.timestamp is silently wrong for the hour near date boundaries.

### Fix options (pick one)
1. **Parse at compare time** — change the helpers to `Date.parse(a) <= Date.parse(b)` (or `.getTime()`). Simple, handles all ISO-8601 variants correctly.
2. **Normalize at write time** — enforce Z-form in schema validators / Drizzle `$defaultFn` and add a runtime assert on read for legacy rows. More invasive but catches the root cause.

Recommend #1 for the rule context (fast, local) plus a linter/validator on writes to prevent new non-Z timestamps.

### Acceptance
- Regression tests for bare-date (`\"2025-12-01\"`), offset-form (`-05:00`), and Z-form timestamps against the same event.
- No behavior change for correct Z-form input.

### LINKED ISSUE #515: ai-oversight: exclude retracted clinical records from rule context

PR #494 switched the rule-context builder (`buildPatientContextForRules` in `services/ai-oversight/src/services/review-service.ts`) from `status === 'active'` filtering to timestamp-based `wasActiveAt()` filtering.

### Problem
The old `status='active'` predicate implicitly excluded records marked `entered_in_error`, `refuted`, etc. The new timestamp-only check includes **any** row with a non-null `onset_date` and null `resolved_date` regardless of status. This means:
- Diagnoses marked `entered_in_error` (clinician struck as a charting mistake) now fire rules as if real.
- Allergies with `verification_status = 'entered_in_error'` or `'refuted'` now appear in `ctx.allergies` and can drive allergy-medication cross-checks.

Including `chronic` diagnoses in the active set is a correct improvement. The `entered_in_error` / `refuted` case is a regression.

### Fix
Add an exclusion list for logical retractions in `wasActiveAt` and the allergy filter:

```ts
// diagnoses
if (row.status === 'entered_in_error') return false;
// allergies
if (a.verification_status === 'entered_in_error' || a.verification_status === 'refuted') return false;
```

### Acceptance
- New tests:
  - Diagnosis with `status: 'entered_in_error'` and null `resolved_date` is excluded.
  - Allergy with `verification_status: 'refuted'` is excluded.
  - Diagnosis with `status: 'chronic'` remains included.

blamechris:
## Summary

Two #494 follow-ups addressed together because they touch the same two
helpers (the event-time snapshot filter) and must stay in parity across
the rule-path (`buildPatientContextForRules` in review-service) and the
LLM-path (`buildPatientContext` in context-builder).

### #513 — normalize timestamp comparisons
Raw string compare (`row.resolved_date <= at`, etc.) only sorts ISO-8601
correctly when every operand is strict UTC Z-form. Offset-form
(`-05:00`) and bare-date (`2025-12-01`) values — both of which the FHIR
importer and seed data produce — silently mis-compare near day
boundaries. Replaced with `isoBefore` / `isoLTE` helpers that normalize
through `Date.parse` so every ISO-8601 variant sorts correctly.

### #515 — exclude retracted clinical records
After #494 dropped `status = 'active'` in favor of timestamp-only
filtering, diagnoses marked `entered_in_error` (charting mistakes) and
allergies with `verification_status IN ('entered_in_error', 'refuted')`
started leaking into rule context. Added `isDiagnosisRetracted` /
`isAllergyRetracted` predicates and applied them in both helpers.

### Schema note
The clinical tables do not carry generic `retracted_at` / `deleted_at`
columns — retraction is modeled via `diagnoses.status =
'entered_in_error'` and `allergies.verification_status IN
('entered_in_error', 'refuted')` per FHIR conventions, which is what
issue #515 explicitly calls for. Medications do not yet have an
equivalent retraction state.

### Two-location parity
Both `services/ai-oversight/src/services/review-service.ts` (rule path)
and `services/ai-oversight/src/workers/context-builder.ts` (LLM path)
import from the new shared module
`services/ai-oversight/src/utils/event-time-snapshot.ts` so the
deterministic rules and the LLM reason over identical row sets for a
given trigger event.

## Test plan
- [x] 10 new unit tests in `event-time-snapshot.test.ts` (Z vs offset
  suffix equality, bare dates, millisecond precision, null/garbage
  operands, retraction predicates)
- [x] +5 tests in `build-patient-context.test.ts` for the rule path
  (offset-form, bare-date, diagnosis `entered_in_error`, allergy
  retractions, null `verification_status` preservation)
- [x] +3 tests in `context-builder.test.ts` for the LLM path
- [x] `pnpm --filter @carebridge/ai-oversight test` — 320 pass
- [x] `pnpm lint` — clean (pre-existing warnings only)

Closes #513
Closes #515

### DIFF
--- a/services/ai-oversight/src/__tests__/build-patient-context.test.ts
+++ b/services/ai-oversight/src/__tests__/build-patient-context.test.ts
@@ -231,4 +231,182 @@ describe("buildPatientContextForRules — recent_labs wiring", () => {
 
     expect(ctx.recent_labs).toEqual([{ name: "WBC", value: 2.1 }]);
   });
+
+  // ─── #513 — normalize ISO timestamp comparisons ────────────────────
+  it("compares offset-form (-05:00) timestamps identically to equivalent Z-form", async () => {
+    // The event timestamp uses UTC Z-form; the diagnosis resolved_date is
+    // the identical instant expressed as -05:00 offset. Lex-compare would
+    // mis-sort these because "T07:00:00.000-05:00" sorts BEFORE
+    // "T12:00:00.000Z" as strings. After Date.parse normalization, both
+    // reduce to the same epoch and the diagnosis is correctly excluded.
+    const event: ClinicalEvent = {
+      ...stubEvent,
+      timestamp: "2026-04-16T12:00:00.000Z",
+    };
+
+    diagnosesSelect.mockResolvedValue([
+      {
+        description: "Flu (offset-form resolved_date)",
+        icd10_code: "J10.1",
+        status: "active",
+        onset_date: "2026-04-01T00:00:00.000Z",
+        // Same instant as event timestamp, different suffix — equal, so
+        // `resolved_date <= eventAt` is true and the diagnosis is excluded.
+        resolved_date: "2026-04-16T07:00:00.000-05:00",
+        created_at: "2026-04-01T00:00:00.000Z",
+      },
+    ]);
+    medicationsSelect.mockResolvedValue([]);
+    labsSelect.mockResolvedValue([]);
+
+    const ctx = await buildPatientContextForRules("p-1", event);
+
+    expect(ctx.active_diagnoses).not.toContain("Flu (offset-form resolved_date)");
+  });
+
+  it("compares bare-date (YYYY-MM-DD) timestamps correctly against Z-form event", async () => {
+    // FHIR + seed data sometimes emit bare dates. Date.parse treats
+    // "2026-04-10" as UTC midnight, so a diagnosis with
+    // resolved_date="2026-04-10" IS before event_ts=2026-04-16T12:00Z and
+    // must be excluded.
+    const event: ClinicalEvent = {
+      ...stubEvent,
+      timestamp: "2026-04-16T12:00:00.000Z",
+    };
+
+    diagnosesSelect.mockResolvedValue([
+      {
+        description: "Bronchitis (bare-date resolved_date)",
+        icd10_code: "J40",
+        status: "active",
+        onset_date: "2026-04-01",
+        resolved_date: "2026-04-10", // bare date, before event
+        created_at: "2026-04-01T00:00:00.000Z",
+      },
+      {
+        description: "Hypertension (bare-date onset, still active)",
+        icd10_code: "I10",
+        status: "active",
+        onset_date: "2025-01-15", // bare date, well before event
+        resolved_date: null,
+        created_at: "2025-01-15T00:00:00.000Z",
+      },
+    ]);
+    medicationsSelect.mockResolvedValue([]);
+    labsSelect.mockResolvedValue([]);
+
+    const ctx = await buildPatientContextForRules("p-1", event);
+
+    expect(ctx.active_diagnoses).not.toContain(
+      "Bronchitis (bare-date resolved_date)",
+    );
+    expect(ctx.active_diagnoses).toContain(
+      "Hypertension (bare-date onset, still active)",
+    );
+  });
+
+  // ─── #515 — exclude logical retractions ────────────────────────────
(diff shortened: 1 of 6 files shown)


blamechris:
## Review — PR #572: normalize ISO timestamps + exclude retracted clinical records

Two tightly coupled fixes (#513, #515) that harden the event-time snapshot filtering in both the rule path and LLM path. The shared `event-time-snapshot.ts` module is a clean extraction that eliminates prior inline duplication and guarantees both paths reason over identical row sets.

### Strengths

- **Shared module** — `isoBefore`, `isoLTE`, `isDiagnosisRetracted`, `isAllergyRetracted` in one file eliminates the silent parity drift risk between review-service and context-builder.
- **Safe NaN handling** — returning `false` on unparseable input preserves the conservative "don't filter out rows on garbage" behavior.
- **Thorough tests** — 18 new tests covering offset-form, bare-date, millisecond precision, null/garbage operands, retraction predicates, and null `verification_status` preservation. Good edge-case discipline.
- **LLM path consolidation** — inline allergy retraction logic in context-builder replaced by the shared helper, reducing future drift surface.

### Issues Found

#### Critical

None.

#### Suggestions

| S1 | `event-time-snapshot.ts` | `Date.parse` on bare dates (`"2026-04-10"`) relies on ECMAScript spec treating date-only as UTC. This is correct per spec but some older runtimes (pre-ES6) treated it as local time. | Add a brief code comment noting the ES2015+ UTC assumption, or consider an explicit `+ "T00:00:00.000Z"` suffix for bare-date inputs to be defense-in-depth. Low risk since Node 18+ is guaranteed. |
| S2 | `event-time-snapshot.ts` | No `isMedicationRetracted` — medications lack a retraction column today, but when one is added this module is the natural home. | Tracked in #581 as a follow-up. |

#### Nitpicks

| # | Location | Note |
|---|----------|------|
| N1 | `review-service.ts:533` | `wasActiveAt` type changed `status: string` → `status?: string \| null` but `status` is only consumed by `isDiagnosisRetracted`. The broader optional type is fine for forward-compat but the signature could be narrowed to `Pick<Parameters<typeof isDiagnosisRetracted>[0], 'status'>` for self-documentation. Very minor. |
| N2 | `context-builder.test.ts` | Only 3 LLM-path tests vs 5 rule-path tests — missing explicit bare-date allergy and null `verification_status` cases on the LLM side. Low risk since both share the same helpers, but full parity in tests would match the code parity goal. |

### Architecture Notes

The extraction into `event-time-snapshot.ts` is the right call. Both consumer files now import the same primitives, so any future timestamp or retraction logic change propagates automatically. The boundary semantics (`isoLTE` for resolved/ended means "resolved at event time = excluded") are consistent with prior `<=` behavior and clinically correct — a condition resolved at the exact moment of the triggering event should not drive flags.

### Verdict: **Approve**

Clean, well-tested, maintains dual-path parity. No critical issues. One follow-up issue created (#581) for medication retraction gap.

PR #573, opened 2026-04-16
fix(ai): broaden review-job idempotency + close in-flight race

### LINKED ISSUE #520: ai-oversight: broaden review-job idempotency predicate to all terminal LLM statuses
Identified during review of PR #495.

## Problem
The idempotency probe added in PR #495 (`services/ai-oversight/src/services/review-service.ts`) only short-circuits when a prior `review_jobs` row exists with `status='completed'`:

\`\`\`ts
eq(reviewJobs.status, 'completed')
\`\`\`

But the pipeline writes three terminal statuses on successful completion — `completed`, `llm_timeout`, and `llm_error` (see `llm-timeout-fallback.test.ts` and `review-service.ts:381` where `jobStatus = llmFailed ? llmFailureStatus! : 'completed'`). Only `failed` (thrown before flag writes) warrants a retry.

A redelivered event whose prior run ended at `llm_timeout`/`llm_error` re-executes the full pipeline, re-runs all deterministic rules (duplicate flag risk on rules whose dedup window doesn't cover the replay interval), and re-calls Claude — defeating much of the PR #495 value for the degraded-LLM path.

## Proposed fix
Replace the single-value `eq` with an `inArray` covering the three successful-terminal statuses:

\`\`\`ts
import { inArray } from 'drizzle-orm';
// ...
.where(
  and(
    eq(reviewJobs.trigger_event_id, event.id),
    inArray(reviewJobs.status, ['completed', 'llm_timeout', 'llm_error']),
  ),
)
\`\`\`

Intentionally exclude `failed` (thrown error before flag writes) and `processing`/`pending` (handled separately — see sibling issue on in-flight race).

## Acceptance criteria
- Probe skips pipeline for prior `completed`, `llm_timeout`, or `llm_error` rows
- Probe still proceeds when prior row is `failed`
- Test added to `review-idempotency.test.ts` parameterized across the three statuses

### LINKED ISSUE #522: ai-oversight: close in-flight race in review-job idempotency probe
Identified during review of PR #495.

## Problem
The idempotency check introduced in PR #495 only matches terminal (`completed`) rows. If two workers are handed the same `trigger_event_id` within the processing window — via BullMQ stalled-scan false positive, outbox replay, or manual requeue — both probes return empty and both execute the full pipeline concurrently. The earlier `processing` row from worker A is invisible to worker B.

Concrete scenario:
1. Worker A picks up event `evt-1` at T=0, inserts `review_jobs` row `status='processing'`
2. Worker A is mid-Claude-call at T=40s
3. BullMQ delivers the same event to worker B (any of the reasons above)
4. Worker B's probe at T=40s queries `status='completed'` → empty → proceeds
5. Both workers finish, creating double flags for any rule whose open-flag dedup can't span the race window

The partial-completion flag dedup (#261) catches most duplicates but is not a full guarantee across all rule categories.

## Proposed fix
Extend the probe to also short-circuit on an in-flight `processing` row inside a freshness window (e.g., `< now() - 150s`, comfortably longer than `lockDuration=120s`). Stale `processing` rows (older than the window) are treated as orphans and the pipeline proceeds normally, matching the current behavior:

\`\`\`ts
.where(
  and(
    eq(reviewJobs.trigger_event_id, event.id),
    or(
      inArray(reviewJobs.status, ['completed', 'llm_timeout', 'llm_error']),
      and(
        eq(reviewJobs.status, 'processing'),
        gte(reviewJobs.created_at, new Date(Date.now() - 150_000).toISOString()),
      ),
    ),
  ),
)
\`\`\`

## Acceptance criteria
- Probe skips pipeline when a recent `processing` row (<150s old) exists for the event
- Probe still proceeds when only stale `processing` rows exist (orphaned from crashed workers)
- Test added covering the in-flight race and the stale-window cases

## Dependency
Pairs with the sibling issue broadening the terminal-status predicate — ideally landed together since both are small edits to the same `where` clause.

blamechris:
## Summary

- **#520 (terminal status predicate)**: The idempotency probe in `processReviewJob` only matched `status='completed'`, missing `llm_timeout` and `llm_error`. A redelivered event whose prior run hit the degraded-LLM path re-executed the full pipeline (duplicate deterministic-rule flags, wasted Claude API calls). Now uses `inArray(status, TERMINAL_REVIEW_STATUSES)` covering all three terminal statuses. `failed` is intentionally excluded — it means the pipeline threw before flag writes, so a retry is desired.

- **#522 (in-flight race)**: Two workers handed the same `trigger_event_id` could both pass the SELECT probe (no terminal row yet) and both INSERT their own `processing` row, running the pipeline concurrently and risking duplicate flags. Extended the probe with an `or()` branch that short-circuits when a recent `processing` row exists (`created_at >= now - 150s`, comfortably beyond BullMQ `lockDuration=120s`). Stale `processing` rows (crashed-worker orphans) fall outside the freshness window and do NOT block retry, preserving crash-recovery semantics.

## Approach

Application-level query predicate changes only — no migration needed. The `TERMINAL_REVIEW_STATUSES` constant and `IN_FLIGHT_WINDOW_MS` window are defined as module-level constants for clarity.

The in-flight window narrows but does not fully close the race (two concurrent probes within the same millisecond can both see no prior row). A DB-level unique constraint would conflict with stale-orphan semantics (old `processing` rows must not block new runs) and is deferred.

## Test plan

- [x] Parameterized test across `completed`, `llm_timeout`, `llm_error` — all skip (10 idempotency tests)
- [x] `failed` status does NOT skip — pipeline proceeds to INSERT
- [x] Recent `processing` row (5s old) blocks duplicate execution
- [x] Stale `processing` row (orphan) does NOT block — pipeline proceeds
- [x] Predicate shape tests: `inArray` called with all 3 terminal statuses, `or()` used
- [x] Existing `llm-timeout-fallback` and `review-service-fallback` tests updated with `or`/`inArray` mocks
- [x] All 310 ai-oversight tests pass, typecheck clean

Closes #520
Closes #522

### DIFF
--- a/services/ai-oversight/src/__tests__/llm-timeout-fallback.test.ts
+++ b/services/ai-oversight/src/__tests__/llm-timeout-fallback.test.ts
@@ -76,6 +76,8 @@ vi.mock("drizzle-orm", () => ({
   desc: vi.fn((col: unknown) => col),
   gte: vi.fn((...args: unknown[]) => args),
   and: vi.fn((...args: unknown[]) => args),
+  or: vi.fn((...args: unknown[]) => args),
+  inArray: vi.fn((...args: unknown[]) => args),
   sql: vi.fn(),
 }));
 
--- a/services/ai-oversight/src/__tests__/review-idempotency.test.ts
+++ b/services/ai-oversight/src/__tests__/review-idempotency.test.ts
@@ -45,6 +45,8 @@ vi.mock("@carebridge/db-schema", () => ({
 vi.mock("drizzle-orm", () => ({
   eq: vi.fn(),
   and: vi.fn(),
+  or: vi.fn(),
+  inArray: vi.fn(),
   desc: vi.fn(),
   gte: vi.fn(),
   sql: vi.fn(),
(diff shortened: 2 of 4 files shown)


blamechris:
## PR #573 Review — Idempotency + In-Flight Race Fix

Solid fix for two real production risks: redelivered events re-running through terminal LLM paths (#520) and concurrent workers racing past the SELECT probe (#522). The approach is well-reasoned and the test coverage is thorough.

---

### Strengths

- Excellent documentation — both the PR description and inline comments explain the "why" for every design choice (freshness window rationale, `failed` exclusion, orphan semantics).
- Parameterized tests across all terminal statuses + the `failed`-is-not-terminal case.
- Predicate-shape tests (verifying `inArray`/`or` usage) are a nice belt-and-suspenders complement to the behavioral tests.
- `IN_FLIGHT_WINDOW_MS` > `lockDuration` alignment is correct and explicitly documented.

---

### Issues Found

#### Critical

| # | Issue | Location | Detail |
|---|-------|----------|--------|
| 1 | **Missing index on `trigger_event_id`** | `packages/db-schema/src/schema/ai-flags.ts:88` | The idempotency probe queries `WHERE trigger_event_id = ? AND (... OR ...)` on every single clinical event. The only index on `review_jobs` is `(patient_id, status)` — this probe does a sequential scan. At scale this is O(n) per event and will degrade. Add a composite index on `(trigger_event_id, status, created_at)` to cover the probe. |
| 2 | **Indentation regression** | `review-service.ts` diff, "Step 1" comment | The diff introduces an extra 2-space indent on `// Step 1: Create a review_jobs record` (4 leading spaces → 6). Minor but will fail lint if indentation rules are enforced and signals a copy-paste slip. |

#### Suggestions

| # | Issue | Location | Detail |
|---|-------|----------|--------|
| 1 | **Clock skew between workers** | `review-service.ts`, `inFlightCutoff` | `IN_FLIGHT_WINDOW_MS` compares `Date.now()` on the querying worker against `created_at` set by the inserting worker. If worker clocks differ by >30s the 150s window can effectively shrink below `lockDuration` (120s), re-opening the race. Consider using a DB-time expression (`sql\`now() - interval '150 seconds'\``) so both sides use the same clock. |
| 2 | **`TERMINAL_REVIEW_STATUSES` loses type narrowing** | `review-service.ts:75` | Typed as `readonly string[]` rather than `as const`. A `const` assertion gives a tuple literal type, so adding a new terminal status to the schema without adding it here would be catchable with an exhaustive-match helper. Also, the `as string[]` cast on line 133 strips `readonly` — use `[...TERMINAL_REVIEW_STATUSES]` spread instead to avoid the cast. |
| 3 | **Test mock schema incomplete** | `review-idempotency.test.ts:32` | The `reviewJobs` mock object doesn't include `created_at`, but the production code now references `reviewJobs.created_at` in the `gte()` call. The test passes because `gte` is a no-op mock, but if you ever add assertion-level mock verification, it will fail silently. Add `created_at: "created_at"` to the mock. |

#### Nitpicks

| # | Issue | Location | Detail |
|---|-------|----------|--------|
| 1 | **Comment in test file is stale** | `review-idempotency.test.ts:9` | Header comment still says "when a `completed` review_jobs row already exists" — should mention all terminal statuses and the in-flight case now. |

---

### Architecture Notes

The PR description correctly identifies that the freshness-window approach does not fully close the race (two probes in the same millisecond). The DB-level unique constraint deferral is reasonable given orphan-row semantics. However, consider a follow-up with a conditional `INSERT ... ON CONFLICT DO NOTHING` using a partial unique index on `(trigger_event_id) WHERE status NOT IN ('failed')` — this would close the race entirely while still allowing retries of `failed` rows (the old orphaned `processing` row would need a status-update sweep, which you already conceptually have via the freshness window expiry).

---

Two items should be addressed before merge:
1. Add a covering index on `trigger_event_id` — the probe runs on every event and a table scan is not acceptable for a hot path.
2. Fix the indentation regression on the "Step 1" comment.

The suggestions (clock skew, type narrowing, mock completeness) can be follow-ups.

blamechris:
C1 (index): Already addressed by PR #560 which adds a btree index on review_jobs(trigger_event_id, status). No duplication needed here.

PR #599, opened 2026-04-17
perf(db): add GIN index on clinical_flags.trigger_event_ids

### LINKED ISSUE #509: perf(db): add GIN index on clinical_flags.trigger_event_ids for replay-safety lookup
Identified during review of PR #492.

## Context
PR #492 added a replay-safety idempotency check at the top of `createFlag` that
filters `clinical_flags` by `patient_id` plus an `OR` of per-event-id jsonb
containment matchers:

```sql
WHERE patient_id = $1
  AND (trigger_event_ids @> '["evt-1"]'::jsonb
       OR trigger_event_ids @> '["evt-2"]'::jsonb
       ...)
```

There is no GIN index on `clinical_flags.trigger_event_ids`. The query is
currently bounded by the `idx_flags_patient (patient_id, status)` btree, so
per-patient cardinality keeps it cheap in dev and in the seeded demo.

## Why
At production scale (long-lived patients, years of flags) a seq-scan of
every flag row for a patient on each replay hop is avoidable. A GIN index
on `trigger_event_ids` would let Postgres answer the containment `OR` list
via a bitmap OR over index scans, and the `@>` operator is the canonical
GIN use case.

## Proposed change
- Add migration: `CREATE INDEX CONCURRENTLY idx_flags_trigger_event_ids ON clinical_flags USING gin (trigger_event_ids jsonb_path_ops);`
- `jsonb_path_ops` is preferred here because we only use `@>`; it is smaller
  and faster than the default `jsonb_ops`.
- Drizzle expression: `index("idx_flags_trigger_event_ids").using("gin", sql\`trigger_event_ids jsonb_path_ops\`)`

## Out of scope
Follow-up only — the PR is correct as-is; this is a scale/cost concern that
is worth landing before the DVT pilot goes live with real event volume.

blamechris:
## Summary
- Adds migration `0034_gin_trigger_event_ids.sql` with a `jsonb_path_ops` GIN index on `clinical_flags.trigger_event_ids`
- Updates Drizzle schema declaration in `ai-flags.ts` to include the new index
- Accelerates `@>` containment lookups used by the replay-safety idempotency check in `createFlag`

## Test plan
- [ ] `pnpm db:migrate` applies the new migration without errors
- [ ] Verify index exists: `\di idx_flags_trigger_event_ids` in psql
- [ ] `EXPLAIN` the replay-safety query confirms bitmap index scan on the GIN index

Closes #509

### DIFF
--- a/packages/db-schema/drizzle/0034_gin_trigger_event_ids.sql
+++ b/packages/db-schema/drizzle/0034_gin_trigger_event_ids.sql
@@ -1,0 +1,5 @@
+-- GIN index on clinical_flags.trigger_event_ids for replay-safety
+-- containment lookups (@>). jsonb_path_ops is smaller and faster than
+-- the default jsonb_ops when only @> is used.
+CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_flags_trigger_event_ids
+  ON clinical_flags USING gin (trigger_event_ids jsonb_path_ops);
--- a/packages/db-schema/drizzle/0034_gin_trigger_event_ids.sql
+++ b/packages/db-schema/drizzle/0034_gin_trigger_event_ids.sql
@@ -1,5 +1,5 @@
 -- GIN index on clinical_flags.trigger_event_ids for replay-safety
 -- containment lookups (@>). jsonb_path_ops is smaller and faster than
 -- the default jsonb_ops when only @> is used.
-CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_flags_trigger_event_ids
+CREATE INDEX IF NOT EXISTS idx_flags_trigger_event_ids
   ON clinical_flags USING gin (trigger_event_ids jsonb_path_ops);
--- a/packages/db-schema/src/schema/ai-flags.ts
+++ b/packages/db-schema/src/schema/ai-flags.ts
@@ -39,6 +39,7 @@ export const clinicalFlags = pgTable("clinical_flags", {
     table.acknowledged_at,
     table.escalation_count,
   ),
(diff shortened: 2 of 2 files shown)


blamechris:
## Review — PR #599

**Verdict: Approve with one blocking caveat**

### Findings

**1. BLOCKING — `CREATE INDEX CONCURRENTLY` cannot run inside a transaction**
The migration file uses `CREATE INDEX CONCURRENTLY`, but Drizzle's default migration runner wraps each `.sql` file in a `BEGIN`/`COMMIT` block. PostgreSQL will reject `CONCURRENTLY` inside a transaction with:

> ERROR: CREATE INDEX CONCURRENTLY cannot run inside a transaction block

Options:
- **A)** Drop the `CONCURRENTLY` keyword (fine for small tables / scheduled maintenance windows).
- **B)** Run this migration outside Drizzle's runner (e.g., `psql -f`), mark it applied manually, and add a `-- manual-migration` header comment.
- **C)** Use `drizzle-kit`'s `{ transaction: false }` option if your runner supports per-file overrides (not available in all versions).

**2. OK — Drizzle schema declaration matches SQL**
The `.using("gin", sql`trigger_event_ids jsonb_path_ops`)` in `ai-flags.ts` correctly mirrors the migration. `jsonb_path_ops` is the right operator class since only `@>` containment is used (confirmed in `flag-service.ts:53`).

**3. OK — Migration numbering**
`0034` has no collision. Three existing `0033_*` files are present but that's an existing pattern in this repo.

**4. Minor — `IF NOT EXISTS` is good defensive practice**
Appreciated for idempotency in manual-run scenarios.

PR #602, opened 2026-04-17
fix(clinical): normalize Unicode micro sign in normalizeUnit

### LINKED ISSUE #574: fix(medical-logic): normalizeUnit should handle Unicode micro sign (μ/µ) from FHIR feeds
## Context
PR #562 added `normalizeUnit()` for case/whitespace tolerance in lab unit validation. However, real-world FHIR feeds (especially from European lab systems) sometimes send the Unicode micro sign `µ` (U+00B5) or Greek lowercase mu `μ` (U+03BC) instead of ASCII `u` — e.g. `µg/dL` instead of `mcg/dL`, or `µmol/L` instead of `umol/L`.

The current `normalizeUnit` does `trim().toLowerCase().replace(/\s+/g, "")` which does **not** normalize these Unicode variants to ASCII `u`.

## Suggested fix
Add a Unicode-to-ASCII micro normalization step to `normalizeUnit`:
```ts
function normalizeUnit(u: string): string {
  return u
    .trim()
    .toLowerCase()
    .replace(/\s+/g, "")
    .replace(/[\u00b5\u03bc]/g, "u"); // µ and μ → u
}
```

## Impact
Low risk today (the `allowed_units` lists use `mcg` not `u` for micro), but this is a latent mismatch waiting to surface when international lab feeds are onboarded.

Discovered during review of #562.

blamechris:
## Summary
- `normalizeUnit` now replaces U+00B5 (µ MICRO SIGN) and U+03BC (μ GREEK SMALL MU) with ASCII `u`, so FHIR unit strings like `µg/dL`, `μg/dL`, and `ug/dL` all match equivalently.
- Added 3 tests covering micro sign, Greek mu, and ASCII equivalence for both error and warning paths.

## Test plan
- [x] All 166 medical-logic tests pass
- [x] Typecheck passes
- [x] Lint passes

Closes #574

### DIFF
--- a/packages/medical-logic/src/__tests__/medical-validation.test.ts
+++ b/packages/medical-logic/src/__tests__/medical-validation.test.ts
@@ -260,6 +260,43 @@ describe("validateLabResult", () => {
     expect(result.errors.some((e) => e.includes("unit"))).toBe(true);
   });
 
+  it("accepts glucose unit with Unicode MICRO SIGN µ (U+00B5)", () => {
+    const result = validateLabResult("Glucose", 95, "\u00b5g/dL");
+    // µg/dL should normalise to ug/dl — but glucose allowed_units is mg/dL,
+    // so this will correctly reject. Instead test with a unit whose canonical
+    // form uses 'u': we test normalisation equivalence directly.
+    // Use a mEq/L test where µ doesn't appear, to isolate: test that
+    // µg/dL, μg/dL, and ug/dL all normalise the same way via Glucose.
+    expect(result.valid).toBe(false); // µg/dL is not mg/dL
+  });
+
+  it("treats µg/dL (U+00B5), μg/dL (U+03BC), and ug/dL as identical units", () => {
+    // All three micro-sign variants should produce the same validation outcome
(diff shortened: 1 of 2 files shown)


blamechris (commented):
## Review: fix(clinical): normalize Unicode micro sign in normalizeUnit

**Verdict: Approve with follow-ups**

The targeted fix is correct for the two known micro-sign codepoints and the test assertions verify normalization equivalence. Two non-blocking gaps worth tracking:

### 1. NFKC normalization would be more robust (low priority)
`normalizeUnit` manually replaces U+00B5 and U+03BC. Using `u.normalize("NFKC")` before the chain would collapse both to ASCII-compatible forms *and* handle future lookalike codepoints (e.g., fullwidth characters, combining sequences from copy-paste). The manual `.replace()` is fine for launch but brittle long-term.

### 2. No positive-path test (medium priority)
All three new tests assert `valid: false` (unit mismatch). There is no test proving `µmol/L` is *accepted* for a lab whose `allowed_units` includes `umol/L`. A positive-path test would catch regressions where normalization silently breaks.

Both items filed as follow-up issues.

PR #605, opened 2026-04-17
fix(gateway): HIPAA minimum-necessary projection for patients.list

### LINKED ISSUE #550: gateway: apply minimum-necessary projection to patients.list procedures

## Context
PR #502 narrowed `patients.getById` to an explicit field projection that excludes `insurance_id`, `emergency_contact_name`, `emergency_contact_phone`, and `notes`. However, three `list` branches in the same router (`services/api-gateway/src/routers/patient-records.ts`) still issue `db.select().from(patients)` and therefore return those exact fields to every caller:

- L152-155 — patient self-lookup branch
- L160 — admin list
- L193 — family_caregiver branch

Only the clinician branch (L198-219) already projects. The admin list is the most impactful — a bulk response with insurance_id / emergency_contact_* / notes for every patient in the system.

## Proposal
1. Extract a shared `patientChartColumns` const matching the `getById` projection in that file.
2. Apply it to all four `list` branches.
3. If any branch legitimately needs fewer columns (e.g. patient picker tile), route it through `getSummary` or a new `listSummary` instead.

## Acceptance
- `db.select().from(patients)` with no column list no longer appears in `patient-records.ts`.
- Response typings for `patients.list` no longer include `insurance_id`, `emergency_contact_*`, `notes`.
- Regression test asserts those fields are absent from `list` output for admin and family_caregiver roles.

blamechris:
## Summary
- Extract shared `patientListColumns` const with 8 chart-relevant fields (id, name, date_of_birth, biological_sex, mrn, mrn_hmac, created_at, updated_at)
- Apply projection to all four `list` branches: patient self-lookup, admin, family_caregiver, and clinician (already had inline projection, now uses shared const)
- Eliminates `db.select().from(patients)` (no column restriction) from `patient-records.ts`
- Fields now excluded from list responses: `insurance_id`, `emergency_contact_name`, `emergency_contact_phone`, `notes`

## Test plan
- [x] Existing list RBAC tests updated to assert column projection is applied for admin and patient roles
- [x] New parameterized regression test verifies all 5 roles exclude `insurance_id`, `emergency_contact_*`, and `notes` from select args
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] All 193 api-gateway tests pass (12 in patient-records-list-rbac)

Closes #550


blamechris (commented):
## HIPAA minimum-necessary projection review

### Findings

**1. family_caregiver missing from regression test (bug)**
The parameterized regression test at the bottom of the test file iterates over `["admin", "patient", "physician", "nurse", "specialist"]` but omits `family_caregiver`. This is the one role that previously had an un-projected `select()` and is the most important to cover. The role is cast with `as string` so it is easy to miss in type-driven test generation.

**2. Silent column exclusion may regress admin/patient/family_caregiver paths**
The shared projection drops `diagnosis`, `primary_provider_id`, `allergy_status`, `weight_kg`, and `name_hmac` in addition to the four fields called out in the PR description. These were already excluded from the clinician path (pre-existing inline projection), but they were previously returned for admin, patient, and family_caregiver paths. If any downstream consumer (portal list view, search-by-name via HMAC) relies on those fields from the list endpoint, this is a breaking change. Please confirm no caller depends on them or add them to the projection.

**3. Line length nit (style)**
The family_caregiver `return db.select(patientListColumns).from(patients).where(...)` line exceeds ~100 chars — wrap for readability.

PR #608, opened 2026-04-17
fix(ai): timestamp consistency in validation and idempotency probe

### LINKED ISSUE #585: fix(ai): capture now() once in validateEventTimestamp to avoid double-call drift
From review of PR #568 (S1).

`validateEventTimestamp` calls `now()` twice on the fallback path — once in the clock-skew guard and again inside the `fallback()` closure. Capture `const currentMs = now()` once at the top and reuse it.

Low risk but improves correctness guarantees.

### LINKED ISSUE #594: fix(ai): use DB-time for in-flight freshness window to avoid clock skew
In `review-service.ts`, the in-flight freshness window computes `inFlightCutoff` using `Date.now()` on the querying worker, but compares it against `created_at` set by the inserting worker's `new Date().toISOString()`.

If worker system clocks differ by >30s, the effective 150s window can shrink below the BullMQ `lockDuration` (120s), re-opening the duplicate-execution race that #522 aimed to close.

**Suggested fix:** Replace the application-level timestamp with a DB-time expression:

```ts
const inFlightCutoff = sql`now() - interval '150 seconds'`;
```

This ensures both the insert timestamp and the comparison use the same PostgreSQL clock.

From review of #573.

blamechris:
## Summary
- **#585**: `validateEventTimestamp` called its `now()` provider twice (once for clock-skew check, once for fallback ISO). Captured the value once at function entry to eliminate second-boundary drift.
- **#594**: In-flight freshness window in `review-service.ts` used `Date.now()` for comparison against DB-set `created_at`. Replaced with Drizzle `sql\`NOW() - interval '150 seconds'\`` so both sides use the PostgreSQL clock, preventing cross-host clock skew from shrinking the window below BullMQ lock duration.

Closes #585
Closes #594

## Test plan
- [x] Existing `validate-event-timestamp` tests pass (11/11)
- [x] New test asserts `now()` provider is called exactly once on fallback path
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes


blamechris (commented):
**Verdict: Approve** — both fixes are correct and well-scoped.

### Findings

**validate-event-timestamp.ts** — Single `nowMs` capture is clean; eliminates the double-call drift. Test for call-count is precise. No issues.

**review-service.ts** — `sql\`NOW() - interval '150 seconds'\`` correctly moves both sides of the comparison onto the DB clock, eliminating cross-host skew. However:

1. **Magic literal drift risk (low):** The interval `'150 seconds'` duplicates `IN_FLIGHT_WINDOW_MS = 150_000` rather than deriving from it. If the constant changes, the SQL literal silently diverges. Filed as follow-up.

2. **No integration-level test for `sql\`NOW()\`` path:** Existing idempotency tests mock the DB layer and won't catch regressions in the raw SQL expression. Filed as follow-up.

Both are non-blocking.

PR #609, opened 2026-04-17
fix(ai): add medication retraction predicate

### LINKED ISSUE #581: fix(ai): add medication retraction predicate (isMedicationRetracted)
## Context

PR #572 added `isDiagnosisRetracted` and `isAllergyRetracted` predicates to the shared `event-time-snapshot.ts` module, but medications lack an equivalent retraction state. The PR description acknowledges this gap explicitly.

## Problem

If medications ever gain a `status = 'entered_in_error'` or similar retraction column, neither the rule path nor the LLM path will filter them out until a corresponding `isMedicationRetracted` predicate is added.

## Proposed fix

1. When the medications schema adds a retraction column, add `isMedicationRetracted` to `event-time-snapshot.ts`.
2. Apply it in both `buildPatientContextForRules` (review-service) and `buildPatientContext` (context-builder) medication filters.
3. Add unit tests mirroring the diagnosis/allergy retraction test patterns.

Surfaced during review of #572.

blamechris:
## Summary
- Add `isMedicationRetracted` predicate to `event-time-snapshot.ts` — returns `true` when `status === 'entered_in_error'`
- Wire predicate into medication filters in both `review-service.ts` (rule path) and `context-builder.ts` (LLM path)
- Add unit tests mirroring existing diagnosis/allergy retraction test patterns

## Test plan
- [x] `isMedicationRetracted` returns `true` for `entered_in_error`, `false` for `active`/`completed`/`stopped`/`null`/undefined
- [x] All 317 tests pass (`pnpm --filter @carebridge/ai-oversight test`)
- [x] Typecheck clean (`pnpm typecheck`)
- [x] Lint clean (`pnpm lint`)

Closes #581


blamechris (commented):
Predicate is correct — `status === 'entered_in_error'` matches FHIR MedicationRequest semantics. Both call-sites (review-service L651, context-builder L164) are patched identically with early-return-false before the timestamp window check, which is the right precedence. Tests cover all five realistic status values plus null/undefined. Doc comment update in event-time-snapshot.ts accurately reflects the expanded scope. No blocking issues found.

PR #611, opened 2026-04-17
fix(portal): runtime guard for allergy_status cast + aria-live

### LINKED ISSUE #583: fix(portal): add runtime guard for allergy_status cast and aria-live on allergy card
## Context

PR #563 fixes the clinical-safety bug where failed allergy fetches showed "NKDA". Two minor improvements were identified during review.

### 1. Unsafe `as AllergyStatus` cast (page.tsx:169)

`patient.allergy_status as AllergyStatus` bypasses runtime validation. If the DB returns an unexpected string, it passes through silently. Replace with a runtime guard that maps unrecognized values to `null`, letting the helper default to "unknown".

### 2. Missing `aria-live` on allergy error region (page.tsx:208-260)

The error state uses `role="alert"` but the element is inserted dynamically after mount (loading -> error transition). Adding `aria-live="assertive"` ensures screen readers announce the state change. Alternatively, wrap the entire allergy card content in an `aria-live="polite"` region.

From review of #563.

blamechris:
## Summary
- Replace unsafe `as AllergyStatus` cast in `page.tsx:173` with a new `parseAllergyStatus()` runtime guard that validates against known values (`nkda`, `unknown`, `has_allergies`) and returns `null` for unexpected strings, letting the display helper default to `"unknown"`.
- Add `aria-live="polite"` on the allergy card so screen readers announce loading/error/content transitions.
- Add unit tests for `parseAllergyStatus` (valid values, unexpected string, number, null, undefined) and the end-to-end unexpected-value path through `deriveAllergyDisplayState`.

## Test plan
- [x] `pnpm --filter @carebridge/clinician-portal test` — 39 tests pass (18 in allergy-display)

Closes #583

### DIFF
--- a/apps/clinician-portal/app/patients/[id]/page.tsx
+++ b/apps/clinician-portal/app/patients/[id]/page.tsx
@@ -14,7 +14,7 @@ import {
 import { VitalsTrendChart } from "@/components/vitals-trend-chart";
 import {
   deriveAllergyDisplayState,
-  type AllergyStatus,
+  parseAllergyStatus,
 } from "@/lib/allergy-display";
 
 const tabs = [
@@ -170,7 +170,7 @@ function OverviewTab({ patientId }: { patientId: string }) {
   // formatAllergies() in @carebridge/ai-prompts.
   const allergyState = deriveAllergyDisplayState(
     allergiesQuery,
-    (patient.allergy_status as AllergyStatus | null | undefined) ?? null,
+    parseAllergyStatus(patient.allergy_status),
   );
 
   return (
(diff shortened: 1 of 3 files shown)


blamechris (commented):
**Verdict: Approve (self-owned PR — posted as comment)**

Runtime guard is correct: `parseAllergyStatus` validates `typeof === "string"` + Set membership, returns `null` for anything unexpected, and `deriveAllergyDisplayState` defaults `null` to `"unknown"` (the clinically safer assumption). The `default` case in the switch provides a second safety net. `aria-live="polite"` is appropriate for the allergy card state transitions. Test coverage is solid across valid values, unexpected string, number, null, undefined, and the end-to-end fallback path.

Two minor observations tracked as follow-up issues:

1. `VALID_ALLERGY_STATUSES` duplicates the `AllergyStatus` type union — adding a new status to one but not the other is a silent bug. A `satisfies` check or codegen from a single source of truth would prevent drift.
2. `parseAllergyStatus` silently returns `null` for unexpected values. In a clinical system, data anomalies reaching the UI deserve at minimum a `console.warn` (or structured telemetry) so ops can detect upstream schema drift before it affects patient safety.

Neither blocks merge.

PR #612, opened 2026-04-17
test(portal): unit tests for formatAge and classifyStaleness

### LINKED ISSUE #524: Unit tests for formatAge and classifyStaleness (VitalsTab)

`formatAge` and `classifyStaleness` in `apps/clinician-portal/app/patients/[id]/page.tsx` are pure functions and currently untested. The only test under `apps/clinician-portal` is `next-config-headers.test.ts`.

### Cases to cover
- `formatAge`: `< 60s` → "just now"; `< 60m` → `Xm ago`; `< 48h` → `Xh ago`; `>= 48h` → `Xd ago`; boundary at 60s, 60m, 48h.
- `classifyStaleness`: exactly 4h, exactly 24h, just above/below each, future-dated timestamps (clock skew), invalid / NaN ISO inputs.

### Notes
- Either extract both helpers into a sibling module (`vitals-staleness.ts`) so they can be imported by a `*.test.ts` without pulling in the Next page, or colocate a `.test.tsx` next to the page.
- Prefer the extraction approach — it also unblocks reuse by the banner in #256/#496.

blamechris:
## Summary
- Extract `formatAge` and `classifyStaleness` from `app/patients/[id]/page.tsx` into `src/lib/vitals-staleness.ts` for testability and reuse
- Add 15 unit tests covering boundary values (60s, 60m, 48h for formatAge; 4h, 24h for classifyStaleness), time-unit transitions, future-dated timestamps (clock skew), and invalid ISO inputs
- Page.tsx now imports from the extracted module; no behavior change

Closes #524

## Test plan
- [x] `pnpm --filter @carebridge/clinician-portal test` — all 46 tests pass (15 new)

### DIFF
--- a/apps/clinician-portal/app/patients/[id]/page.tsx
+++ b/apps/clinician-portal/app/patients/[id]/page.tsx
@@ -17,6 +17,11 @@ import {
   parseAllergyStatus,
 } from "@/lib/allergy-display";
 import type { LabResult } from "@carebridge/shared-types";
+import {
+  formatAge,
+  classifyStaleness,
+  type StalenessTier,
+} from "@/lib/vitals-staleness";
 
 const tabs = [
   { key: "overview", label: "Overview" },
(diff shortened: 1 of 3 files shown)


blamechris (commented):
## Review — PR #612

**Extraction correctness**: `formatAge` and `classifyStaleness` are byte-identical to the originals in `page.tsx`. The `page.tsx` import path (`@/lib/vitals-staleness`) and test import (`../lib/vitals-staleness.js`) are both correct. `StalenessTier` type is re-exported. No behavior change — good.

### Issues found

**1. Test description/assertion mismatch (bug in test)**
`vitals-staleness.test.ts` line 98: the `it()` string says `"invalid ISO string results in 'stale'"` but the assertion expects `"current"`. The inline comment correctly explains NaN fall-through, but the test title is misleading and will confuse future readers. Should either fix the title or, better, guard against NaN in the implementation.

**2. `formatAge` returns `"NaNd ago"` for invalid input**
`classifyStaleness` silently falls through to `"current"` for NaN (documented in test), but `formatAge("not-a-date")` will return `"NaNd ago"` — an unguarded path with no test. In a clinical UI this would render a nonsensical string.

**3. Boundary tests are timing-sensitive (no fake timers)**
Both `formatAge` and `classifyStaleness` call `Date.now()` internally, and the `ago()` test helper also calls `Date.now()`. A few ms of wall-clock drift between the two calls can push `Math.round` across a boundary (e.g. 59.5 min rounding to 60). Tests should use `vi.useFakeTimers()` / `vi.setSystemTime()` for deterministic boundaries.

**4. Missing `formatAge` coverage for future-dated and invalid inputs**
`classifyStaleness` has tests for future timestamps (clock skew) and invalid ISO strings. `formatAge` has neither. For symmetry and safety, both should be covered.

**5. Minor: `afterEach` restores mocks but no mocks are set up**
The top-level `afterEach(() => vi.restoreAllMocks())` is harmless but currently a no-op since no `vi.spyOn` or `vi.fn` calls exist. Once fake timers are added this becomes relevant.

PR #614, opened 2026-04-17
fix(ai,notifications): JSON truncation log, single category resolve, FlagCategory constraint

### LINKED ISSUE #578: fix(ai): structured truncation log emits util.inspect not JSON
## Context
PR #564 added a structured `console.warn({ event: "llm_findings_truncated", ... })` for log aggregator consumption. However, Node's `console.warn` serializes objects via `util.inspect` (single quotes, no valid JSON), so Datadog/Loki JSON auto-parsers won't reliably parse it.

## Proposed Fix
Either:
1. Wrap in `JSON.stringify()` as a quick fix, or
2. Wait for the service-wide structured logger (pino/winston) to land and replace the call then.

Option 2 is preferable if the structured logger is imminent.

## Location
`packages/phi-sanitizer/src/llm-validator.ts`, truncation block around line 249.

From PR #564 review.

### LINKED ISSUE #591: fix(notifications): resolve category label once per event to avoid double-warn
## Context
PR #567 introduced `resolveCategoryLabel()` for HIPAA lock-screen safety. The function is called twice per notification event — once in `buildNotificationTitle` and once in `buildSafeSummary`. For unknown categories, this emits `console.warn` twice per event, which will double-count in alerting dashboards.

## Suggested fix
Resolve the label once in `processNotificationJob` and pass it into both helpers:

```ts
const categoryLabel = resolveCategoryLabel(event.category);
const title = buildNotificationTitle(event, categoryLabel);
const safeSummary = buildSafeSummary(categoryLabel);
```

Also add `expect(warnSpy).toHaveBeenCalledTimes(1)` to the unknown-category test.

## References
- PR #567 review comment
- Issue #551

### LINKED ISSUE #592: refactor(notifications): use FlagCategory type constraint on CATEGORY_LABELS
## Context
PR #567 introduced `CATEGORY_LABELS` as `Record<string, string>` with `as const satisfies`. This means a typo in a category key (e.g. `"cros-specialty"`) would silently compile and fall through to the unknown-category fallback at runtime.

## Suggested fix
Change the satisfies constraint to use the `FlagCategory` type:

```ts
import type { FlagCategory } from "@carebridge/shared-types/ai-flags.js";

const CATEGORY_LABELS = {
  ...
} as const satisfies Record<FlagCategory, string>;
```

This makes drift between the whitelist and the upstream union a compile-time error. Trade-off: re-introduces coupling, but new categories become a build failure rather than a silent runtime fallback.

## References
- PR #567 review comment
- Issue #551

blamechris:
## Summary
- **#578**: Wrap `console.warn` structured payload in `JSON.stringify()` in `llm-validator.ts` so log aggregators (Datadog, Loki) can parse truncation events as JSON instead of Node `util.inspect` format.
- **#591**: Resolve `resolveCategoryLabel()` once per event in `dispatch-worker.ts` and pass the label to both `buildNotificationTitle` and `buildSafeSummary`, eliminating duplicate `console.warn` for unknown categories.
- **#592**: Constrain `CATEGORY_LABELS` from `Record<string, string>` to `Record<FlagCategory, string>` so key typos are compile-time errors.

## Test plan
- [x] `llm-validator.test.ts` updated: truncation log assertion now parses JSON string instead of matching raw object
- [x] `dispatch-worker.test.ts` updated: unknown-category test asserts exactly 1 warn call (not 2)
- [x] `pnpm typecheck` passes (34/34)
- [x] `pnpm lint` passes
- [x] All 75 relevant tests pass

Closes #578
Closes #591
Closes #592

### DIFF
--- a/packages/phi-sanitizer/src/__tests__/llm-validator.test.ts
+++ b/packages/phi-sanitizer/src/__tests__/llm-validator.test.ts
@@ -242,15 +242,21 @@ describe("validateLLMResponse — truncation observability", () => {
     const result = validateLLMResponse(input);
 
     expect(result.ok).toBe(true);
-    // Find the structured truncation log (object arg, not a raw string).
+    // Find the structured truncation log (JSON string, not a raw object).
     const structuredCall = warnSpy.mock.calls.find(
-      (call) =>
-        typeof call[0] === "object" &&
-        call[0] !== null &&
-        (call[0] as { event?: string }).event === "llm_findings_truncated",
+      (call) => {
+        if (typeof call[0] !== "string") return false;
+        try {
+          const parsed = JSON.parse(call[0]);
+          return parsed.event === "llm_findings_truncated";
+        } catch {
+          return false;
+        }
+      },
     );
     expect(structuredCall).toBeDefined();
-    expect(structuredCall![0]).toMatchObject({
+    const parsed = JSON.parse(structuredCall![0] as string);
+    expect(parsed).toMatchObject({
       event: "llm_findings_truncated",
       received: 65,
       kept: 50,
@@ -270,10 +276,15 @@ describe("validateLLMResponse — truncation observability", () => {
 
     expect(result.ok).toBe(true);
     const structuredCall = warnSpy.mock.calls.find(
-      (call) =>
-        typeof call[0] === "object" &&
-        call[0] !== null &&
-        (call[0] as { event?: string }).event === "llm_findings_truncated",
+      (call) => {
+        if (typeof call[0] !== "string") return false;
+        try {
+          const parsed = JSON.parse(call[0]);
+          return parsed.event === "llm_findings_truncated";
+        } catch {
+          return false;
+        }
+      },
     );
     expect(structuredCall).toBeUndefined();
   });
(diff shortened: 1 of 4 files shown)


blamechris (commented):
**Verdict: Approve**

All three fixes are correct and well-tested:

1. **JSON.stringify truncation log** — Properly ensures log aggregators receive parseable JSON. Test correctly updated to parse the string.
2. **Single-resolve category label** — `resolveCategoryLabel` is now called once per entry point (`processNotificationJob` / `processDelayedNotification`), eliminating duplicate warnings. Test tightened to assert exactly 1 warn call.
3. **FlagCategory type constraint** — `Record<FlagCategory, string>` is compile-safe; the 8 union members match the 8 CATEGORY_LABELS keys exactly.

**Minor observations (non-blocking):**

- `buildSafeSummary` dropping the `event` parameter narrows future extensibility, but the current signature is cleaner. Acceptable trade-off.
- `processDelayedNotification` still calls `resolveCategoryLabel` independently, which is correct since it is a separate entry point — not a duplicate within the same flow.

No regressions detected.

PR #615, opened 2026-04-17
fix(clinical): wire validateLabResult into lab-repo and MedLens write paths

### LINKED ISSUE #539: wire validateLabResult into lab-repo write path (labs + MedLens bridge)
## Context
Identified during review of PR #500.

PR #500 added unit allow-list validation in `validateLabResult`, but the validator is not yet invoked by any write path. A FHIR-imported or MedLens-imported Glucose in `mmol/L` would persist to the database unchecked — the PR's safety logic only runs if a future caller opts in.

## Call sites needing the gate
- `services/clinical-data/src/repositories/lab-repo.ts` → `createLabPanel(input)` — primary tRPC write path used by manual entry and most inbound data.
- `services/fhir-gateway/src/medlens-bridge.ts` → `importLabs(tokenStr, labs)` — device ingest path; wraps each `MedLensLabResult` into a `CreateLabPanelInput` then calls `labRepo.createLabPanel`. Currently only confidence is checked; unit is trusted.

## Proposed behavior
Before `tx.insert(labResults)`:
1. For each result in `input.results`, call `validateLabResult(test_name, value, unit)`.
2. If `valid === false`, reject the whole panel with the joined error strings — partial-write on a lab panel is a correctness hazard.
3. Warnings (non-fatal) should attach to the emitted `clinical-events` payload so the ai-oversight worker can surface them, rather than being silently discarded.

## Acceptance
- New test: `createLabPanel` with `{ test_name: "Glucose", value: 200, unit: "mmol/L" }` throws / returns an error before insert.
- New test: MedLens `importLabs` counts the unit mismatch in `result.rejected` with the rejection reason.
- Existing tests still pass — canonical-unit fixtures should be unchanged.

## Dependency note
Best landed *after* #XXX (unit normalization) so the hard gate doesn't false-reject real FHIR lowercase-UCUM data on day one.

blamechris:
## Summary
- Import and call `validateLabResult` in `lab-repo.ts#createLabPanel` before the DB transaction — rejects the entire panel if any result has an invalid unit (e.g. Glucose in `mmol/L`)
- Import and call `validateLabResult` in `medlens-bridge.ts#importLabs` before `createLabPanel` — rejected labs appear in `ImportResult.rejected` with the reason string
- Non-fatal validation warnings are forwarded on the `clinical-events` payload (`data.validationWarnings`) so ai-oversight can surface them

## Test plan
- [x] New test: `createLabPanel` with Glucose in `mmol/L` throws before insert (lab-repo.test.ts)
- [x] New test: `createLabPanel` with high Glucose attaches warnings to emitted event (lab-repo.test.ts)
- [x] New test: MedLens `importLabs` with `mmol/L` counts rejection with reason (medlens-bridge.test.ts)
- [x] All 117 existing tests pass (46 clinical-data + 71 fhir-gateway)
- [x] `pnpm typecheck` and `pnpm lint` clean

Closes #539

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -465,6 +465,9 @@ importers:
       '@carebridge/db-schema':
         specifier: workspace:*
         version: link:../../packages/db-schema
+      '@carebridge/medical-logic':
+        specifier: workspace:*
+        version: link:../../packages/medical-logic
       '@carebridge/redis-config':
         specifier: workspace:*
         version: link:../../packages/redis-config
@@ -542,6 +545,9 @@ importers:
       '@carebridge/db-schema':
         specifier: workspace:*
         version: link:../../packages/db-schema
+      '@carebridge/medical-logic':
+        specifier: workspace:*
+        version: link:../../packages/medical-logic
       '@carebridge/phi-sanitizer':
         specifier: workspace:*
         version: link:../../packages/phi-sanitizer
--- a/services/clinical-data/package.json
+++ b/services/clinical-data/package.json
@@ -22,6 +22,7 @@
     "@carebridge/shared-types": "workspace:*",
     "@carebridge/validators": "workspace:*",
     "@carebridge/db-schema": "workspace:*",
+    "@carebridge/medical-logic": "workspace:*",
     "drizzle-orm": "^0.38.0",
     "bullmq": "^5.30.0",
     "@trpc/server": "^11.0.0",
--- a/services/clinical-data/src/__tests__/lab-repo.test.ts
+++ b/services/clinical-data/src/__tests__/lab-repo.test.ts
@@ -117,6 +117,47 @@ describe("createLabPanel", () => {
       },
     });
   });
+
+  it("rejects a panel when a result has an invalid unit", async () => {
+    await expect(
+      createLabPanel({
+        patient_id: PATIENT_ID,
+        panel_name: "BMP",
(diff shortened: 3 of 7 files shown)


blamechris (commented):
## Review — fix(clinical): wire validateLabResult into lab-repo and MedLens write paths

**Verdict: Approve with follow-ups** — Validation wiring is correct and the error/warning split is sound. Three items worth tracking:

### Findings

**1. Double validation on the MedLens path (low)**
`medlens-bridge.ts` calls `validateLabResult` then immediately calls `createLabPanel` which validates again. The bridge pre-check is correct for UX (populates `rejected`/`rejectionReasons`), but the second pass inside `createLabPanel` is redundant work. Consider adding a `skipValidation` option to `createLabPanel` for pre-validated callers.

**2. Missing test: multi-result panel with mixed validity (medium)**
`createLabPanel` rejects the entire panel if *any* result has an invalid unit (correct per the "partial writes are a correctness hazard" comment), but there is no test exercising a panel with e.g. one valid and one invalid result to prove the all-or-nothing behavior.

**3. Missing test: MedLens warnings integration (low)**
`lab-repo.test.ts` covers `validationWarnings` on the emitted event, but there is no MedLens-level test confirming that a valid-but-warned lab (e.g., Glucose 350 mg/dL) flows through `importLabs` and results in `validationWarnings` on the downstream clinical event.

Issue #627, opened 2026-04-17
test(ai): integration test for DB-clock idempotency probe


blamechris:
## Context
PR #608 moved the in-flight freshness comparison to `sql\`NOW()\`` so both sides use the PostgreSQL clock. The existing idempotency tests mock the DB layer and don't exercise the raw SQL expression.

## Proposed fix
Add an integration test (against the test DB) that inserts a `processing` review_jobs row with a known `created_at`, then calls `processReviewJob` and asserts the probe correctly matches or skips based on the DB-side `NOW()` comparison.

PR #645, opened 2026-04-17
fix(ai): derive SQL interval from IN_FLIGHT_WINDOW_MS constant

### LINKED ISSUE #622: fix(ai): derive SQL interval from IN_FLIGHT_WINDOW_MS constant
## Context
PR #608 replaced `Date.now()` with `sql\`NOW() - interval '150 seconds'\`` in the idempotency probe. The `'150 seconds'` literal duplicates `IN_FLIGHT_WINDOW_MS = 150_000` rather than deriving from it.

## Problem
If `IN_FLIGHT_WINDOW_MS` is changed (e.g. tuned for longer pipelines), the SQL interval silently diverges, reintroducing the timing bug #608 set out to fix.

## Proposed fix
Interpolate the constant into the SQL template:
```ts
const windowSec = IN_FLIGHT_WINDOW_MS / 1000;
const inFlightCutoff = sql\`NOW() - interval '${sql.raw(String(windowSec))} seconds'\`;
```

blamechris:
## Summary
- The idempotency probe in `review-service.ts` used a hardcoded SQL interval `'150 seconds'` instead of deriving from the `IN_FLIGHT_WINDOW_MS` constant. If the constant changed, the SQL interval would silently diverge.
- Now computes `windowSec = IN_FLIGHT_WINDOW_MS / 1000` and interpolates it into the SQL expression via `sql.raw()`.

Closes #622

## Test plan
- [ ] Verify `pnpm typecheck` passes
- [ ] Verify `pnpm lint` passes
- [ ] Confirm the generated SQL interval matches `IN_FLIGHT_WINDOW_MS / 1000` seconds

### DIFF
--- a/services/ai-oversight/src/__tests__/llm-timeout-fallback.test.ts
+++ b/services/ai-oversight/src/__tests__/llm-timeout-fallback.test.ts
@@ -71,15 +71,21 @@ vi.mock("@carebridge/db-schema", () => ({
   careTeamAssignments: { patient_id: "patient_id", user_id: "user_id" },
 }));
 
-vi.mock("drizzle-orm", () => ({
-  eq: vi.fn((...args: unknown[]) => args),
-  desc: vi.fn((col: unknown) => col),
-  gte: vi.fn((...args: unknown[]) => args),
-  and: vi.fn((...args: unknown[]) => args),
-  or: vi.fn((...args: unknown[]) => args),
-  inArray: vi.fn((...args: unknown[]) => args),
-  sql: vi.fn(),
-}));
+vi.mock("drizzle-orm", () => {
+  const sqlTag = vi.fn((_strings: TemplateStringsArray, ..._values: unknown[]) => ({
+    __sql: true,
+  }));
+  (sqlTag as unknown as Record<string, unknown>).raw = vi.fn((v: string) => ({ __raw: v }));
+  return {
+    eq: vi.fn((...args: unknown[]) => args),
+    desc: vi.fn((col: unknown) => col),
+    gte: vi.fn((...args: unknown[]) => args),
+    and: vi.fn((...args: unknown[]) => args),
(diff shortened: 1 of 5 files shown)


blamechris:
## Code Review — PR #645

### Verdict: Approve with suggestion

### Summary
Two-line fix that derives the SQL interval from the `IN_FLIGHT_WINDOW_MS` constant instead of hardcoding `'150 seconds'`. The intent is correct and the fix closes #622.

### Findings

**1. `sql.raw()` introduces a new pattern to the codebase (low risk, but avoidable)**

`sql.raw(String(windowSec))` bypasses Drizzle's parameterized query binding. Since the value is derived from a compile-time constant (`150_000`), there is zero injection risk today. However, this would be the first `sql.raw()` call in the entire codebase, and the idiom could be cargo-culted into places where the input *is* dynamic.

A parameterized alternative that avoids `sql.raw()` entirely:

```ts
const windowSec = IN_FLIGHT_WINDOW_MS / 1000;
const inFlightCutoff = sql`NOW() - ${windowSec} * interval '1 second'`;
```

Postgres supports `integer * interval` natively, so `150 * interval '1 second'` yields `interval '150 seconds'`. This keeps the value parameterized (`$1 * interval '1 second'`) and avoids establishing a `sql.raw()` precedent.

**2. No test coverage for the change** — The PR test plan mentions typecheck and lint but no runtime assertion. Issue #627 already tracks an integration test for this idempotency probe, so no new issue needed.

**3. Correctness** — The arithmetic `IN_FLIGHT_WINDOW_MS / 1000` yields `150` (integer division of `150_000 / 1000`), which matches the original `'150 seconds'`. No floating-point edge case here.

### Recommendation
Consider the `${windowSec} * interval '1 second'` alternative to stay within parameterized-query conventions. Either way, the fix is correct and safe to ship — this is a non-blocking suggestion.

Issue #631, opened 2026-04-17
test(portal): use fake timers in vitals-staleness boundary tests


blamechris:
**Source**: PR #612 review

The boundary tests in `vitals-staleness.test.ts` are timing-sensitive — both the `ago()` helper and the functions under test call `Date.now()` independently, so a few ms of wall-clock drift can push `Math.round` across a threshold boundary.

**Suggested fix**: Use `vi.useFakeTimers()` + `vi.setSystemTime()` to pin the clock, making all boundary assertions deterministic. Also fix the misleading test description on the invalid-ISO test (says "stale" but asserts "current").

PR #647, opened 2026-04-17
fix(portal): guard formatAge against NaN / invalid ISO input

### LINKED ISSUE #628: fix(portal): guard formatAge against NaN / invalid ISO input
**Source**: PR #612 review

`formatAge` in `src/lib/vitals-staleness.ts` returns `"NaNd ago"` when given an invalid ISO string (e.g. `"not-a-date"`). Unlike `classifyStaleness` which silently falls through to `"current"`, `formatAge` has no NaN guard and will render a nonsensical string in the clinical UI.

**Suggested fix**: Add an early `Number.isNaN(ageMs)` check that returns a safe fallback like `"unknown"` or throws. Add corresponding test case.

blamechris:
## Summary
- Adds a `Number.isNaN(ageMs)` early-return guard in `formatAge` that returns `"unknown"` instead of `"NaNd ago"` when given an invalid ISO string
- Adds test cases covering invalid ISO inputs (`"not-a-date"`, `""`, `"abc123"`)

Closes #628

## Test plan
- [x] `pnpm --filter clinician-portal test` — all 16 formatAge tests pass (3 new)
- [x] `pnpm lint` — clean (no new warnings)

### DIFF
--- a/apps/clinician-portal/src/__tests__/vitals-staleness.test.ts
+++ b/apps/clinician-portal/src/__tests__/vitals-staleness.test.ts
@@ -56,6 +56,12 @@ describe("formatAge", () => {
   it("boundary: exactly 48 hours transitions to days", () => {
     expect(formatAge(ago(48 * HOUR))).toBe("2d ago");
   });
+
+  it('returns "unknown" for invalid ISO strings', () => {
+    expect(formatAge("not-a-date")).toBe("unknown");
+    expect(formatAge("")).toBe("unknown");
+    expect(formatAge("abc123")).toBe("unknown");
+  });
 });
 
 // ---------------------------------------------------------------------------
--- a/apps/clinician-portal/src/lib/vitals-staleness.ts
+++ b/apps/clinician-portal/src/lib/vitals-staleness.ts
@@ -12,6 +12,7 @@
  */
 export function formatAge(recordedAtIso: string): string {
   const ageMs = Date.now() - new Date(recordedAtIso).getTime();
+  if (Number.isNaN(ageMs)) return "unknown";
   if (ageMs < 60_000) return "just now";
   const mins = Math.round(ageMs / 60_000);
   if (mins < 60) return `${mins}m ago`;


blamechris:
## Code Review — PR #647

### Verdict: **Approve** (no blockers)

| Area | Status | Notes |
|------|--------|-------|
| Correctness | Pass | `Number.isNaN(ageMs)` is the right guard — `new Date("garbage").getTime()` returns `NaN`, and `Date.now() - NaN` is `NaN` |
| TypeScript / ESM | Pass | No new imports; `.js` extension convention not applicable (no new module refs) |
| Tests | Pass | 3 cases cover representative invalid inputs; assertions are deterministic |
| Clinical safety | Pass | `"unknown"` is a safe fallback — never renders misleading time-ago text in the clinical UI |
| Architecture | Pass | Change is scoped to the pure utility; no side effects |

### Observations

**1. `classifyStaleness` has the same NaN-input gap (non-blocking)**
`classifyStaleness("not-a-date")` silently returns `"current"` because `NaN > threshold` is `false` for every comparison, falling through to the default. The existing test at line 99 documents this. A parallel `Number.isNaN` guard returning `"stale"` (safest clinical default) would be consistent with this PR's approach, but it is a separate concern and already observable to anyone reading the test suite. Not a blocker.

**2. Boundary tests still use wall-clock time (non-blocking)**
The new test and the existing boundary tests both call `Date.now()` via the `ago()` helper independently from the production code. A few ms of drift can theoretically flip `Math.round` at thresholds. Issue #631 tracks migrating to `vi.useFakeTimers()` — this PR correctly stays scoped and does not need to absorb that refactor.

**3. Closes #628**
This PR directly implements the fix described in issue #628. That issue can be closed.

### Summary
Clean, minimal, correct fix for a clinical-UI safety gap. The guard placement is right (earliest possible point after computing `ageMs`), the fallback string is appropriate, and the tests cover the obvious invalid-input shapes. Ship it.

PR #648, opened 2026-04-17
fix(portal): log warning when parseAllergyStatus receives unexpected value

### LINKED ISSUE #629: fix(portal): log warning when parseAllergyStatus receives unexpected value
**Context**: PR #611 added `parseAllergyStatus()` which silently returns `null` for any value outside the known enum.

**Problem**: In a clinical system, unexpected data reaching the UI indicates upstream schema drift or data corruption. Silently defaulting to "unknown" is the correct safety behavior, but the anomaly should be observable — otherwise ops has no signal that something is wrong until a clinician reports a confusing UI state.

**Suggested fix**: Add a `console.warn` (or structured telemetry call) when `parseAllergyStatus` encounters an unexpected value:
```ts
if (typeof value === "string" && statusSet.has(value)) {
  return value as AllergyStatus;
}
console.warn(`[allergy-display] unexpected allergy_status: ${JSON.stringify(value)}`);
return null;
```

`apps/clinician-portal/src/lib/allergy-display.ts`

blamechris:
## Summary
- Adds `console.warn` in `parseAllergyStatus()` when an unexpected non-null/non-undefined value is encountered, making unexpected clinical data observable in logs
- Null and undefined inputs still return `null` silently (these are expected absent-value cases)
- Adds three new test cases verifying warning emission for unexpected strings, non-string values, and no-warn for null/undefined

Closes #629

## Test plan
- [x] Existing allergy-display tests still pass (57/57)
- [x] New tests verify `console.warn` is called with the expected message format for unexpected values
- [x] New test verifies no warning for null/undefined inputs
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

### DIFF
--- a/apps/clinician-portal/src/__tests__/allergy-display.test.ts
+++ b/apps/clinician-portal/src/__tests__/allergy-display.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect } from "vitest";
+import { describe, it, expect, vi } from "vitest";
 import {
   deriveAllergyDisplayState,
   parseAllergyStatus,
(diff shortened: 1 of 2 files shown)


blamechris:
## PR #648 Review — `fix(portal): log warning when parseAllergyStatus receives unexpected value`

### Summary Table

| Area | Verdict | Notes |
|---|---|---|
| Correctness | Pass | Warning fires only for unexpected non-null/non-undefined values; null return preserved |
| TypeScript / ESM | Pass | No new imports needed; `vi` added to vitest import |
| Clinical Safety | Pass | Safety behavior unchanged — still returns `null` and defaults to "unknown" |
| Test Coverage | Pass (minor note) | 3 new test cases cover warn/no-warn paths; see note below |
| Performance | Pass | `JSON.stringify` on a single scalar is negligible |
| Architecture | Pass | Aligns with portal observability goals |

### Detailed Notes

**1. Existing tests emit unsilenced warnings**
The two pre-existing tests at lines 139-145 (`"returns null for an unexpected string"` and `"returns null for a number"`) now trigger `console.warn` as a side effect of this change. They still pass, but they produce warning output in the test runner without suppressing it. Consider either:
- Adding `vi.spyOn(console, "warn").mockImplementation(() => {})` + `mockRestore()` in those tests too, or
- Wrapping the `parseAllergyStatus` tests in a `beforeEach`/`afterEach` that globally suppresses and restores `console.warn`.

This is cosmetic — not blocking.

**2. Correct null/undefined guard placement**
The `if (value !== null && value !== undefined)` guard is correctly placed *after* the valid-status check and *before* the `return null`. This means:
- Valid values: returned directly (no warn)
- null/undefined: returned as null silently (expected absent-value)
- Everything else: warned then returned as null

This matches the issue #629 spec exactly.

**3. Future consideration: structured telemetry**
Issue #629's description mentioned `console.warn` *or structured telemetry* as options. `console.warn` is the right choice for now (portal app, client-side). If/when the portal adopts a structured logger, this should migrate — but that's out of scope for this PR.

### Verdict: **Approve**

Clean, focused fix that makes unexpected clinical data observable without changing safety behavior. The only suggestion (silencing warnings in the two pre-existing tests) is non-blocking.

PR #650, opened 2026-04-17
fix(ai): normalize brand names in QTC dedup to prevent false positives (#635)

### LINKED ISSUE #635: fix(ai): brand+generic same-drug false positive in QTC dedup
Found in PR #613 review.

The new dedup logic compares matched regex substrings (`matchA[0] !== matchB[0]`), but brand+generic pairs of the same drug (e.g. "pacerone" vs "amiodarone") produce different match strings and incorrectly fire DI-QTC-COMBO.

Need a synonym/normalization map so brand names resolve to their generic before the dedup comparison.

Affected pairs include: pacerone/cordarone/amiodarone, betapace/sotalol, tikosyn/dofetilide, haldol/haloperidol, seroquel/quetiapine, risperdal/risperidone, zyprexa/olanzapine, geodon/ziprasidone, zithromax/azithromycin, etc.

blamechris:
## Summary
- Added a `BRAND_TO_GENERIC` normalization map for QTc-prolonging drug synonyms (46 brand-to-generic mappings covering all drugs in `QTC_PATTERN`)
- Updated the DI-QTC-COMBO same-list dedup comparison to normalize matched substrings through the map before checking distinctness, so brand+generic pairs of the same drug (e.g. "pacerone" vs "amiodarone") no longer fire false positives
- Added 14 tests covering: brand+generic same-drug pairs (should NOT fire), brand+generic different-drug pairs (SHOULD fire), two-brand same-generic pairs, and generic+generic dedup

## Test plan
- [x] All 14 new tests in `drug-interaction-rules.test.ts` pass
- [x] `pnpm typecheck` passes across the monorepo
- [ ] Verify existing `rules.test.ts` suite still passes in CI

Closes #635

### DIFF
--- a/services/ai-oversight/src/__tests__/drug-interaction-rules.test.ts
+++ b/services/ai-oversight/src/__tests__/drug-interaction-rules.test.ts
@@ -1,0 +1,95 @@
+import { describe, it, expect } from "vitest";
+
+import { checkDrugInteractions } from "../rules/drug-interactions.js";
+
+describe("DI-QTC-COMBO brand/generic dedup", () => {
+  describe("brand + generic of the SAME drug should NOT fire", () => {
+    it("pacerone (brand) + amiodarone (generic)", () => {
+      const flags = checkDrugInteractions(["pacerone 200mg", "amiodarone 400mg"]);
+      const qtc = flags.filter((f) => f.rule_id === "DI-QTC-COMBO");
+      expect(qtc).toHaveLength(0);
+    });
+
+    it("cordarone (brand) + amiodarone (generic)", () => {
+      const flags = checkDrugInteractions(["cordarone 200mg", "amiodarone 200mg"]);
+      const qtc = flags.filter((f) => f.rule_id === "DI-QTC-COMBO");
+      expect(qtc).toHaveLength(0);
+    });
+
+    it("betapace (brand) + sotalol (generic)", () => {
+      const flags = checkDrugInteractions(["betapace 80mg", "sotalol 120mg"]);
+      const qtc = flags.filter((f) => f.rule_id === "DI-QTC-COMBO");
+      expect(qtc).toHaveLength(0);
+    });
+
+    it("haldol (brand) + haloperidol (generic)", () => {
+      const flags = checkDrugInteractions(["haldol 5mg", "haloperidol 10mg"]);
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review — PR #650

### Summary

| Area | Verdict | Notes |
|------|---------|-------|
| Correctness | **Pass** | Brand-to-generic normalization correctly prevents same-drug false positives while preserving cross-drug true positives |
| TypeScript / ESM | **Pass** | Strict types (`Record<string, string>`), `.js` import extension, functional style |
| Test coverage | **Pass** | 14 tests covering same-drug brand+generic, two-brand same-generic, cross-drug brand+brand, generic+generic dedup |
| Map completeness | **Pass** | All 46 brand names in `QTC_PATTERN` have entries in `BRAND_TO_GENERIC`; 5 generic-only names (droperidol, erythromycin, procainamide, quinidine, quinine) correctly omitted |
| Performance | **Pass** | `normalizeToGeneric` is O(1) hash lookup, called only on the match-comparison path — negligible overhead |
| Clinical safety | **Pass** | Cross-drug pairs still fire correctly; normalization is conservative (unknown strings pass through unchanged) |

### Detailed Notes

**Strengths**
- Clean separation: normalization map is distinct from the regex, making each independently maintainable
- Fallback via `?? lower` ensures new drugs added to the regex without a map entry still compare correctly (they just won't normalize, which is the safe default — two distinct generics still differ)
- Test matrix is thorough: covers same-drug (brand+generic, brand+brand), different-drug (brand+brand, brand+generic), and generic+generic

**Minor observations (non-blocking)**
1. **Maintenance coupling**: `QTC_PATTERN` and `BRAND_TO_GENERIC` must stay in sync manually. If a new brand name is added to the regex but not the map, the dedup won't fire a false positive (the brand name won't match the generic), but it *will* fire a false positive when paired with itself under the generic name. Consider adding a compile-time or test-time assertion that every non-generic alternate in the regex has a map entry. Filed as follow-up issue.
2. **`droperidol` has no brand name in the regex** — this is fine since Inapsine is rarely used, but worth noting if the regex expands later.

### Verdict: **Approve**

This is a clean, well-scoped fix for a real clinical false-positive bug. Tests are comprehensive. Ship it.

Issue #579, opened 2026-04-16
refactor(types): extract shared AllergyStatus union to @carebridge/types


blamechris:
## Context

PR #563 introduced `AllergyStatus = "nkda" | "unknown" | "has_allergies"` in `apps/clinician-portal/src/lib/allergy-display.ts`. The same union already exists in `packages/ai-prompts/src/clinical-review.ts:60`.

## Problem

Two independent definitions of the same type. If a fourth allergy status is added to the DB (`patients.allergy_status` column), only one module may be updated, causing a silent semantic divergence between the clinician UI and the LLM reviewer.

## Suggested fix

1. Add the type to `packages/types/src/patient.ts` (or a new `allergy.ts`).
2. Import in both `ai-prompts/clinical-review.ts` and `clinician-portal/lib/allergy-display.ts`.
3. Add a Zod enum in `packages/validators` so the tRPC router can validate the value at the boundary.

From review of #563.

PR #654, opened 2026-04-17
refactor(portal): unify AllergyStatus type and runtime Set (#625)

### LINKED ISSUE #625: refactor(portal): unify AllergyStatus type and runtime Set into single source of truth
**Context**: PR #611 introduced `VALID_ALLERGY_STATUSES` (a `ReadonlySet<string>`) alongside the existing `AllergyStatus` type union. Both enumerate the same values (`nkda`, `unknown`, `has_allergies`) independently.

**Problem**: Adding a new allergy status to the type but not the Set (or vice versa) is a silent bug — TypeScript won't catch the mismatch since the Set is typed as `Set<string>`.

**Suggested fix**: Derive one from the other. For example:
```ts
const VALID_ALLERGY_STATUSES = ["nkda", "unknown", "has_allergies"] as const;
export type AllergyStatus = (typeof VALID_ALLERGY_STATUSES)[number];
const statusSet: ReadonlySet<string> = new Set(VALID_ALLERGY_STATUSES);
```

`apps/clinician-portal/src/lib/allergy-display.ts`

blamechris:
## Summary
- Derive `AllergyStatus` type from the `VALID_ALLERGY_STATUSES` const array instead of maintaining both independently, preventing silent drift between the type union and the runtime validation set.
- Replaces the duplicated `ReadonlySet<string>` with a derived `statusSet` built from the single source-of-truth array.

## Test plan
- [x] `pnpm typecheck` passes (all 34 tasks)
- [ ] Existing `allergy-display.test.ts` tests continue to pass
- [ ] Verify allergy display renders correctly in clinician portal

Closes #625

### DIFF
--- a/apps/clinician-portal/src/lib/allergy-display.ts
+++ b/apps/clinician-portal/src/lib/allergy-display.ts
@@ -20,13 +20,9 @@
  * LLM reviewer agree on what an empty list means.
  */
 
-export type AllergyStatus = "nkda" | "unknown" | "has_allergies";
-
-const VALID_ALLERGY_STATUSES: ReadonlySet<string> = new Set<string>([
-  "nkda",
-  "unknown",
-  "has_allergies",
-]);
+const VALID_ALLERGY_STATUSES = ["nkda", "unknown", "has_allergies"] as const;
+export type AllergyStatus = (typeof VALID_ALLERGY_STATUSES)[number];
+const statusSet: ReadonlySet<string> = new Set(VALID_ALLERGY_STATUSES);
 
 /**
  * Runtime guard for allergy_status values coming from the DB or API.
@@ -36,7 +32,7 @@ const VALID_ALLERGY_STATUSES: ReadonlySet<string> = new Set<string>([
 export function parseAllergyStatus(
   value: unknown,
 ): AllergyStatus | null {
-  if (typeof value === "string" && VALID_ALLERGY_STATUSES.has(value)) {
+  if (typeof value === "string" && statusSet.has(value)) {
     return value as AllergyStatus;
   }
   return null;


blamechris:
## Code Review — PR #654

### Summary

| Area | Verdict | Notes |
|------|---------|-------|
| Correctness | Pass | Single source of truth eliminates type/runtime drift |
| TypeScript strict | Pass | `as const` + indexed access is idiomatic TS strict |
| ESM / code style | Pass | No import changes needed; functional style maintained |
| Clinical safety | Pass | `parseAllergyStatus` contract unchanged; `statusSet` is semantically identical to prior `VALID_ALLERGY_STATUSES` |
| Test coverage | Pass | Existing `allergy-display.test.ts` (14 cases) covers all three valid statuses and invalid inputs; CI green |
| Performance | Pass | `Set` constructed once at module scope — no regression |
| Architecture | Note | See below |

### Detailed Findings

**Approve — clean, minimal refactor that does exactly what #625 asked for.**

The change replaces a manually-synchronized type union + `ReadonlySet<string>` with a single `as const` array, derives the type via indexed access, and builds the Set from the same array. This is textbook TS pattern for keeping types and runtime validators in sync.

#### Minor observations (non-blocking)

1. **`statusSet` naming**: The rename from `VALID_ALLERGY_STATUSES` (SCREAMING_SNAKE for a module-level constant) to `statusSet` (camelCase) is a small style shift. Both conventions appear elsewhere in the codebase, so this is fine, but worth noting for consistency if the team has a preference.

2. **Issue #579 remains open**: That issue asks for the `AllergyStatus` type to be extracted to `@carebridge/types` so the duplicate definition in `packages/ai-prompts/src/clinical-review.ts:60` (`allergy_status?: "nkda" | "unknown" | "has_allergies"`) shares the same source. This PR improves the portal-side definition but the cross-package duplication persists. Not a reason to block — #579 is the right venue.

3. **Issue #629 (log warning on unexpected value)**: `parseAllergyStatus` still silently returns `null` for unexpected values. Issue #629 tracks adding a `console.warn`; this PR correctly stays scoped to the type unification.

### CI Status

All checks passing (Build, Test, Typecheck).

PR #657, opened 2026-04-17
fix(ai): thread event timestamp through PatientContext for time-sensitive rules (#587)

### LINKED ISSUE #587: fix(ai): thread event timestamp through PatientContext for time-sensitive rules
## Context
From review of PR #561 (suggestion S1).

## Problem
`isActiveVTEDiagnosis` defaults its `now` parameter to `new Date()` (wall-clock time), but the rule context is built as an event-time snapshot. For VTE onset dates near the 6-month recency boundary, clock skew between event emission and rule processing could produce inconsistent results.

## Proposed fix
Add an optional `event_timestamp?: string` field to `PatientContext`. Populate it from the triggering `ClinicalEvent.timestamp` in both `buildPatientContextForRules` and `buildPatientContext`. Time-sensitive rules like ONCO-VTE-NEURO-001 should use this instead of `new Date()`.

## Acceptance criteria
- `PatientContext.event_timestamp` is populated in both context builders
- `isActiveVTEDiagnosis` uses `event_timestamp` when available
- Tests use a fixed reference date instead of `new Date()`

blamechris:
## Summary
- Adds `event_timestamp` field to `PatientContext` interface so time-sensitive rules use event time instead of wall-clock time
- Populates `event_timestamp` from the validated `triggerEvent.timestamp` in `buildPatientContextForRules`
- Updates `isActiveVTEDiagnosis` (used by ONCO-VTE-NEURO-001) to use `context.event_timestamp` when available, falling back to `new Date()` for backward compatibility
- Fixes all recency-gate tests to use a fixed reference date (`2025-06-15T12:00:00.000Z`) via `event_timestamp` instead of wall-clock-relative date calculations

## Test plan
- [x] All 129 cross-specialty rule tests pass with fixed reference dates
- [x] `pnpm typecheck` passes across all packages

Closes #587


blamechris:
## Code Review — PR #657

### Summary

| Area | Verdict |
|---|---|
| Code quality / TypeScript strict | Pass |
| ESM / import style | Pass |
| Architecture alignment | Pass |
| Clinical data correctness | Pass |
| Test coverage | Pass (minor gap noted) |
| Performance | No concerns |
| Security / HIPAA | N/A (no PHI surface change) |

### Detailed Review

**What this PR does well:**

1. **Correct root-cause fix.** Threading `event_timestamp` through `PatientContext` is the right approach — anchoring time-sensitive rule evaluation to the triggering event rather than worker processing time eliminates a real non-determinism bug. BullMQ workers can have significant queue delay, making wall-clock time unreliable.

2. **Backward compatibility preserved.** The `referenceDate?: Date` parameter with `const now = referenceDate ?? new Date()` fallback means all other callers (and any future rules that skip `event_timestamp`) degrade gracefully. Good defensive design.

3. **Tests use a fixed reference date.** Replacing `new Date()` with `"2025-06-15T12:00:00.000Z"` across all five recency-gate tests eliminates flaky clock-dependent failures — exactly what #587 called for.

4. **Single responsibility.** The change touches only three files with a tight +43/−21 footprint. Easy to audit and low blast radius.

**Observations / suggestions:**

1. **`build-patient-context.test.ts` still uses wall-clock time (line 36).** `const nowIso = new Date().toISOString()` in the test fixture for `buildPatientContextForRules` has the same class of fragility this PR fixes elsewhere. Not blocking since that test does not exercise recency gates, but worth a follow-up to harden it with a fixed timestamp for consistency.

2. **No unit test for `event_timestamp` propagation in `buildPatientContextForRules`.** The `build-patient-context.test.ts` file tests `recent_labs` wiring but does not assert that `event_timestamp` appears in the returned `PatientContext`. A one-line assertion (`expect(ctx.event_timestamp).toBe(event.timestamp)`) would catch accidental regressions. Will file as a follow-up.

3. **Consider validating `event_timestamp` at parse time.** The value flows from the already-validated `eventAt` (via `validateEventTimestamp`), so it is safe today. However, `PatientContext.event_timestamp` is typed as a bare `string?` — if any future code path populates it from a different source, there is no parse-time guard. A `z.string().datetime()` refinement or an inline `Date.parse` assertion would add defense in depth. Not blocking.

4. **Shared `REFERENCE_DATE` constant in tests.** All five test cases independently declare `const referenceDate = "2025-06-15T12:00:00.000Z"`. Extracting this to a top-level constant (`REFERENCE_DATE`) at the describe-block scope would reduce repetition and make it obvious the tests share a single time anchor. Purely cosmetic — not blocking.

### Verdict

**Approve.** Clean fix with good backward compat and solid test coverage. The two follow-up items are minor hardening tasks.

Issue #618, opened 2026-04-17
refactor(clinical): use NFKC normalization in normalizeUnit


blamechris:
## Context
PR #602 added manual `.replace()` for U+00B5 (MICRO SIGN) and U+03BC (GREEK MU) in `normalizeUnit`. This works but is brittle — future Unicode lookalikes (fullwidth chars, combining sequences) won't be caught.

## Proposed change
Add `u.normalize("NFKC")` as the first step in `normalizeUnit` (before `.trim().toLowerCase()`). NFKC decomposes compatibility characters, collapsing µ variants and handling edge cases like fullwidth digits or combining marks from copy-paste.

## Files
- `packages/medical-logic/src/medical-validation.ts` — `normalizeUnit()`

## References
- PR #602
- Issue #574

PR #658, opened 2026-04-17
test(clinical): add positive-path test for µ unit normalization (#619)

### LINKED ISSUE #619: test(clinical): add positive-path test for µ unit normalization
## Context
PR #602 added 3 tests for Unicode micro sign normalization, but all assert `valid: false` (unit mismatch). No test proves that a µ-prefixed unit is *accepted* when it should be.

## Proposed change
Add a test where a lab's `allowed_units` includes `umol/L` and the input is `µmol/L` (U+00B5), asserting `valid: true`. This catches regressions where normalization silently breaks.

## Files
- `packages/medical-logic/src/__tests__/medical-validation.test.ts`

## References
- PR #602
- Issue #574

blamechris:
## Summary
- Adds `umol/L` to Creatinine's `allowed_units` in `COMMON_LAB_TESTS` (clinically correct — SI unit for creatinine)
- Adds a positive-path test asserting that `µmol/L` (U+00B5 MICRO SIGN) normalizes to match `umol/L` and returns `valid: true`
- Closes #619

## Test plan
- [x] New test passes locally (`npx vitest run` in `packages/medical-logic` — 179/179 pass)
- [x] `pnpm typecheck` passes (34/34 tasks)

### DIFF
--- a/packages/medical-logic/src/__tests__/medical-validation.test.ts
+++ b/packages/medical-logic/src/__tests__/medical-validation.test.ts
@@ -375,6 +375,14 @@ describe("validateLabResult", () => {
     expect(microSign.valid).toBe(asciiU.valid);
   });
 
+  it("accepts µmol/L (U+00B5 MICRO SIGN) when allowed_units includes umol/L", () => {
+    // Creatinine allowed_units includes "umol/L" (SI unit). The µ (U+00B5)
+    // in the input should normalise to ASCII 'u', matching the allow-list.
+    const result = validateLabResult("Creatinine", 88.4, "\u00b5mol/L");
+    expect(result.valid).toBe(true);
+    expect(result.errors).toHaveLength(0);
+  });
+
   it("error message quotes the caller's original (non-normalized) unit", () => {
     // Normalisation is only used for comparison; the operator's typed
     // string should surface verbatim so they can see what they entered.
--- a/packages/medical-logic/src/__tests__/medical-validation.test.ts
+++ b/packages/medical-logic/src/__tests__/medical-validation.test.ts
@@ -375,12 +375,17 @@ describe("validateLabResult", () => {
     expect(microSign.valid).toBe(asciiU.valid);
   });
 
-  it("accepts µmol/L (U+00B5 MICRO SIGN) when allowed_units includes umol/L", () => {
-    // Creatinine allowed_units includes "umol/L" (SI unit). The µ (U+00B5)
-    // in the input should normalise to ASCII 'u', matching the allow-list.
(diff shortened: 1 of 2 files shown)


blamechris:
## PR #658 Review — `test(clinical): add positive-path test for µ unit normalization`

### Summary Table

| Area | Verdict | Notes |
|------|---------|-------|
| Code quality | Pass | TypeScript strict, ESM conventions followed |
| Test correctness | **Warn** | Test passes but masks a clinical range-check gap (see below) |
| Clinical safety | **Block** | Missing `UNIT_CONVERSIONS` entry for Creatinine umol/L → mg/dL |
| Architecture | Pass | Change is in the right packages (shared-types + medical-logic) |
| Performance | Pass | Trivial diff, no concerns |

---

### Blocking: Missing Creatinine unit conversion (clinical safety)

Adding `umol/L` to Creatinine's `allowed_units` means the validator will accept creatinine values submitted in µmol/L. However, the canonical unit is **mg/dL** with typical range **0.6–1.2 mg/dL**. Normal creatinine in SI units is approximately **53–106 µmol/L** (conversion factor: ÷88.4).

Without a corresponding entry in `UNIT_CONVERSIONS`:

```ts
const UNIT_CONVERSIONS = {
  // ...existing HbA1c entry...
  Creatinine: {
    "umol/l": (v: number) => v / 88.4,  // µmol/L → mg/dL
  },
};
```

...the validator compares the raw µmol/L value (e.g. 88.4) against the mg/dL typical range (0.6–1.2). A perfectly normal creatinine of 88.4 µmol/L will produce a spurious "above typical range" warning. This is the exact class of sentinel-event confusion the code comments at line 222–225 of `clinical-data.ts` were written to prevent.

The new test only asserts `valid: true` and `errors: []`, so it doesn't catch this — the unit passes the allow-list check (no error), but the range comparison is clinically meaningless without conversion.

**Recommended fix**: Add the `UNIT_CONVERSIONS` entry for Creatinine and extend the test to also assert `warnings` is empty (or at least doesn't contain a range warning for a normal value).

### Minor observations

1. **Test value choice**: 88.4 µmol/L is a nice choice (≈1.0 mg/dL), but without conversion it actually exposes the gap. Consider adding an assertion like `expect(result.warnings).toHaveLength(0)` — this will fail until the conversion is added, which is the correct behavior.

2. **Issue #618** (NFKC normalization) is not resolved by this PR and remains open — confirmed, no action needed.

3. **Issue #619** is directly addressed. Once the conversion gap is fixed, closing #619 is appropriate.

---

**Verdict: Request changes** — the `allowed_units` expansion without a matching `UNIT_CONVERSIONS` entry creates a silent clinical data misinterpretation path.

PR #661, opened 2026-04-17
fix(ai): add retraction predicates for labs and procedures

### LINKED ISSUE #638: fix(ai): add retraction predicates for labs and procedures
Diagnoses (#515), allergies (#515), and medications (#581/#609) now have retraction predicates that exclude `entered_in_error` rows from rule and LLM paths. Labs and procedures do not yet have equivalent guards.

If a lab result or procedure is marked `entered_in_error` in FHIR, it will still appear in the patient context and could trigger false-positive clinical flags. Add `isLabRetracted` and `isProcedureRetracted` predicates to `event-time-snapshot.ts` and wire them into both filter sites.

Ref: #609, #515

blamechris:
## Summary
- Add `isLabRetracted` and `isProcedureRetracted` predicates to `event-time-snapshot.ts`, matching the existing pattern for diagnoses, allergies, and medications
- Wire `isLabRetracted` into both the LLM context path (`context-builder.ts`) and the deterministic rule path (`review-service.ts`) so `entered_in_error` lab rows are excluded from clinical review
- `isLabRetracted` checks both `status` and `flag` fields since `lab_results` currently uses `flag` rather than a dedicated `status` column
- `isProcedureRetracted` checks `status` on `procedures` (which has an existing status column); not yet wired into filter sites since procedures aren't currently queried in the review pipeline
- Add unit tests for both new predicates

Closes #638

## Test plan
- [x] `isLabRetracted` returns `true` for `status=entered_in_error` and `flag=entered_in_error`
- [x] `isLabRetracted` returns `false` for normal flag values (H, L), other statuses, null, missing
- [x] `isProcedureRetracted` returns `true` only for `status=entered_in_error`
- [x] `isProcedureRetracted` returns `false` for scheduled, completed, cancelled, null, missing
- [x] All 18 event-time-snapshot tests pass
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

### DIFF
--- a/services/ai-oversight/src/__tests__/event-time-snapshot.test.ts
+++ b/services/ai-oversight/src/__tests__/event-time-snapshot.test.ts
@@ -6,6 +6,8 @@ import {
   isDiagnosisRetracted,
   isAllergyRetracted,
   isMedicationRetracted,
+  isLabRetracted,
+  isProcedureRetracted,
 } from "../utils/event-time-snapshot.js";
 
 describe("event-time-snapshot helpers", () => {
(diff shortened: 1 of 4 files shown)


blamechris:
## PR #661 Review — `fix(ai): add retraction predicates for labs and procedures`

### Summary

| Area | Verdict | Notes |
|------|---------|-------|
| Code quality | Pass | TypeScript strict, ESM `.js` imports, functional style, no classes |
| Architecture alignment | Pass | Follows established predicate pattern from #515 (diagnoses, allergies, meds) |
| Testing | Pass | 10 new test cases cover both predicates; positive, negative, null, and missing-field paths |
| Clinical data safety | Pass | Retracted labs excluded from both rule and LLM paths; prevents false clinical flags |
| Performance | Pass | Filter is O(n) in-memory check on already-fetched rows; negligible cost |

### Detailed Findings

**1. Correctness — both filter sites covered (Pass)**
`isLabRetracted` is wired into:
- `context-builder.ts` line 240 (LLM path) — filters before `isoLTE` timestamp check
- `review-service.ts` line 641 (rule path) — filters inside the dedup loop

Both paths now share the same retraction semantics, which is the explicit goal of the `event-time-snapshot` module. The dual `status || flag` check in `isLabRetracted` is well-motivated by the comment: `lab_results` currently lacks a dedicated `status` column, so `flag` is the only carrier today, but checking both future-proofs the predicate.

**2. `flag` column now projected in the rule path (Pass)**
The `select()` in `buildPatientContextForRules` previously omitted `flag`, which would have caused `isLabRetracted` to always return `false` in that path. Adding `flag: labResults.flag` to the projection fixes this. Good catch.

**3. `isProcedureRetracted` — declared but not wired (Acceptable, noted)**
The PR summary explicitly states procedures are "not yet wired into filter sites since procedures aren't currently queried in the review pipeline." This is honest and the predicate will be ready when procedure review is added. No issue needed — the predicate exists as a forward declaration alongside its tests.

**4. Test coverage (Pass)**
- `isLabRetracted`: 6 assertions covering `status`, `flag`, both, normal flags (H/L), other statuses, null, empty
- `isProcedureRetracted`: 5 assertions covering entered_in_error, scheduled, completed, cancelled, null, empty
- Tests follow the existing describe/it pattern in the file
- The test file imports are updated with correct `.js` ESM extensions

**5. Module-header docblock updated (Pass)**
The retraction section in `event-time-snapshot.ts` lines 18-22 now lists "labs, procedures" alongside diagnoses and medications. Keeps the module documentation accurate.

**6. Minor observation — filter ordering in context-builder**
In `context-builder.ts` the retraction check `!isLabRetracted(r)` is placed before `isoLTE(r.created_at, eventAt)`. This is fine (short-circuit `&&`), though reversing the order would skip the string field comparison for retracted rows. The performance difference is negligible on a max-50 result set, so this is purely cosmetic — no action needed.

**7. Minor observation — `isLabRetracted` test imports**
The test file imports `isLabRetracted` and `isProcedureRetracted` but does not import them into the top-level `import` alongside the timestamp helpers. It would be slightly cleaner to group all predicate imports together, but the current placement (appended after `isMedicationRetracted`) follows alphabetical/addition order and is fine.

### Verdict

**Approve.** Clean, well-scoped fix that extends an established pattern to labs and procedures. Both critical filter sites (LLM and rule paths) are covered. Tests are thorough. No clinical safety concerns.

This PR resolves from-review issue #638.

PR #662, opened 2026-04-17
test(ai): add idempotency probe test for DB-clock interval derivation (#627)

### LINKED ISSUE #627: test(ai): integration test for DB-clock idempotency probe

blamechris:
## Summary
- Export `IN_FLIGHT_WINDOW_MS` and derive the SQL interval from the constant (replacing the hardcoded `'150 seconds'` literal) so the two values cannot drift apart
- Add `review-service-db-clock-idempotency.test.ts` with 6 tests validating the constant's properties and the SQL interval derivation at runtime
- Fix the `sql` mock in 3 existing test files (`review-idempotency`, `review-service-fallback`, `llm-timeout-fallback`) to include `sql.raw()`, which the derivation now requires

## Test plan
- [x] New test file passes: 6/6 tests green
- [x] Existing idempotency tests pass: 10/10 tests green
- [x] Existing fallback tests pass: 13/13 tests green (review-service-fallback + llm-timeout-fallback)
- [x] review-service.test.ts unaffected: 5/5 tests green

Closes #627

### DIFF
--- a/services/ai-oversight/src/__tests__/llm-timeout-fallback.test.ts
+++ b/services/ai-oversight/src/__tests__/llm-timeout-fallback.test.ts
@@ -71,15 +71,21 @@ vi.mock("@carebridge/db-schema", () => ({
   careTeamAssignments: { patient_id: "patient_id", user_id: "user_id" },
 }));
 
-vi.mock("drizzle-orm", () => ({
-  eq: vi.fn((...args: unknown[]) => args),
-  desc: vi.fn((col: unknown) => col),
-  gte: vi.fn((...args: unknown[]) => args),
-  and: vi.fn((...args: unknown[]) => args),
-  or: vi.fn((...args: unknown[]) => args),
-  inArray: vi.fn((...args: unknown[]) => args),
-  sql: vi.fn(),
-}));
+vi.mock("drizzle-orm", () => {
+  const sqlTag = vi.fn((_strings: TemplateStringsArray, ..._values: unknown[]) => ({
+    __sql: true,
+  }));
+  (sqlTag as unknown as Record<string, unknown>).raw = vi.fn((v: string) => ({ __raw: v }));
+  return {
+    eq: vi.fn((...args: unknown[]) => args),
+    desc: vi.fn((col: unknown) => col),
+    gte: vi.fn((...args: unknown[]) => args),
+    and: vi.fn((...args: unknown[]) => args),
+    or: vi.fn((...args: unknown[]) => args),
+    inArray: vi.fn((...args: unknown[]) => args),
+    sql: sqlTag,
+  };
+});
 
 vi.mock("@carebridge/ai-prompts", () => ({
   CLINICAL_REVIEW_SYSTEM_PROMPT: "system prompt",
--- a/services/ai-oversight/src/__tests__/review-idempotency.test.ts
+++ b/services/ai-oversight/src/__tests__/review-idempotency.test.ts
@@ -42,15 +42,21 @@ vi.mock("@carebridge/db-schema", () => ({
   encounters: {},
 }));
 
-vi.mock("drizzle-orm", () => ({
-  eq: vi.fn(),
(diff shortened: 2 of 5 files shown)


blamechris:
## Code Review — PR #662

| Area | Verdict | Notes |
|------|---------|-------|
| Code quality | **Pass** | TypeScript strict, ESM `.js` imports, clean functional style |
| Architecture | **Pass** | Single-source-of-truth derivation eliminates hardcoded drift |
| Testing | **Pass** | 6 new tests + 3 existing mock fixes; good coverage of invariants |
| Clinical data rules | **N/A** | No rule logic changed |
| Performance | **Pass** | `sql.raw` inlined at query build time, zero runtime cost |
| Security | **Pass** | `sql.raw(String(windowSec))` is safe — derived from a hardcoded integer constant, no user input |

### Production change (review-service.ts)

**Good:**
- Exporting `IN_FLIGHT_WINDOW_MS` is the right call — the constant was already documented as a tuning knob, and tests need to reference it.
- `Math.round(IN_FLIGHT_WINDOW_MS / 1000)` + `sql.raw(String(windowSec))` correctly derives the interval from the constant. The `Math.round` guard is defensive but harmless since the constant is `150_000` (exactly divisible by 1000).
- This also closes the drift bug described in issue #622.

**Nit (non-blocking):** `windowSec` is computed inside `processReviewJob`, meaning it runs on every invocation. Since `IN_FLIGHT_WINDOW_MS` is a module-level constant, the derivation could be hoisted to module scope:
```ts
export const IN_FLIGHT_WINDOW_MS = 150_000;
const IN_FLIGHT_WINDOW_SEC = Math.round(IN_FLIGHT_WINDOW_MS / 1000);
```
This is purely cosmetic — the computation is trivial — but it makes the relationship between the two values visible at the module level and eliminates the repeated allocation.

### New test file (review-service-db-clock-idempotency.test.ts)

**Good:**
- The three `IN_FLIGHT_WINDOW_MS constant` tests validate the contract well: positive integer, >= BullMQ lockDuration, and whole-number seconds conversion.
- The `sql.raw()` capture approach is clever — intercepting the mock to verify the derivation actually flows through `sql.raw` at call time.
- The `NOW() - interval` template assertion confirms the SQL shape is correct.

**Observation (non-blocking):** The third test in the "DB-clock interval derivation" describe block (`interval seconds match IN_FLIGHT_WINDOW_MS / 1000`) hardcodes `expect(expectedSec).toBe(150)`. This makes it a snapshot of the current value rather than a drift-detection test. If someone changes `IN_FLIGHT_WINDOW_MS` to 180_000 this test fails even though the derivation is correct. Consider dropping the `toBe(150)` assertion and keeping only the `Number.isInteger` check — the `sql.raw` capture test already validates that the derived value actually reaches the SQL template.

### Mock updates (3 existing test files)

The `sql` mock upgrade from `vi.fn()` to a callable + `.raw` property is correct and consistent across all three files. The pattern:
```ts
const sqlTag = vi.fn((...) => ({ __sql: true }));
(sqlTag as unknown as Record<string, unknown>).raw = vi.fn((v: string) => ({ __raw: v }));
```
works but the cast chain is a bit noisy. A minor improvement would be `Object.assign(sqlTag, { raw: vi.fn(...) })` — but this is stylistic and not worth a revision.

### Verdict

**Approve.** Clean, well-scoped change that eliminates a real drift hazard and adds solid test coverage. The two nits (module-level derivation, hardcoded `150` assertion) are non-blocking and can be addressed in a follow-up.

PR #678, opened 2026-04-17
test(ai): remove hardcoded 150 assertion in db-clock idempotency test

### LINKED ISSUE #672: test(ai): remove hardcoded 150 assertion in db-clock idempotency test
## Context
In PR #662, the test `interval seconds match IN_FLIGHT_WINDOW_MS / 1000` contains:
```ts
expect(expectedSec).toBe(150);
```

This hardcodes the current value of `IN_FLIGHT_WINDOW_MS / 1000`, making the test a snapshot rather than a drift-detection invariant. If someone tunes `IN_FLIGHT_WINDOW_MS` to 180_000 this test fails even though the derivation is correct.

## Proposed fix
Drop the `toBe(150)` assertion and keep only `Number.isInteger(expectedSec)`. The `sql.raw` capture test already validates that the derived value reaches the SQL template.

Ref: PR #662

blamechris:
## Summary
- Removes the hardcoded `expect(expectedSec).toBe(150)` assertion in the db-clock idempotency test
- Keeps the `Number.isInteger` check so the test still validates the constant produces a whole-second interval without being brittle to value changes

Closes #672

## Test plan
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

### DIFF
--- a/services/ai-oversight/src/__tests__/review-service-db-clock-idempotency.test.ts
+++ b/services/ai-oversight/src/__tests__/review-service-db-clock-idempotency.test.ts
@@ -212,7 +212,6 @@ describe("DB-clock interval derivation in idempotency probe", () => {
     // vice versa), this test catches the drift because the derivation now
     // uses the constant directly.
     const expectedSec = IN_FLIGHT_WINDOW_MS / 1000;
-    expect(expectedSec).toBe(150);
     expect(Number.isInteger(expectedSec)).toBe(true);
   });
 });


blamechris (commented):
**Inspector Review — PR #678**

| Check | Status |
|---|---|
| Correctness | PASS — removes brittle `toBe(150)` while keeping the `Number.isInteger` invariant that actually matters |
| Test coverage | PASS — existing assertions (positive integer, >= lockDuration, whole-second conversion, sql.raw embedding, NOW()-interval template) remain intact |
| Commit format | PASS — `test(ai):` with clear scope |
| Zero Attribution | PASS — no violations |
| CI | PASS — Build, Test, Typecheck all green |
| Security | N/A — test-only change |

Closes #672 as stated. No deferred items.

PR #682, opened 2026-04-17
test(gateway): harden mock thenable to respect selectColumns filtering

### LINKED ISSUE #630: test(gateway): harden mock thenable to respect selectColumns filtering
In `patient-records-projection.test.ts` (PR #604), the mock's `.from` chain is double-wrapped with a `.then` that resolves `resolvedData` unfiltered, bypassing the `selectColumns`-based row filtering that only occurs in `.where()`. If future production code awaits `db.select().from(table)` without `.where()`, the test would silently pass with the full patient row including sensitive fields.

**Fix:** Make the `.then` on `.from` also apply `selectColumns` filtering before resolving, matching the behavior in `.where()`.

Ref: PR #604

blamechris:
## Summary
- Fixes #630
- The mock `.from` chain's `.then` handler was resolving raw `resolvedData` without applying `selectColumns` filtering, which meant bare `await db.select(cols).from(table)` calls bypassed column projection in tests
- Updated `.then` on `.from` to apply the same `selectColumns` key-filtering logic already used in `.where`, ensuring the mock faithfully simulates Drizzle column selection

## Test plan
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] `pnpm test` in `services/api-gateway` confirms projection tests still pass with the hardened mock

### DIFF
--- a/services/api-gateway/src/__tests__/patient-records-projection.test.ts
+++ b/services/api-gateway/src/__tests__/patient-records-projection.test.ts
@@ -74,7 +74,20 @@ const mocks = vi.hoisted(() => {
       (result as Record<string, unknown>).then = (
         resolve: (v: unknown) => void,
       ) => {
-        resolve(resolvedData);
+        if (selectColumns) {
+          const keys = Object.keys(selectColumns);
+          resolve(
+            resolvedData.map((row) => {
+              const out: Record<string, unknown> = {};
+              for (const k of keys) {
+                out[k] = (row as Record<string, unknown>)[k];
+              }
+              return out;
+            }),
+          );
+        } else {
+          resolve(resolvedData);
+        }
         return result;
       };
       return result;


blamechris (commented):
**Review: Approve**

The fix correctly mirrors the `selectColumns` filtering that `.where` already applies, so bare `await db.select(cols).from(table)` paths no longer bypass column projection in tests. CI is green (Typecheck, Build, Test all pass).

Two minor observations (neither blocks merge):

1. **Duplicated filtering logic** -- the `selectColumns` key-filtering block now appears identically in both `.where` and `.then`. Consider extracting a small `applyProjection(rows, selectColumns)` helper inside `vi.hoisted` to keep the mock DRY.

2. **`.limit` mock still returns raw `resolvedData`** (line 68) -- if any code path chains `.limit()` after a projected `.select(cols)`, it will return unfiltered rows, the same class of bug this PR fixes for `.then`. Filing a from-review issue for this.

PR #685, opened 2026-04-17
test(ai): sync assertion for QTC_PATTERN vs BRAND_TO_GENERIC

### LINKED ISSUE #668: test(ai): add compile-time sync assertion for QTC_PATTERN vs BRAND_TO_GENERIC map
Found in PR #650 review.

`QTC_PATTERN` and `BRAND_TO_GENERIC` must stay in sync manually. If a new brand name is added to the regex but not the map, same-drug brand+generic pairs will produce false-positive DI-QTC-COMBO flags — the exact bug #635 fixed.

Add a test that:
1. Extracts all alternates from `QTC_PATTERN` source
2. Partitions them into generics (values of `BRAND_TO_GENERIC` + unmapped names) and brands (keys of `BRAND_TO_GENERIC`)
3. Asserts every alternate is accounted for in one of those two sets
4. Asserts every key in `BRAND_TO_GENERIC` appears in `QTC_PATTERN`

This prevents future regressions when the drug list is expanded.

blamechris:
## Summary
- Export `QTC_PATTERN` and `BRAND_TO_GENERIC` from `drug-interactions.ts` so they can be tested directly
- Add three sync-assertion tests that verify every `BRAND_TO_GENERIC` key appears in `QTC_PATTERN` alternates, every generic value appears in the alternates, and every brand in the pattern maps back correctly
- Guards against drift between the regex and the normalization map

Closes #668

## Test plan
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] All 17 tests in `drug-interaction-rules.test.ts` pass (14 existing + 3 new)

### DIFF
--- a/services/ai-oversight/src/__tests__/drug-interaction-rules.test.ts
+++ b/services/ai-oversight/src/__tests__/drug-interaction-rules.test.ts
@@ -1,6 +1,44 @@
 import { describe, it, expect } from "vitest";
 
-import { checkDrugInteractions } from "../rules/drug-interactions.js";
+import {
+  checkDrugInteractions,
+  QTC_PATTERN,
+  BRAND_TO_GENERIC,
+} from "../rules/drug-interactions.js";
+
+describe("QTC_PATTERN ↔ BRAND_TO_GENERIC sync", () => {
+  const alternates = QTC_PATTERN.source.split("|");
+
+  it("every BRAND_TO_GENERIC key appears in QTC_PATTERN alternates", () => {
+    for (const brand of Object.keys(BRAND_TO_GENERIC)) {
+      expect(alternates).toContain(brand);
+    }
+  });
+
+  it("every BRAND_TO_GENERIC value (generic) appears in QTC_PATTERN alternates", () => {
+    const genericValues = [...new Set(Object.values(BRAND_TO_GENERIC))];
+    for (const generic of genericValues) {
+      expect(alternates).toContain(generic);
+    }
+  });
+
+  it("every alternate is either a generic (not in BRAND_TO_GENERIC keys) or a known brand", () => {
+    const brandKeys = new Set(Object.keys(BRAND_TO_GENERIC));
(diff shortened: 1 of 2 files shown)


blamechris (commented):
## Inspector Review — PR #685

**CI**: Build, Test, Typecheck all pass.
**Commit format**: `test(ai): ...` — correct type/scope per project standards.
**Zero Attribution Policy**: Clean.
**Closes**: #668 — directly addresses the from-review issue.

### Findings

**1. Tautological assertion in third test (bug)**

The third test ("every alternate is either a generic or a known brand") contains a tautology:

```ts
const isBrand = brandKeys.has(alt);
const isGeneric = !brandKeys.has(alt);
// ...
expect(isBrand || isGeneric).toBe(true); // always true
```

Since `isGeneric` is defined as `!isBrand`, the final assertion `isBrand || isGeneric` is logically equivalent to `true` — it can never fail. The test *appears* to verify that orphan alternates (typos that are neither a brand key nor a known generic value) are caught, but it actually catches nothing.

The fix: `isGeneric` should check `genericValues.has(alt)` instead of `!brandKeys.has(alt)`:

```ts
const isGeneric = genericValues.has(alt);
expect(isBrand || isGeneric).toBe(true);
```

This way, an alternate like `"amiodaroone"` (typo) that is neither a brand key nor a generic value would correctly fail the test.

The inner `if (isBrand)` block does add value (verifying the mapping target exists in the pattern), but the final expect is dead weight.

**2. Export surface expansion is minimal and appropriate**

Exporting `QTC_PATTERN` and `BRAND_TO_GENERIC` is the lightest-touch way to enable the sync tests. Both are `const` and immutable at runtime. No concern here.

**3. First two tests are well-structured**

The "every brand key in pattern" and "every generic value in pattern" tests correctly guard against drift in both directions.

### Verdict

Request the tautology fix (item 1) before merge — without it, the third test gives false confidence and does not fulfill the "every alternate is accounted for" requirement from #668 point 3. The other two tests are solid.

PR #686, opened 2026-04-17
test(portal): use fake timers in vitals-staleness boundary tests

### LINKED ISSUE #631: test(portal): use fake timers in vitals-staleness boundary tests

blamechris:
## Summary
- Pin the clock with `vi.useFakeTimers()` + `vi.setSystemTime()` so boundary tests in `vitals-staleness.test.ts` are deterministic and not sensitive to real-time drift between `Date.now()` calls
- Derive all timestamps from the pinned `PINNED_NOW` constant instead of calling `Date.now()` at test runtime
- Fix misleading test description: "invalid ISO string results in stale" now correctly reads "falls through to current (NaN comparisons)"

Closes #631

## Test plan
- [ ] `pnpm typecheck` passes
- [ ] `pnpm lint` passes
- [ ] `pnpm test --filter=@carebridge/clinician-portal` passes with all vitals-staleness tests green

### DIFF
--- a/apps/clinician-portal/src/__tests__/vitals-staleness.test.ts
+++ b/apps/clinician-portal/src/__tests__/vitals-staleness.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, vi, afterEach } from "vitest";
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
 import { formatAge, classifyStaleness } from "../lib/vitals-staleness.js";
 
 const SECOND = 1_000;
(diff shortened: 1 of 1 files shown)


blamechris (commented):
## Review: APPROVED

Clean, well-scoped fix for #631. The fake-timer approach is the right solution for deterministic boundary tests.

### What's good
- `vi.useFakeTimers()` / `vi.setSystemTime()` in `beforeEach` correctly pins the clock for both `formatAge` and `classifyStaleness` test suites
- `PINNED_NOW` constant and updated `ago()` helper derive all timestamps from the pinned clock, eliminating wall-clock drift
- `vi.useRealTimers()` in `afterEach` properly restores real timers (and replaces the previous `vi.restoreAllMocks()` which was not actually needed for timer cleanup)
- The misleading test description is corrected: NaN comparisons fall through to "current", not "stale"

### Deferred items
- **`classifyStaleness` silently returns "current" for invalid input**: The NaN fall-through behavior is now correctly documented in the test, but in a clinical context this is a latent bug -- garbage timestamps should not be classified as "current" vitals. The production code should guard with `Number.isNaN(ageMs)` (like `formatAge` already does) and return "stale" or a distinct tier. Filing a from-review issue.

PR #687, opened 2026-04-17
test(gateway): add family_caregiver to HIPAA projection regression test

### LINKED ISSUE #634: test(gateway): add family_caregiver to HIPAA projection regression test
PR #605 introduced a parameterized regression test that verifies all roles exclude sensitive columns from the `patients.list` projection. The test covers `admin`, `patient`, `physician`, `nurse`, and `specialist` but omits `family_caregiver`.

Since the family_caregiver path was one of the branches that previously used an un-projected `db.select().from(patients)`, it is the most important role to cover in the regression suite.

**Action:** Add `family_caregiver` to the role iteration array in the `excludes sensitive columns from list response for ${role} role` parameterized test.

Ref: PR #605

blamechris:
## Summary
- Adds `family_caregiver` to the parameterized role iteration array in the HIPAA minimum-necessary projection regression test
- Includes proper setup with `setResolvedDataQueue` for the family_caregiver multi-query flow and checks the correct (3rd) select call for excluded fields

Closes #634

## Test plan
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] CI passes

### DIFF
--- a/services/api-gateway/src/__tests__/patient-records-list-rbac.test.ts
+++ b/services/api-gateway/src/__tests__/patient-records-list-rbac.test.ts
@@ -385,8 +385,29 @@ describe("patientRecordsRbacRouter.list — HIPAA minimum-necessary filtering", 
     "notes",
   ] as const;
 
-  for (const role of ["admin", "patient", "physician", "nurse", "specialist"] as const) {
+  for (const role of ["admin", "patient", "physician", "nurse", "specialist", "family_caregiver"] as const) {
     it(`excludes sensitive columns from list response for ${role} role`, async () => {
+      if (role === "family_caregiver") {
+        const CAREGIVER_ID = "77777777-7777-4777-8777-777777777777";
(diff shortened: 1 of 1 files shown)


blamechris (commented):
**Inspector Review: PASS**

Clean, focused test addition that closes the coverage gap identified in #634.

**What looks good:**
- Correctly queues three resolved-data entries matching the family_caregiver multi-query flow (familyRelationships -> users -> patients)
- Asserts against `mock.calls[2]` (the 3rd select), which is the actual patients projection -- matches the production code path
- Early `return` keeps the generic path untouched for other roles
- Type cast `as User["role"]` is consistent with the existing caregiver tests above (line 350)

**CI note:** The failing test (`vitals-staleness.test.ts > boundary: exactly 24h is overdue`) is a pre-existing flaky boundary test unrelated to this PR. Tracked by #631.

**No blocking issues. No deferred items.**

PR #689, opened 2026-04-17
test(gateway): projection regression guard in afterEach (#633)

### LINKED ISSUE #633: test(gateway): add negative regression guard for projection revert
PR #604's test plan includes "Verify tests fail when projection is reverted to `db.select().from(patients)`" but this was left unchecked and is not automated.

**Suggestion:** Add a dedicated test or inline assertion that `mocks.mockDb.select` was called with a non-empty column map argument (`expect(mocks.getSelectColumns()).toBeDefined()`), which already exists in two tests but could be consolidated into a shared `afterEach` guard to catch any new test that accidentally skips the check.

Ref: PR #604

blamechris:
## Summary
- Adds a shared `afterEach` guard (`assertProjectionWasApplied`) to both `getById` and `getSummary` projection test suites
- Every test now automatically verifies that `db.select()` received a non-empty column map with no sensitive fields, preventing silent regressions if someone reverts the projection to a bare `select()`
- Consolidates redundant assertions from dedicated column-map tests into the shared guard while keeping endpoint-specific shape checks

## Test plan
- [ ] `pnpm typecheck` passes
- [ ] `pnpm lint` passes
- [ ] Existing projection tests continue to pass
- [ ] Removing the column map from `db.select()` in the router causes afterEach failures across all tests in both suites

Closes #633

### DIFF
--- a/services/api-gateway/src/__tests__/patient-records-projection.test.ts
+++ b/services/api-gateway/src/__tests__/patient-records-projection.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, vi, beforeEach } from "vitest";
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
 import type { User } from "@carebridge/shared-types";
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


blamechris (commented):
## Review — PR #689

**Verdict: Approve**

### What changed
Single-file change to `services/api-gateway/src/__tests__/patient-records-projection.test.ts` that adds a shared `afterEach` guard (`assertProjectionWasApplied`) to both `getById` and `getSummary` projection test suites. The guard asserts that every test exercising these endpoints ends with `db.select()` having received a non-empty column map with no sensitive fields. Redundant assertions from individual column-map tests are consolidated into the guard; endpoint-specific shape checks are preserved.

### Checklist
- [x] Commit format: `test(gateway): ...` — correct type + scope
- [x] Zero Attribution Policy: no violations
- [x] CI: Build, Test, Typecheck all pass
- [x] ESM / TypeScript strict: no issues
- [x] Functional style: helper function, no classes
- [x] Closes #633 as stated

### Notes
- Clean, well-scoped change — no production code touched, only test harness
- The `selectCols!` non-null assertions in individual tests are safe because the `afterEach` guard runs `toBeDefined()` first, so any undefined case will already fail
- No deferred items identified; no from-review issues created or closed

PR #693, opened 2026-04-17
refactor(notifications): cache resolveCategoryLabel in delayed path

### LINKED ISSUE #640: refactor(notifications): cache resolveCategoryLabel in processDelayedNotification path
## Context
PR #614 de-duplicated `resolveCategoryLabel` within `processNotificationJob`, but `processDelayedNotification` still calls it independently per delayed-notification job. For batch delayed processing of the same event category, the label resolution and potential unknown-category warning repeat for every job.

## Proposed
Consider passing the resolved label through the delayed-job payload (or a lightweight memo) so `resolveCategoryLabel` runs at most once per original event, even across the delayed path.

## Files
- `services/notifications/src/workers/dispatch-worker.ts` (lines ~358-362)

blamechris:
## Summary
- Pass the already-resolved `categoryLabel` through the delayed-job payload (`_resolved_category_label`) so `processDelayedNotification` reuses it instead of calling `resolveCategoryLabel` independently
- Falls back to `resolveCategoryLabel` if the cached field is absent (backward-compatible with any in-flight delayed jobs from before this change)
- Eliminates duplicate warning logs for unknown categories on the delayed path

Closes #640

## Test plan
- [ ] Verify typecheck and lint pass (`pnpm typecheck && pnpm lint`)
- [ ] Trigger a notification for a user in quiet hours; confirm the delayed job payload includes `_resolved_category_label`
- [ ] Confirm unknown categories log a warning only once (on the initial dispatch), not again on the delayed path

### DIFF
--- a/services/notifications/src/workers/dispatch-worker.ts
+++ b/services/notifications/src/workers/dispatch-worker.ts
@@ -287,6 +287,7 @@ async function processNotificationJob(event: NotificationEvent): Promise<number>
         {
           ...event,
           _targeted_user_id: userId,
+          _resolved_category_label: categoryLabel,
         },
         { delay: decision.delay_ms },
       );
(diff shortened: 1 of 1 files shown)


blamechris (commented):
## Inspector Review -- PR #693

Clean, minimal change that correctly addresses the duplicate `resolveCategoryLabel` call on the delayed path.

**Reviewed:**
- Payload threading: `_resolved_category_label` is set alongside `_targeted_user_id` in the delayed-job creation (line 290), consistent with the existing underscore-prefixed internal field pattern.
- Backward compatibility: The nullish coalescing fallback (`event._resolved_category_label ?? resolveCategoryLabel(event.category)`) ensures in-flight delayed jobs queued before this deploy still resolve correctly.
- Type narrowing: The optional field is correctly added to both the `processDelayedNotification` parameter type and the worker's top-level job cast.
- Serialization: String values survive BullMQ's JSON round-trip without issue.

**One deferred item (non-blocking):**
The inline type intersection `NotificationEvent & { _targeted_user_id?: string; _resolved_category_label?: string }` is now repeated across the worker cast (line 426) and the function signature (line 364). As internal payload fields accumulate, extracting a named `DelayedNotificationPayload` type alias would reduce duplication and make the contract easier to grep. Filed as a from-review issue.

**Verdict:** Approve -- no blocking issues found.

PR #707, opened 2026-04-17
fix(ai): correct tautological assertion in QTC_PATTERN sync test

### LINKED ISSUE #701: fix(ai): correct tautological assertion in QTC_PATTERN sync test
Found in PR #685 review.

The third test in the `QTC_PATTERN ↔ BRAND_TO_GENERIC sync` describe block contains a tautological assertion:

```ts
const isBrand = brandKeys.has(alt);
const isGeneric = !brandKeys.has(alt);  // always !isBrand
expect(isBrand || isGeneric).toBe(true); // always true
```

`isGeneric` should check membership in `genericValues` rather than being the negation of `isBrand`:

```ts
const isGeneric = genericValues.has(alt);
expect(isBrand || isGeneric).toBe(true);
```

Without this fix, an orphan alternate (e.g. a typo like `"amiodaroone"`) would pass silently, defeating the purpose of the sync guard from #668 requirement 3.

blamechris:
## Summary
- Fix tautological assertion in the third QTC_PATTERN sync test where `isGeneric = !brandKeys.has(alt)` was always the negation of `isBrand`, making `isBrand || isGeneric` trivially true
- Changed `isGeneric` to check `genericValues.has(alt)` so the assertion actually validates that every alternate in `QTC_PATTERN` is either a known brand key or a known generic value

Closes #701

## Test plan
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

### DIFF
--- a/services/ai-oversight/src/__tests__/drug-interaction-rules.test.ts
+++ b/services/ai-oversight/src/__tests__/drug-interaction-rules.test.ts
@@ -28,7 +28,7 @@ describe("QTC_PATTERN ↔ BRAND_TO_GENERIC sync", () => {
 
     for (const alt of alternates) {
       const isBrand = brandKeys.has(alt);
-      const isGeneric = !brandKeys.has(alt);
+      const isGeneric = genericValues.has(alt);
       // If it's a brand, it must map to a generic that's also in the pattern
       if (isBrand) {
         expect(genericValues).toContain(BRAND_TO_GENERIC[alt]);
--- a/services/ai-oversight/src/__tests__/drug-interaction-rules.test.ts
+++ b/services/ai-oversight/src/__tests__/drug-interaction-rules.test.ts
@@ -24,7 +24,12 @@ describe("QTC_PATTERN ↔ BRAND_TO_GENERIC sync", () => {
 
   it("every alternate is either a generic (not in BRAND_TO_GENERIC keys) or a known brand", () => {
     const brandKeys = new Set(Object.keys(BRAND_TO_GENERIC));
-    const genericValues = new Set(Object.values(BRAND_TO_GENERIC));
+    // Generic names from BRAND_TO_GENERIC values, plus standalone generics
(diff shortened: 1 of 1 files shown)


blamechris (commented):
## Review: fix(ai): correct tautological assertion in QTC_PATTERN sync test

### Verdict: Changes Needed

The diagnosis is correct — `isGeneric = !brandKeys.has(alt)` is indeed tautological and the fix to `genericValues.has(alt)` is the right approach. However, **CI is failing** because the fix has revealed a real data gap.

### Blocking: CI Test failure (must fix before merge)

The corrected assertion now catches alternates in `QTC_PATTERN` that are standalone generics with no brand-name entry in `BRAND_TO_GENERIC`. These drugs appear in the regex but are not values of any brand mapping because they have no brand alias defined:

- `quinidine`
- `procainamide`
- `droperidol`
- `erythromycin`
- `quinine`

Because `genericValues` is built from `Object.values(BRAND_TO_GENERIC)`, these standalone generics are not in `genericValues` and not in `brandKeys`, so `isBrand || isGeneric` evaluates to `false`.

**Two possible fixes** (pick one):

1. **Add identity mappings** in `BRAND_TO_GENERIC` for standalone generics (e.g. `quinidine: "quinidine"`). This keeps the test assertion clean and ensures `toGeneric()` works uniformly.

2. **Derive a complete generic set** in the test by computing generics as alternates that are NOT brand keys (the old behavior), then intersect that with `genericValues` only for drugs that *have* a brand mapping. This is more complex and less desirable.

Option 1 is recommended — it aligns with how `toGeneric()` at line 102 already falls through to returning the input unchanged for unmapped names, but making it explicit in the map documents intent and makes the sync test a true guard.

### Non-blocking observations

- Commit message format is correct: `fix(ai): ...` with proper scope.
- Single-file, single-line change — appropriately scoped.
- No attribution violations.

PR #708, opened 2026-04-17
fix(clinical): preserve validationWarnings when skipValidation is true

### LINKED ISSUE #704: fix(clinical): preserve validationWarnings when skipValidation is true in createLabPanel
From review of #695.

When `createLabPanel` is called with `{ skipValidation: true }`, the `allWarnings` array stays empty and the clinical event never includes `validationWarnings`. The MedLens bridge validates each result (medlens-bridge.ts L242) but only checks `validation.valid` and `validation.errors` — it discards warnings.

This means valid-but-warned results (e.g., Glucose 350 mg/dL — valid unit but clinically alarming value) lose their warning metadata on the emitted clinical event.

Options:
1. Have the MedLens bridge collect warnings and pass them through to `createLabPanel` (add a `warnings` field to options).
2. Always run the warning-collection pass in `createLabPanel` even when `skipValidation` is true (only skip the error/reject path).
3. Accept the gap and document that `skipValidation` also suppresses warnings.

Option 2 is the simplest and preserves the existing contract.

blamechris:
## Summary
- Always run the `validateLabResult` warning-collection pass in `createLabPanel`, even when `skipValidation: true`
- Only skip the error-collection/reject path when `skipValidation` is set
- Fixes #704: `allWarnings` array was empty when callers (e.g. MedLens bridge) passed `skipValidation: true`, causing `validationWarnings` to be missing from emitted clinical events

## Test plan
- [ ] Call `createLabPanel` with `skipValidation: true` and results that trigger warnings; verify `validationWarnings` is present in the emitted `lab.resulted` event
- [ ] Call `createLabPanel` without `skipValidation` and results with errors; verify the panel is still rejected
- [ ] Call `createLabPanel` without `skipValidation` and results with warnings only; verify warnings are collected and errors array is empty

### DIFF
--- a/services/clinical-data/src/repositories/lab-repo.ts
+++ b/services/clinical-data/src/repositories/lab-repo.ts
@@ -37,11 +37,11 @@ export async function createLabPanel(
   const allWarnings: string[] = [];
   const allErrors: string[] = [];
 
-  if (!options?.skipValidation) {
-    for (const r of input.results) {
-      const v = validateLabResult(r.test_name, r.value, r.unit);
+  for (const r of input.results) {
+    const v = validateLabResult(r.test_name, r.value, r.unit);
+    allWarnings.push(...v.warnings);
+    if (!options?.skipValidation) {
       allErrors.push(...v.errors);
-      allWarnings.push(...v.warnings);
     }
   }
 


blamechris (commented):
## Agent Review — PR #708

**Verdict: Approve (self-owned PR; posted as comment)**

### What changed
Single-file, 4-line change in `services/clinical-data/src/repositories/lab-repo.ts`. The `validateLabResult` loop is hoisted out of the `if (!options?.skipValidation)` guard so warnings are always collected, while `allErrors.push` remains gated behind `skipValidation`. This ensures the `validationWarnings` array on emitted `lab.resulted` clinical events is populated even when callers (e.g. the MedLens bridge) pass `skipValidation: true`.

### Standards checklist
| Check | Status |
|---|---|
| Commit format `type(scope): ...` | Pass |
| TypeScript strict / ESM | Pass (no new imports) |
| Functional style | Pass |
| Zero Attribution Policy | Pass |
| CI (Build / Test / Typecheck) | All green |
| Fixes referenced issue | Yes — closes #704 |

### Correctness
The restructured loop is logically equivalent to the original for the `skipValidation: false` path (both errors and warnings collected). For the `skipValidation: true` path, only warnings are collected — which is the intended fix. The error-throw guard on L48 still fires only when errors are present, so `skipValidation: true` callers cannot trigger a rejection.

### Deferred item
The block comment at L33-36 ("Callers that have already validated can pass `{ skipValidation: true }` to avoid redundant work") is now slightly misleading — `skipValidation` no longer skips all validation, only the error/reject path. A one-line addendum would prevent future confusion. Filed as a from-review issue.

PR #709, opened 2026-04-17
test(gateway): apply selectColumns projection in mock .limit() handler

### LINKED ISSUE #700: test(gateway): apply selectColumns projection in mock .limit() handler
## Context
From review of #682. The mock Drizzle select chain in `patient-records-projection.test.ts` now correctly applies `selectColumns` filtering in both `.where` and `.then` (on `.from`), but the `.limit` handler (line 68) still returns raw `resolvedData` without projection.

## Problem
If any production code path uses `db.select(cols).from(table).limit(n)` (without `.where`), the mock will return all columns instead of only the projected ones, masking potential PHI-leaking regressions in tests.

## Suggested fix
Extract a shared `applyProjection(rows, selectColumns)` helper inside `vi.hoisted` and apply it in `.where`, `.then`, and `.limit` handlers.

## File
`services/api-gateway/src/__tests__/patient-records-projection.test.ts` (line 68)

blamechris:
## Summary
- Extract shared `applyProjection` helper in patient-records-projection test mock to eliminate duplicated projection logic
- Apply projection in `.limit()` handler which previously returned raw data without filtering by `selectColumns`
- Consolidate identical projection logic from `.where()` and `.then()` handlers to use the shared helper

Closes #700

## Test plan
- [ ] `pnpm typecheck` passes
- [ ] `pnpm lint` passes
- [ ] Existing patient-records-projection tests continue to pass with corrected `.limit()` mock behavior

### DIFF
--- a/services/api-gateway/src/__tests__/patient-records-projection.test.ts
+++ b/services/api-gateway/src/__tests__/patient-records-projection.test.ts
@@ -43,29 +43,31 @@ const mocks = vi.hoisted(() => {
   let selectColumns: Record<string, unknown> | undefined = undefined;
   let resolvedData: unknown[] = [];
 
+  /**
+   * Shared projection helper — if `selectColumns` is set, strip each row
+   * down to only those keys so the mock faithfully simulates Drizzle
+   * column selection. Used by `.where()`, `.limit()`, and `.then()`.
+   */
+  function applyProjection(rows: unknown[]): unknown[] {
+    if (!selectColumns) return rows;
+    const keys = Object.keys(selectColumns);
+    return rows.map((row) => {
+      const out: Record<string, unknown> = {};
+      for (const k of keys) {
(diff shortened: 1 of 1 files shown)


blamechris (commented):
Clean DRY refactor. The shared `applyProjection` helper correctly consolidates three copies of the same projection logic into one, and the `.limit()` handler now faithfully applies `selectColumns` filtering.

**Verified:**
- `applyProjection` correctly short-circuits when `selectColumns` is undefined
- All three call-sites (`.where()`, `.limit()`, `.then()`) now use the shared helper
- No attribution violations, no secrets, commit format matches project conventions
- Net reduction of 11 lines with improved maintainability

**Deferred item (issue filed):**
- No test currently exercises a code path that reaches `.limit()` without `.where()`, so the fix is structurally correct but not yet regression-tested end-to-end. Filed as a from-review issue.

PR #712, opened 2026-04-17
test(logger): guard against metadata key collision with base fields

### LINKED ISSUE #703: test(logger): add metadata key collision test and guard against base field override
## Context
From review of PR #691 (logger unit tests).

In `packages/logger/src/index.ts`, the `emit()` function spreads metadata after base fields:

```ts
const entry: LogEntry = {
  timestamp: new Date().toISOString(),
  level,
  service,
  msg,
  ...meta,
};
```

If a caller passes `{ level: "debug", service: "evil" }` as metadata, the spread silently overrides the base `level` and `service` fields. This could corrupt log data and make filtering unreliable.

## Proposed
1. Add a test case that asserts base fields (`timestamp`, `level`, `service`, `msg`) cannot be overridden by metadata
2. Guard against this in production code -- either strip reserved keys from metadata before spreading, or spread metadata first so base fields win

## Files
- `packages/logger/src/__tests__/index.test.ts`
- `packages/logger/src/index.ts`

blamechris:
## Summary
- Fixes #703: In `createLogger`, metadata spread (`...meta`) was applied after base fields, allowing callers to accidentally (or maliciously) override `level`, `service`, `timestamp`, or `msg`.
- Swapped spread order so `...meta` comes first and base fields overwrite any collisions.
- Added test suite in `packages/logger/__tests__/index.test.ts` with dedicated collision-protection assertions for all four base fields.

## Test plan
- [x] `pnpm --filter @carebridge/logger test` passes (18 tests)
- [x] `pnpm typecheck` passes across monorepo

### DIFF
--- a/packages/logger/__tests__/index.test.ts
+++ b/packages/logger/__tests__/index.test.ts
@@ -1,0 +1,91 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import { createLogger } from "../src/index.js";
+
+describe("createLogger", () => {
+  let logSpy: ReturnType<typeof vi.spyOn>;
+  let errorSpy: ReturnType<typeof vi.spyOn>;
+
+  beforeEach(() => {
+    logSpy = vi.spyOn(console, "log").mockImplementation(() => {});
+    errorSpy = vi.spyOn(console, "error").mockImplementation(() => {});
+  });
+
+  afterEach(() => {
+    logSpy.mockRestore();
+    errorSpy.mockRestore();
+  });
+
+  function parseLastLog(): Record<string, unknown> {
+    const call = logSpy.mock.calls.at(-1) ?? errorSpy.mock.calls.at(-1);
+    return JSON.parse(call![0] as string) as Record<string, unknown>;
+  }
+
+  function parseLastError(): Record<string, unknown> {
+    const call = errorSpy.mock.calls.at(-1);
+    return JSON.parse(call![0] as string) as Record<string, unknown>;
+  }
+
+  it("emits structured JSON with base fields", () => {
+    const logger = createLogger("test-service");
+    logger.info("hello");
+
+    const entry = parseLastLog();
+    expect(entry.level).toBe("info");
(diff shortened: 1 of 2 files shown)


blamechris (commented):
## CareBridge Inspector Review -- PR #712

### Verdict: Request Changes (minor, non-blocking path available)

### What the PR does
Fixes issue #703 by swapping the spread order in `createLogger` so `...meta` comes before the base fields (`timestamp`, `level`, `service`, `msg`), preventing callers from overriding reserved log fields. Adds a test suite asserting collision protection for all four base fields.

### Production code change
The one-line fix in `packages/logger/src/index.ts` is correct and minimal. Moving `...meta` above the base fields is the right approach -- clean, no runtime cost, no API change.

### Issues found

**1. Test file placed outside `src/` -- typecheck gap (must fix)**
The new tests live at `packages/logger/__tests__/index.test.ts`, but the package `tsconfig.json` sets `include: ["src"]`. This means `pnpm --filter @carebridge/logger typecheck` does not cover these tests. Every other package in the monorepo keeps tests under `src/__tests__/`. The collision-protection tests should be added to the existing file at `packages/logger/src/__tests__/index.test.ts` instead.

**2. Duplicate coverage with existing test suite**
`packages/logger/src/__tests__/index.test.ts` already has 14 tests covering structured JSON output, level routing, and metadata merging -- three of the four top-level `describe` groups in the new file are redundant. Only the `"base field collision protection"` describe block is net-new. Merging those four collision tests into the existing file avoids duplication and keeps a single source of truth.

**3. Unused `parseLastError` helper**
`parseLastError()` (lines 23-26) is defined but never called. Should be removed or used.

### Suggested resolution
Move the four collision-protection test cases into the existing `packages/logger/src/__tests__/index.test.ts` under a new `describe("base field collision protection", ...)` block, and delete `packages/logger/__tests__/index.test.ts` entirely.

### What looks good
- The spread-order fix is the simplest correct solution
- Collision tests cover all four reserved fields
- CI is green (build, test, typecheck all pass)
- PR body is well-structured with a clear test plan

PR #719, opened 2026-04-17
chore(ai): add timestamp_fallback_total metric for ops alerting

### LINKED ISSUE #586: chore(ai): add timestamp_fallback_total metric for ops alerting
From review of PR #568.

Add a metric counter (e.g. `timestamp_fallback_total{reason=...}`) in `validateEventTimestamp` so ops can alert on a spike of bad timestamps indicating an upstream publisher regression.

blamechris:
## Summary
- Replaces raw `console.warn` in `validateEventTimestamp` with structured logging via `@carebridge/logger`
- Emits a `timestamp_fallback_total` warn-level log line with `metric`, `caller`, `eventId`, `reason`, `value`, and `fallbackTimestamp` fields so log aggregators can count fallback occurrences and ops can set threshold alerts
- Adds `@carebridge/logger` as a dependency of `@carebridge/ai-oversight`

Closes #586

## Test plan
- [ ] Verify `pnpm typecheck` and `pnpm lint` pass (confirmed locally)
- [ ] Confirm structured JSON log line appears in stderr when a fallback path is triggered (e.g., missing or malformed timestamp in a clinical event)
- [ ] Validate that existing `validate-event-timestamp` unit tests still pass

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -289,6 +289,9 @@ importers:
       '@carebridge/db-schema':
         specifier: workspace:*
         version: link:../../packages/db-schema
+      '@carebridge/logger':
+        specifier: workspace:*
+        version: link:../../packages/logger
       '@carebridge/medical-logic':
         specifier: workspace:*
         version: link:../../packages/medical-logic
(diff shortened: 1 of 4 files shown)


blamechris (commented):
## Review — PR #719

### Blocking: CI test failure (8 of 10 tests red)

The implementation correctly replaces `console.warn` with structured `logger.warn(...)` from `@carebridge/logger`, but the existing tests in `validate-event-timestamp.test.ts` were not updated to match.

The tests spy on `console.warn`:
```ts
warnSpy = vi.spyOn(console, "warn").mockImplementation(() => {});
```

However, `@carebridge/logger` routes `warn`-level output through `console.error` (see `packages/logger/src/index.ts` line 49). Every assertion like `expect(warnSpy).toHaveBeenCalledTimes(1)` now fails because `console.warn` is never called — the output goes to `console.error` instead.

**Fix options (pick one):**
1. **Mock the logger module** via `vi.mock("@carebridge/logger")` and assert against the mock's `.warn()` — this is the cleanest approach since it decouples the test from the logger's internal routing.
2. **Spy on `console.error`** instead of `console.warn` — quick but brittle (couples tests to logger internals).

Additionally, the "uses the caller label in the log prefix" test (line 107) asserts `toContain("[context-builder]")`, matching the old `console.warn` format string `[${caller}]`. The structured logger emits JSON with a `caller` field instead, so this assertion needs to change to check the JSON structure or the `caller` metadata field.

### Nit: redundant `metric` field

The call passes `"timestamp_fallback_total"` as both the `msg` argument and the `metric` metadata key:
```ts
logger.warn("timestamp_fallback_total", {
  metric: "timestamp_fallback_total",
  ...
});
```
The logger already writes `msg` into the JSON output, so the `metric` key duplicates it. I'd keep `metric` only if your log aggregator queries by a dedicated `metric` field rather than `msg` — otherwise drop it to reduce noise.

### What looks good

- Correct scope: replacing a raw `console.warn` with structured logging is the right call for ops alerting on fallback spikes.
- The `@carebridge/logger` dependency addition in `package.json` and lockfile looks clean.
- Metadata fields (`caller`, `eventId`, `reason`, `value`, `fallbackTimestamp`) give operators everything needed to correlate and alert.
- No PHI exposure — event ID is safe to log per the existing doc comment.

Issue #533, opened 2026-04-16
Migrate context-builder and labs UI to calculateDeltaFromBaseline


blamechris:
Identified during review of PR #499.

PR #499 adds `calculateDeltaFromBaseline` / `getTrendInfoFromBaseline` to `packages/medical-logic/src/trend-utils.ts` but no call sites are migrated, so the clinical value (catching slow baseline drift) is dormant.

## Scope

**1. `services/ai-oversight/src/workers/context-builder.ts:143`** — currently calls `calculateDelta(values)` on per-type vitals series, which under-reports slow drift to the LLM. Switch to `calculateDeltaFromBaseline` using either:
- The first value of the retrieval window as the baseline (preserves current behavior when only one window is loaded), or
- A per-patient baseline pulled from prior-admission data if available.

This is the highest-leverage call site because it directly shapes what the LLM reviewer sees.

**2. `apps/patient-portal/app/labs/page.tsx`** and **`apps/patient-portal/app/health-summary/page.tsx`** — neither currently consumes `getTrendInfo`, but if trend arrows/colors are added they should use the baseline-aware variant so creatinine/BUN drift surfaces correctly.

## Acceptance

- [ ] `context-builder` uses `calculateDeltaFromBaseline`
- [ ] A test asserts the 0.7 → 0.8 → 0.9 → 1.1 → 1.2 creatinine series produces a `rising` trend in the context payload
- [ ] Any new patient-portal trend rendering uses `getTrendInfoFromBaseline`

## References

- PR #499
- Closes-relationship issue #249

PR #720, opened 2026-04-17
docs(clinical): clarify zero-baseline and single-value semantics in calculateDeltaFromBaseline

### LINKED ISSUE #534: Clarify zero-baseline and single-value semantics in calculateDeltaFromBaseline
Identified during review of PR #499.

`packages/medical-logic/src/trend-utils.ts` — two subtle semantics in `calculateDeltaFromBaseline` need documentation and/or type tightening.

## 1. Zero baseline returns `pctChange: 0`

`calculateDeltaFromBaseline([5], 0)` returns `{ current: 5, previous: 0, change: 5, pctChange: 0 }`. The `pctChange: 0` is explicitly tested as "falls back to 0 rather than Infinity", but for a metric where 0 is a real measurement (pain level, output volumes, some tumor markers post-treatment) this reads as "no change" when the absolute change is +5. Anything thresholding on `pctChange` will under-fire.

**Options:**
- Change `Delta.pctChange` to `number | null` and return `null` when `previous === 0`. Callers must then consult `change` explicitly.
- Keep 0 but add a JSDoc warning + update `formatDelta` to say "+5 (no baseline)" when `previous === 0 && change !== 0`.

## 2. Asymmetric single-value behavior is undocumented

`calculateDeltaFromBaseline([1.2])` returns `null`, but `calculateDeltaFromBaseline([1.2], 0.7)` returns a valid delta. This diverges from `calculateDelta` (always requires length ≥ 2). The behavior is intentional and tested, but the JSDoc never says so.

**Fix:** add a line to the JSDoc: "Returns a delta even for a single-value series when an explicit baseline is supplied; returns `null` only when the series is empty or when length < 2 and no baseline is given."

## 3. Guard ordering is subtle

```ts
const current = values[values.length - 1]!;
const base = baseline ?? values[0]!;
if (values.length < 2 && baseline === undefined) return null;
```

Correct today (the early `values.length === 0` return protects `values[0]!`), but relies on two guards composing correctly. Hoisting the length check above the `base` assignment would make the invariants explicit.

## 4. Cross-reference `classifyTrend` in JSDoc

`classifyTrend` is already baseline-aware (uses `values[0]` → `values[values.length - 1]`). Add "See also `classifyTrend` for direction-only classification over the same baseline semantics" to `calculateDeltaFromBaseline` and `getTrendInfoFromBaseline` so future contributors don't write a third overlapping helper.

## References

- PR #499

blamechris:
## Summary

- Expanded JSDoc for `calculateDeltaFromBaseline` to document three previously undocumented edge-case behaviors: empty array, single-value-with-explicit-baseline (valid), and zero-baseline (`pctChange` falls back to `0`).
- Hoisted the `values.length < 2 && baseline === undefined` guard above the `base` assignment so the early-return logic reads top-down without a "look-ahead" surprise.
- Added `@see` cross-references to `classifyTrend` and `calculateDelta`.

Closes #534

## Test plan

- [x] `pnpm typecheck` passes (no behavior changes, JSDoc + reorder only)
- [x] `pnpm lint` passes
- [ ] Verify existing `calculateDeltaFromBaseline` tests still pass (no logic change)

### DIFF
--- a/packages/medical-logic/src/trend-utils.ts
+++ b/packages/medical-logic/src/trend-utils.ts
@@ -101,15 +101,33 @@ export function calculateDelta(values: number[]): Delta | null {
  * When `baseline` is omitted, the first value in `values` is used as the
  * baseline so callers that only know the series still get the correct
  * "change-from-baseline" semantics.
+ *
+ * ### Edge-case semantics
+ *
+ * - **Empty array**: returns `null` regardless of `baseline`.
+ * - **Single value, no explicit baseline**: returns `null` — a lone reading
+ *   has no implicit baseline to compare against (the value *is* the baseline).
+ * - **Single value with explicit `baseline`**: returns a valid `Delta`. This
+ *   is intentional: a prior-admission creatinine of 0.7 with a single new
+ *   reading of 1.4 is a meaningful 100% rise and should surface in the UI.
+ * - **Zero baseline** (`base === 0`): `pctChange` falls back to `0` instead
+ *   of `null` or `Infinity` so downstream formatters and comparisons never
+ *   encounter non-finite numbers. The absolute `change` is still accurate.
+ *
+ * @see {@link classifyTrend} for directional classification (rising/falling/stable)
+ *      independent of good-direction coloring.
+ * @see {@link calculateDelta} for the simpler last-two-point variant.
  */
 export function calculateDeltaFromBaseline(
   values: number[],
   baseline?: number,
 ): Delta | null {
   if (values.length === 0) return null;
+  // A single value with no explicit baseline is meaningless — the value *is*
+  // its own baseline, so there is no delta to report.
+  if (values.length < 2 && baseline === undefined) return null;
   const current = values[values.length - 1]!;
   const base = baseline ?? values[0]!;
-  if (values.length < 2 && baseline === undefined) return null;
   const change = current - base;
   const pctChange = base !== 0 ? (change / base) * 100 : 0;
   return { current, previous: base, change, pctChange };


blamechris (commented):
## Inspector Review -- PR #720

**Verdict: Approve** -- clean docs + reorder PR that fully addresses all four items in #534.

### Verified

- **JSDoc edge-case docs**: Empty array, single-value-no-baseline, single-value-with-baseline, and zero-baseline semantics are now explicitly documented with clinical rationale (creatinine example).
- **Guard reorder**: The `values.length < 2 && baseline === undefined` guard is hoisted above the `base` assignment. Behavior is identical -- the `base` line was unreachable in the early-return path -- but the top-down read is cleaner and eliminates the look-ahead surprise noted in #534 point 3.
- **`@see` cross-references**: `classifyTrend` and `calculateDelta` are linked, satisfying #534 point 4.
- **Zero attribution**: No violations.
- **Commit format**: `docs(clinical): ...` matches project conventions.

### Observations (non-blocking)

1. **Issue #534 point 1 (zero-baseline `pctChange: 0`) is only JSDoc-mitigated, not type-hardened.** The issue offered two options; this PR chose the lighter JSDoc path. That is fine for now, but any future threshold-based consumer of `pctChange` should guard against `previous === 0` explicitly. Issue #533 (migrate call sites to `calculateDeltaFromBaseline`) is the natural place to address this when those call sites are wired up.

2. **Test plan checkbox is unchecked.** The PR body says "Verify existing `calculateDeltaFromBaseline` tests still pass (no logic change)" but it is unchecked. Since the guard reorder is provably behavior-preserving and CI presumably ran the test suite, this is cosmetic -- but worth ticking before merge.

PR #738, opened 2026-04-17
fix(ai): update validate-event-timestamp tests for structured logger

### LINKED ISSUE #737: fix(ai): update validate-event-timestamp tests for structured logger
From review of PR #719.

PR #719 replaces `console.warn` with `@carebridge/logger` in `validateEventTimestamp`, but the test file (`services/ai-oversight/src/__tests__/validate-event-timestamp.test.ts`) still spies on `console.warn`. Since the logger routes warn-level output through `console.error`, all 8 fallback-path test assertions fail in CI.

**Required changes:**
1. Mock the `@carebridge/logger` module (`vi.mock`) instead of spying on `console.warn`
2. Assert against the mock logger's `.warn()` calls and their structured metadata arguments
3. Update the "caller label in log prefix" test — it currently checks for `[context-builder]` in a format string, but the structured logger emits a JSON `caller` field instead

blamechris:
## Summary
- Replace `console.error` spy with `vi.mock` of `@carebridge/logger` so tests assert against the mock logger's `.warn()` calls and structured metadata arguments directly
- Use `vi.hoisted()` to avoid the reference-before-initialization issue with vitest's module mock hoisting
- Update the "caller label" test to assert against the metadata `caller` field instead of parsing a JSON string from console output

## Test plan
- [x] All 11 tests in `validate-event-timestamp.test.ts` pass
- [x] No new typecheck errors introduced in the changed file

Closes #737

### DIFF
--- a/services/ai-oversight/src/__tests__/validate-event-timestamp.test.ts
+++ b/services/ai-oversight/src/__tests__/validate-event-timestamp.test.ts
@@ -1,6 +1,17 @@
-import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import { describe, it, expect, vi, beforeEach } from "vitest";
 
-import type { LogEntry } from "@carebridge/logger";
+// Mock the @carebridge/logger module so we can assert against logger.warn()
+// calls directly rather than spying on console.error (which is an
+// implementation detail of the logger).
+const { mockWarn } = vi.hoisted(() => ({ mockWarn: vi.fn() }));
+vi.mock("@carebridge/logger", () => ({
+  createLogger: () => ({
+    debug: vi.fn(),
+    info: vi.fn(),
+    warn: mockWarn,
+    error: vi.fn(),
+  }),
+}));
 
 import { validateEventTimestamp } from "../utils/validate-event-timestamp.js";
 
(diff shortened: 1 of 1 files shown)


blamechris:
## Code Review: PR #738

### Summary
Migrates `validate-event-timestamp.test.ts` from spying on `console.error` (an implementation detail of the logger) to mocking the `@carebridge/logger` module directly. This aligns the test assertions with the public API of the logger — `logger.warn(msg, meta)` — rather than the internal transport mechanism (stderr JSON lines). Closes #737.

### Strengths
- **Correct use of `vi.hoisted()`** to avoid the temporal dead-zone issue with vitest's module mock hoisting. This is the idiomatic vitest pattern.
- **Mock shape matches the real `Logger` interface exactly** — all four methods (`debug`, `info`, `warn`, `error`) are present, matching `packages/logger/src/index.ts:21-26`.
- **Assertions are now testing the right contract.** The old tests parsed a serialised JSON string from console output and checked `entry.level === "warn"`, which is redundant when you are already spying on `console.error`. The new tests assert against `mockWarn` calls directly, which is both simpler and more meaningful.
- **Clean removal of `afterEach` + `mockRestore`** — since the mock is module-scoped, `mockClear()` in `beforeEach` is the correct cleanup strategy.
- **Net zero line change** (47 added, 47 removed) — pure refactor with no behavioural change. CI passes clean (build, test, typecheck).

### Issues Found

#### Critical (Must Fix)
None.

None.

#### Nitpicks (Optional)
1. **Unused `LogEntry` type removal is good, but the mock could export it for other consumers.** Not an issue today since `validate-event-timestamp.test.ts` is the only consumer, but if future tests in this file import `LogEntry`, the mock would need updating. Fine to leave as-is — cross that bridge when needed.

None — this is a clean, focused test refactor.

### Verdict
**Approve.** Clean, correct migration from implementation-detail spying to proper module mocking. No issues found. Ship it.

PR #752, opened 2026-04-17
fix(fhir): enforce recommended_purge_at with security headers

### LINKED ISSUE #557: Enforce recommended_purge_at for FHIR bundle exports (delivery hardening)
Identified during review of PR #503.

PR #503 adds `recommended_purge_at` (generation + 15 min) to the `exportPatient` response but provides no enforcement. The docstring correctly notes this is advisory and points at #290 for the real fix, but the follow-up deserves its own tracking issue so it doesn't get lost inside #290's broader scope.

## Gaps today
- No `Cache-Control: no-store` on the response.
- No `Content-Disposition: attachment` (browsers / proxies may cache JSON responses).
- tRPC responses flow through any HTTP intermediary that doesn't know about the advisory field.
- No signed short-TTL URL delivery layer — the bundle is inline in the tRPC response.
- No server-side record of delivered exports to drive a purge job.

## Proposed fix (scoped, not dependent on #290)
Short-term, without adding a new HTTP surface:
1. Set `Cache-Control: no-store, no-cache, must-revalidate` on `exportPatient` responses at the Fastify layer.
2. Set `Content-Disposition: attachment; filename="bundle-<export_id>.json"`.
3. Log a warning in `ai-oversight` or an audit dashboard when the client re-requests the same export past `recommended_purge_at`.

Long-term (#290):
- Signed short-TTL URL delivery layer; inline tRPC return deprecated.

## Files
- `services/fhir-gateway/src/router.ts`
- possibly `services/api-gateway/src/index.ts` for header middleware

## Acceptance
- [ ] Response carries no-cache headers
- [ ] Content-Disposition forces download semantics
- [ ] Test asserts headers on the exportPatient path

blamechris:
## Summary

- Adds `Cache-Control: no-store, no-cache, must-revalidate` to `exportPatient` responses so intermediaries and browsers never cache PHI bundles
- Adds `Content-Disposition: attachment; filename="bundle-<export_id>.json"` so clients receive the bundle as a file download
- Introduces an optional `setHeader` callback on the fhir-gateway `Context`, threaded from the api-gateway Fastify reply, so the pure tRPC router can influence HTTP transport headers without a direct Fastify dependency

## Test plan

- [x] 4 new tests in `services/fhir-gateway/src/__tests__/export-headers.test.ts`:
  - Cache-Control header set on success
  - Content-Disposition header with correct export-id filename on success
  - Graceful no-op when `setHeader` absent (direct `createCaller` usage)
  - Headers not set on failure path (patient not found)
- [x] All 76 existing fhir-gateway tests still pass
- [x] `tsc` build succeeds for both fhir-gateway and api-gateway

Closes #557


blamechris:
## Review: fix(fhir): enforce recommended_purge_at with security headers

**Verdict: Approve** -- clean, well-scoped change with good test coverage. CI green (Build, Test, Typecheck all pass).

---

### Findings

**0 blocking issues.**

#### Positive observations

1. **Transport-layer decoupling is well designed.** The optional `setHeader` callback on Context avoids coupling fhir-gateway to Fastify directly. The `if (ctx.setHeader)` guard means direct `createCaller` usage (tests, internal callers, non-HTTP contexts) degrades gracefully with zero risk of runtime errors.

2. **Header selection is correct.** `Cache-Control: no-store, no-cache, must-revalidate` covers the full spectrum of intermediary caching behaviors (HTTP/1.0 proxies, CDN edge caches, browser disk caches). `Content-Disposition: attachment` prevents inline rendering of PHI bundles in browsers.

3. **Ordering is right.** Headers are set *after* `writeAudit({ success: true })`, which means a failure in the audit write correctly prevents the headers from being set and the response from succeeding. The failure path correctly skips header-setting.

4. **Tests are thorough.** Four tests cover the key scenarios: both headers on success, graceful no-op without `setHeader`, and no headers on failure. The Content-Disposition test extracts `exportId` from the bundle meta extension rather than hardcoding, which makes it resilient to implementation changes.

5. **`getByPatient` correctly does NOT receive `setHeader`.** Individual resource queries are not PHI export bundles and do not need download/no-cache semantics.

#### Non-blocking observations (informational)

1. **Pragma: no-cache not included.** Some legacy HTTP/1.0 caches only respect `Pragma: no-cache`, not `Cache-Control`. In practice this is unlikely to matter for a JSON API behind TLS, but if the bundle ever traverses an enterprise proxy chain, adding `Pragma: no-cache` is a one-liner hardening step. Not worth blocking on.

2. **Re-request audit warning (issue #557 item 3) not addressed.** The issue body proposed logging a warning when a client re-requests the same export past `recommended_purge_at`. This PR does not claim to address that and it is clearly out of scope -- filing a follow-up issue for tracking.

---

No changes requested. This is ready to merge.

Issue #543, opened 2026-04-16
portal(labs): add component tests for LabsTab reference-range and inferred-flag logic


blamechris:
**Identified during review of PR #505.**

### Context
PR #505 added branching render logic to `LabsTab`:
- Reference-range formatting (both bounds, low-only, high-only, neither).
- Inferred-flag derivation (`low` vs. `high` from value-vs-bounds).
- Value color precedence (`critical` > server `low`/`high` > inferred out-of-range > default).

### Problem
None of this is covered by a test. The original regression (reading non-existent `reference_range`) shipped because the render logic had no component-level assertions. The PR body's manual-test checkbox is also unchecked.

### Proposal
Add component tests (Vitest + Testing Library or Playwright component test) covering at minimum:
- K+ 3.2 with ref 3.5–5.0, no server flag → value text uses warning color, `LOW` badge with `role="status"` and "Inferred" in `aria-label`.
- K+ 5.5 with ref 3.5–5.0, no server flag → `HIGH` inferred badge.
- K+ 2.5 with server `flag="critical"` → critical color, `CRITICAL` badge, no inferred badge.
- K+ 4.0 in range → default color, no badge.
- Reference range formatting: `{low, high}`, `{low only}`, `{high only}`, `{neither}` → renders `3.5–5.0`, `> 3.5`, `< 5.0`, `—`.

### Acceptance
- Tests live under the clinician-portal test suite and run in CI.
- Tests fail if `reference_range` is read again or if inferred-flag logic breaks.

`apps/clinician-portal/app/patients/[id]/page.tsx`

PR #755, opened 2026-04-17
fix(portal): document and enforce flag precedence in LabsTab

### LINKED ISSUE #544: portal(labs): document and enforce precedence between server flag and client-inferred flag
**Identified during review of PR #505.**

### Context
PR #505 introduces a client-inferred H/L badge that only renders when `!flag && isOutOfRange`. This correctly lets any truthy server-provided `flag` take precedence.

### Problem
If the server (or AI oversight pipeline / instrument-config work) later starts emitting `flag="low"` for borderline values that the client would also infer, a future contributor might "simplify" by removing the inference block — reintroducing the gap this PR fixes for instruments that still don't flag borderlines. There is no comment or invariant documenting the contract.

### Proposal
1. Add a short block comment above the `inferredFlag` derivation stating the contract: "Server flag is authoritative when present; client inference is a safety net for instruments/legacy results that do not flag borderline out-of-range."
2. Consider a guard/assertion in dev builds: if both a server flag and an inferred flag are computed and they disagree (e.g. server says `high`, value is below `reference_low`), log a warning — this would surface data-quality issues.

### Acceptance
- Inline comment in `LabsTab` explaining precedence.
- Optional dev-only console warning for server/inferred disagreement.

`apps/clinician-portal/app/patients/[id]/page.tsx`

blamechris:
## Summary

Closes #544.

- Add a block comment above the `inferredFlag` derivation in `LabsTab` documenting the precedence contract: server flag is authoritative when present; client inference is a safety net for instruments/legacy results that omit borderline out-of-range flags.
- Add a dev-only `console.warn` when the server flag direction (H/L) disagrees with what the client would infer from `reference_low`/`reference_high`. This catches mapping bugs and stale reference ranges early without polluting production logs.
- Add 14 unit tests covering both the inferred-flag derivation and the disagreement detection logic.

## Test plan

- [x] `vitest run` passes all 14 new tests in `lab-flag-precedence.test.ts`
- [ ] Manual: open a patient with labs in dev mode, verify no spurious warnings in console for correctly-flagged results
- [ ] Manual: temporarily mock a disagreement (e.g., server flag "H" on a value below `reference_low`) and confirm the console warning fires


blamechris:
## Review: fix(portal): document and enforce flag precedence in LabsTab

**Verdict: Approve with minor suggestions**

---

### What this PR does well

1. **Clear precedence contract.** The block comment above `inferredFlag` is well-written and explains *why* the server flag wins — not just *that* it wins. Future contributors will hesitate before "simplifying" this away.

2. **Dev-only disagreement warning.** Gating on `process.env.NODE_ENV !== "production"` is the right call — Next.js dead-code-eliminates this branch in production builds, so zero runtime cost in prod.

3. **Thorough test coverage.** 14 tests covering both the inferred-flag derivation and the disagreement detection, including edge cases (null value, in-range value, non-directional "critical" flag).

---

### Findings

#### 1. Duplicated logic between component and test (low severity, informational)

The test file replicates the inline derivation from `LabsTab` in two standalone functions (`deriveInferredFlag`, `checkDisagreement`). The file itself acknowledges this with the comment "If the derivation is ever extracted to a shared helper, these tests should be pointed at the real export instead." This is a pragmatic choice given the cost of mounting a full React tree with tRPC context, but it does mean the tests can pass while the component drifts. Issue #543 already tracks adding component-level tests that would close this gap — no new issue needed.

#### 2. `valueColor` only checks `"H"` and `"L"`, but disagreement check also handles `"high"` and `"low"` (low severity)

The existing `valueColor` ternary (line ~500-506) only matches `flag === "H" || flag === "L"` for the warning color path. The new disagreement check correctly normalizes `flag.toLowerCase()` and matches against both `"h"/"l"` and `"high"/"low"`. This means if a server ever sends `flag="high"` or `flag="low"`, the disagreement check would catch it but `valueColor` would fall through to the `isOutOfRange` branch (which happens to produce the same warning color, so the visual result is fine). Not a bug in this PR, but worth noting as a pre-existing inconsistency. Would be a good cleanup for a follow-up.

#### 3. Boundary values at exactly `refLow` or `refHigh` (informational)

`isOutOfRange` uses strict `<` and `>`, meaning a value exactly equal to a boundary is considered in-range. This is the pre-existing convention and is preserved correctly here. The test suite tests values clearly inside and outside the range but not at the boundary. A boundary-value test (e.g., `value === refLow`, `value === refHigh`) would strengthen confidence, but this is minor.

#### 4. Line length in the `if` guard (style, nitpick)

The `if (process.env.NODE_ENV !== "production" && flag && isOutOfRange && typeof value === "number")` line is 96 characters. The project doesn't seem to enforce a strict line length, so this is fine, but extracting the condition into a named boolean (e.g., `const shouldCheckDisagreement = ...`) would improve readability slightly.

---

### Summary

| Category | Count |
|----------|-------|
| Blocking issues | 0 |
| Suggestions | 2 (boundary tests, named boolean) |
| Informational | 2 (duplicated logic, valueColor inconsistency) |

PR is clean, well-documented, and correctly addresses #544. Approve.

PR #757, opened 2026-04-17
fix(gateway): per-user rate limit on patients.getSummary and getById

### LINKED ISSUE #552: gateway: per-user rate limit on patients.getSummary and getById

## Context
PR #502 introduced `patients.getSummary(id)` returning `{ id, name, mrn }`. It is gated by `enforcePatientAccess`, so it is not an unauthenticated enumeration oracle, but a logged-in admin or a compromised clinician session can still hit it in a tight loop to enumerate names and MRNs against guessed UUIDs. The same concern applies to `getById`.

The repo already has an IP-bucket rate-limit primitive used on `/accept-invite` (see `services/api-gateway/src/__tests__/accept-invite-rate-limit.test.ts`).

## Proposal
1. Add a per-user (userId-keyed) rate limit to `patients.getSummary` and `patients.getById`.
2. Conservative defaults, e.g. 60 req/min per user for `getSummary`, 120 req/min per user for `getById`.
3. Emit an audit event on limit exceedance (suspicious enumeration pattern).

## Acceptance
- Both procedures reject requests over the configured threshold with `TOO_MANY_REQUESTS`.
- Test covers: normal caller passes; burst exceeds → `TOO_MANY_REQUESTS`; audit row recorded on exceedance.

blamechris:
## Summary
- Adds per-user (userId-keyed) rate limiting to `patients.getSummary` (60 req/min) and `patients.getById` (120 req/min) to prevent patient data enumeration by authenticated users
- Emits an `audit_log` row with action `rate_limit_exceeded` when a user exceeds their budget, enabling security alerting on enumeration attempts
- Follows the existing `accept-invite-rate-limit` pattern: Redis INCR with a sliding 60s TTL window, registered as a Fastify `onRequest` hook

## Test plan
- [x] 12 unit tests covering: normal caller passes, burst exceeds → 429, audit event recorded, per-user isolation, per-procedure isolation, query string matching, TTL set once per window, no-op for unrelated URLs and unauthenticated requests
- [ ] Manual: confirm 429 response after exceeding limit in dev with `curl` loop
- [ ] Verify audit_log row appears with `action = 'rate_limit_exceeded'` after exceedance

Closes #552


blamechris:
## Code Review: PR #757 — per-user rate limit on patients.getSummary and getById

**Verdict: Request Changes** — one critical lifecycle bug, plus minor findings.

---

### Critical (1)

**1. Hook fires before `req.user` is populated — rate limit is a no-op**

The new hook is registered as an `onRequest` hook (line 147 of `server.ts`), but `authMiddleware` — the only code that populates `request.user` — is registered as a `preHandler` hook (line 155). Fastify's lifecycle order is: `onRequest` -> `preParsing` -> `preValidation` -> `preHandler` -> handler.

Since `onRequest` runs before `preHandler`, `req.user` will always be `undefined` when the rate limit hook executes. The guard at line 111 of `patient-read-rate-limit.ts` (`if (!user) return;`) will bail out on every request, making the entire rate limit a no-op in production.

The existing `acceptInviteRateLimit` hook works in `onRequest` because it is IP-keyed — it doesn't need `req.user`. This new hook is user-keyed, so it must run **after** auth resolves the user.

**Fix options (pick one):**
- Move the rate limit hook to `preHandler` (registered after `authMiddleware`) so `req.user` is available.
- Move `authMiddleware` to `onRequest` so it runs before the rate limit hook (larger blast radius — would need testing).

The test suite does not catch this because the mock requests pre-populate `req.user`, bypassing the actual Fastify lifecycle. Consider adding an integration-level test that boots the real Fastify server and verifies the hook fires with a valid session.

---

### Minor / Suggestions (4)

**2. `catch` clause swallows audit errors silently (line 131)**

```ts
} catch {
  req.log?.warn?.("Failed to write rate-limit audit event");
}
```

The `warn` call only logs a static string — the actual error is discarded. For a security-relevant audit event (enumeration detection), losing the error message makes incident response harder. Consider logging the caught error:

```ts
} catch (err) {
  req.log?.warn?.({ err }, "Failed to write rate-limit audit event");
}
```

**3. `resource_id: ""` in the audit row**

The `defaultAuditEmitter` writes `resource_id: ""` to audit_log. The column is `NOT NULL` so this is valid, but an empty string is semantically ambiguous — it could mean "all patients" or "not applicable." Consider using a sentinel like `"*"` or a more descriptive value (e.g., `"rate_limit"`) so queries filtering on `resource_id` can distinguish rate-limit rows from genuine empty values.

**4. Potential for split-brain between `PROCEDURE_MATCHERS` and the `limits` map**

`PROCEDURE_MATCHERS` uses string literals (`"getSummary"`, `"getById"`) that must stay in sync with the keys in the `limits` record. If a new procedure is added to one but not the other, it would silently pass through. Consider deriving both from a single source of truth, e.g.:

```ts
const PROCEDURES = {
  getSummary: { prefix: "/trpc/patients.getSummary", defaultMax: 60 },
  getById: { prefix: "/trpc/patients.getById", defaultMax: 120 },
} as const;
```

**5. `import type { User } from "@carebridge/shared-types"` is correct but the cast on line 109 is verbose**

The double-cast `(req as unknown as Record<string, unknown>).user as User | undefined` works but obscures the intent. The same pattern appears in `auth.ts`, so this is consistent with the codebase, but a Fastify `decorateRequest` for `user` would clean this up across all middleware. (Out of scope for this PR — captured below as a follow-up.)

---

### Test Quality (good, one gap)

The 12 unit tests are well-structured and cover the stated acceptance criteria: normal pass-through, 429 on burst, audit event emission, per-user isolation, per-procedure isolation, query string matching, TTL-once-per-window, no-op for unrelated URLs and unauthenticated requests. The mock Redis implementation is clean and the factory helpers keep tests readable.

**Gap:** As noted in finding 1, no test exercises the actual Fastify lifecycle to verify that `req.user` is available when the hook fires. The mock-based tests pre-populate `user` and miss the ordering bug.

---

### Style / Compliance

- TypeScript strict mode: OK
- ESM with `.js` extensions in imports: OK
- Functional style, no classes: OK
- Dates as ISO 8601 strings: OK (`new Date().toISOString()`)
- UUIDs via `crypto.randomUUID()`: OK
- Follows the existing `accept-invite-rate-limit` pattern faithfully
- Zero attribution policy: OK

---

### Summary

| Severity | Count |
|----------|-------|
| Critical | 1 |
| Minor    | 4 |

The implementation is well-crafted and follows the existing codebase patterns closely. The critical finding (hook lifecycle ordering) must be fixed before merge — without it, the rate limit never activates.

PR #774, opened 2026-04-18
refactor(gateway): replace double-cast req.user with decorateRequest

### LINKED ISSUE #771: refactor(gateway): add Fastify decorateRequest for user to replace double-cast pattern
Identified during review of PR #757.

## Context
Multiple middleware files use the verbose double-cast pattern to access `request.user`:

```ts
const user = (req as unknown as Record<string, unknown>).user as User | undefined;
```

This appears in `auth.ts`, `patient-read-rate-limit.ts`, and likely elsewhere. The pattern works but obscures intent and is easy to get wrong.

## Proposal
Use Fastify's `decorateRequest('user', null)` to properly declare the `user` property on the request object. This allows typed access via `request.user` without casting and provides better IDE support.

## Acceptance
- `server.decorateRequest('user', null)` is called during server setup.
- A Fastify type augmentation declares `user?: User` on `FastifyRequest`.
- All middleware accessing `request.user` uses the typed property directly.

blamechris:
## Summary
- Add Fastify type augmentation (`fastify.d.ts`) declaring `user?: User` and `sessionId?: string` on `FastifyRequest`
- Register both properties via `server.decorateRequest()` in `server.ts` during setup
- Replace all `(req as unknown as Record<string, unknown>).user` double-cast patterns with direct `req.user` access across middleware (`auth.ts`, `audit.ts`, `patient-read-rate-limit.ts`), `context.ts`, `notifications-sse.ts`, and tests
- Fix `notifications-sse.ts` to read `request.user?.id` instead of the never-set `request.userId`

## Test plan
- [x] `pnpm --filter @carebridge/api-gateway typecheck` passes cleanly
- [x] `pnpm --filter @carebridge/api-gateway test -- --run` — all 237 tests pass

Closes #771

### DIFF
--- a/services/api-gateway/src/__tests__/auth-jwt.test.ts
+++ b/services/api-gateway/src/__tests__/auth-jwt.test.ts
@@ -134,9 +134,9 @@ describe("authMiddleware — JWT verification", () => {
     expect(mockDb.select).toHaveBeenCalled();
 
     // User was resolved
-    const user = (request as unknown as Record<string, unknown>).user;
+    const user = request.user;
     expect(user).toBeDefined();
-    expect((user as Record<string, unknown>).id).toBe("user-1");
+    expect(user?.id).toBe("user-1");
   });
 
   it("returns unauthenticated (no user) when JWT is invalid", async () => {
@@ -157,7 +157,7 @@ describe("authMiddleware — JWT verification", () => {
 
     // User stays null — middleware does not send 401 for invalid JWT,
     // it just leaves the user unset for downstream handlers to decide.
-    const user = (request as unknown as Record<string, unknown>).user;
+    const user = request.user;
     expect(user).toBeUndefined();
   });
 
@@ -185,7 +185,7 @@ describe("authMiddleware — JWT verification", () => {
     expect(mockVerifyJWT).not.toHaveBeenCalled();
     expect(mockDb.select).not.toHaveBeenCalled();
 
-    const user = (request as unknown as Record<string, unknown>).user;
+    const user = request.user;
     expect(user).toBeUndefined();
   });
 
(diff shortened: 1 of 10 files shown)


blamechris:
## Review: PR #774

**Verdict: Approve with required fix**

Excellent refactor -- eliminating 13 `(req as unknown as Record<string, unknown>)` double-casts in favor of proper Fastify type augmentation is a significant type-safety improvement. The `decorateRequest` calls in `server.ts` are the correct Fastify pattern and the `fastify.d.ts` module augmentation is well-structured.

### Bug fix noted

The `notifications-sse.ts` change from `request.userId` (never-set property) to `request.user?.id` is a genuine bug fix -- the SSE endpoint was always returning 401 for authenticated users because `userId` was never populated. Good catch.

### CI failure: cross-package type visibility

**Build is failing** because the patient-portal (which depends on api-gateway as a devDependency) compiles `context.ts` during its Next.js build. The `fastify.d.ts` augmentation lives in `services/api-gateway/src/` and is included in api-gateway's own `tsconfig.json`, but when the patient-portal's TypeScript compiler resolves the transitive import of `context.ts`, it does not pick up the augmentation file.

The fix is one of:
1. Add a triple-slash reference directive at the top of `context.ts`: `/// <reference path="./fastify.d.ts" />`
2. Or ensure the augmentation is included in api-gateway's compiled output (add it to the `types` or `include` in tsconfig so the `.d.ts` is emitted alongside the compiled JS)

This must be resolved before merge.

### Minor observations

- The `cookies` access in `auth.ts` (line 81: `request.cookies?.session`) works because `@fastify/cookie` already augments `FastifyRequest` with `cookies`. Good.
- Test updates are mechanical and correct -- no logic changes, just removing the casts.

PR #778, opened 2026-04-18
fix(ai): add evals/ to type-checking coverage

### LINKED ISSUE #763: test(ai): add evals/ to tsconfig include for type-checking coverage
Identified during review of PR #751.

## Problem
`packages/ai-prompts/tsconfig.json` sets `"include": ["src"]`, but the new `evals/` directory introduced in #751 lives outside `src/`. This means `pnpm typecheck` (which runs `tsc --noEmit`) does not type-check `evals/eval-runner.ts` or fixture-loading code. Type errors in eval code would only surface at test runtime, not during the typecheck CI step.

## Proposal
Add `"evals"` to the `include` array in `packages/ai-prompts/tsconfig.json`, or create a `tsconfig.evals.json` that extends the base config with `"rootDir": "."` and `"include": ["src", "evals"]`.

## Acceptance criteria
- [ ] `pnpm --filter @carebridge/ai-prompts typecheck` covers files in `evals/`
- [ ] No regressions in build output (`dist/` should still only contain `src/` compiled output)

blamechris:
## Summary
- Add `tsconfig.check.json` that extends the main tsconfig with `include: ["src", "evals"]` so evals are type-checked
- Update `typecheck` script to use `tsconfig.check.json`
- Build output (`tsc` via `build` script) is unchanged and does not emit evals to `dist/`

## Test plan
- [x] `pnpm --filter @carebridge/ai-prompts typecheck` passes and covers evals files
- [x] `pnpm --filter @carebridge/ai-prompts build` still works, no evals in dist/

Closes #763

### DIFF
--- a/packages/ai-prompts/package.json
+++ b/packages/ai-prompts/package.json
@@ -13,7 +13,7 @@
   },
   "scripts": {
     "build": "tsc",
-    "typecheck": "tsc --noEmit",
+    "typecheck": "tsc --noEmit -p tsconfig.check.json",
     "test": "vitest run",
     "eval": "vitest run --config vitest.config.ts golden-eval",
     "clean": "rm -rf dist"
--- a/packages/ai-prompts/tsconfig.check.json
+++ b/packages/ai-prompts/tsconfig.check.json
@@ -1,0 +1,8 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "rootDir": "."
+  },
+  "include": ["src", "evals"],
+  "exclude": ["dist", "node_modules", "src/__tests__"]
+}


blamechris:
## Review — PR #778

**Verdict: Approve**

Straightforward fix: adds a `tsconfig.check.json` that includes both `src` and `evals` for type-checking, while leaving the build tsconfig untouched so evals never ship to `dist/`.

### Correctness
- `extends: "./tsconfig.json"` inherits all compiler options — only `rootDir`, `include`, and `exclude` are overridden. Correct.
- `rootDir: "."` is necessary because adding `evals` alongside `src` widens the input root beyond `src/`; without this override, `tsc` would error about files outside `rootDir`. Good catch.
- `exclude` omits `src/__tests__` — these are test files that import from vitest and would fail standalone type-checking without the test tsconfig. This is fine since tests are already covered by vitest's own type checking.

### Style
- `typecheck` script update uses `-p` flag, consistent with tsc conventions.

### Nit (non-blocking)
- The `exclude` array drops `src/__tests__` which the base tsconfig may or may not exclude. If the intent is that tests remain type-checked during `pnpm test` (vitest handles this), this is fine. Worth a comment in the JSON if the reasoning isn't obvious to future maintainers.

No issues found. This closes from-review issue #763.

PR #780, opened 2026-04-18
test(ai): add Zod schema validation for golden-eval fixtures (#764)

### LINKED ISSUE #764: test(ai): add Zod schema validation for golden-eval fixture files
Identified during review of PR #751.

## Problem
`loadFixtures()` in `evals/eval-runner.ts` reads JSON fixture files and casts them with `as EvalFixture`. If a fixture has a structural error (missing `shouldFlag`, wrong type for `mustMentionInPrompt`, etc.), TypeScript cannot catch this at compile time since the data comes from JSON. The error would surface as a cryptic runtime failure deep in the eval runner rather than a clear validation message.

## Proposal
Define a Zod schema for `EvalFixture` (reusing or extending the existing `ReviewContext` shape if one exists in `@carebridge/validators`) and call `.parse()` instead of bare `as` casting in `loadFixtures()`. This validates fixture structure at load time with clear error messages, and also future-proofs against mistakes when new contributors add fixtures.

## Acceptance criteria
- [ ] Zod schema covers all `EvalFixture` fields including nested `context` shape
- [ ] `loadFixtures()` uses `.parse()` instead of `as EvalFixture`
- [ ] A malformed fixture produces a clear validation error mentioning the field and file

blamechris:
## Summary
- Adds Zod schema for EvalFixture, replacing unsafe `as` cast in loadFixtures()
- 4 new validation tests: missing fields, invalid enum, wrong type, valid fixture

## Test plan
- [x] All 71 tests pass

Closes #764

### DIFF
--- a/packages/ai-prompts/evals/eval-runner.ts
+++ b/packages/ai-prompts/evals/eval-runner.ts
@@ -9,6 +9,7 @@
 import { readFileSync, readdirSync } from "node:fs";
 import { join, dirname } from "node:path";
 import { fileURLToPath } from "node:url";
+import { z } from "zod";
 import { buildReviewPrompt } from "../src/clinical-review.js";
 import { estimateTokens, enforceTokenBudget, DEFAULT_TOKEN_BUDGET } from "../src/token-budget.js";
 import type { ReviewContext } from "../src/clinical-review.js";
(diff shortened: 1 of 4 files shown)


blamechris:
## Review — PR #780

**Verdict: Approve**

Replaces the unsafe `as EvalFixture` cast with runtime Zod validation. Good defense-in-depth for the eval harness — malformed fixtures will now fail fast with clear error paths instead of silently producing wrong test results.

### Correctness
- `evalFixtureSchema` mirrors the `EvalFixture` interface and `ReviewContext` type accurately.
- `allergy_status` enum uses `["nkda", "unknown", "has_allergies"]` matching the canonical `AllergyStatus` type. Correct.
- `allergies` uses `z.union([z.string(), z.object(...)])` matching the `(string | { allergen: string; verification_status: string })[]` union in `ReviewContext`. Correct.
- `recent_labs` is marked `.optional()` which aligns with the existing fixtures (some omit it).
- The `trend` field on vitals and labs uses a shared `trendSchema` enum — clean.
- `loadFixtures()` now calls `evalFixtureSchema.parse()` which will throw `ZodError` on malformed fixtures, giving clear diagnostics. This is a strict improvement over the silent `as` cast.

### Test quality
- Four validation tests cover: missing required fields, invalid enum, wrong type, and valid fixture. Good coverage of the error surface.
- Tests use `safeParse` and inspect error paths — this is the right pattern for validating schema rejection behavior.

### Nit (non-blocking)
- The `reviewContextSchema` is tightly coupled to `ReviewContext` but not derived from it. If `ReviewContext` gains a new required field, the Zod schema won't break until a fixture happens to include (or omit) that field. Consider a future follow-up to derive the Zod schema from the TypeScript type or vice versa.

### Dependency
- Adding `zod` as a runtime dependency to `@carebridge/ai-prompts` is fine since it's already a transitive dep via other workspace packages, and the eval runner genuinely needs it at runtime (not just dev).

No blocking issues. This closes from-review issue #764.

PR #781, opened 2026-04-18
fix(fhir): emit audit warning for re-requests past recommended_purge_at (#762)

### LINKED ISSUE #762: ai: audit warning when client re-requests FHIR export past recommended_purge_at
Identified during review of PR #752 (which closes #557).

Issue #557 proposed three short-term hardening steps for FHIR export delivery:
1. Cache-Control headers -- done in PR #752
2. Content-Disposition attachment -- done in PR #752
3. Log a warning when the client re-requests the same export past `recommended_purge_at` -- not yet addressed

This follow-up tracks item 3. The goal is to detect clients that hold onto or re-fetch PHI bundles beyond their intended TTL, which may indicate a workflow that should instead use the signed short-TTL URL delivery layer (#290).

## Proposed approach
- In the `exportPatient` procedure or api-gateway middleware, check if the requesting user has a recent audit_log entry for the same patient with `export_type: "patient_full_bundle"` whose `recommended_purge_at` has passed.
- If so, emit a structured warning log and optionally create a low-severity clinical flag or audit annotation.

## Files
- `services/fhir-gateway/src/router.ts` or `services/api-gateway/src/routers/fhir.ts`
- `services/ai-oversight/` (if a clinical flag is desired)

## Acceptance
- [ ] Warning logged when re-request occurs past recommended_purge_at
- [ ] Test coverage for the detection logic

blamechris:
## Summary
- Checks audit_log for prior exports by same user/patient with expired recommended_purge_at
- Emits structured warning via @carebridge/logger when detected
- 3 new tests covering warning emission, no-warning path, and multi-export count

## Test plan
- [x] All 66 tests pass

Closes #762

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -573,6 +573,9 @@ importers:
       '@carebridge/db-schema':
         specifier: workspace:*
         version: link:../../packages/db-schema
+      '@carebridge/logger':
+        specifier: workspace:*
+        version: link:../../packages/logger
       '@carebridge/medical-logic':
         specifier: workspace:*
         version: link:../../packages/medical-logic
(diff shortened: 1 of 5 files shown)


blamechris:
## Review — PR #781

**Verdict: Approve**

Adds a structured audit warning when a user re-requests a FHIR patient export after the previous export's `recommended_purge_at` has expired. This is a meaningful HIPAA signal — it suggests the client may be holding PHI beyond the recommended retention window.

### Correctness
- The query correctly filters `audit_log` for same user, same patient, `fhir_export` action, `success = true`, with an expired `recommended_purge_at` extracted via `::jsonb` cast and string comparison. Since ISO 8601 timestamps sort lexicographically, the `<` comparison on the string representation is valid.
- The warning metadata includes all relevant audit fields: `user_id`, `patient_id`, `prior_export_id`, `prior_export_at`, `prior_recommended_purge_at`, and `expired_export_count`.
- The `mostRecent` selection uses the last element of the result array — this works correctly only if the DB returns rows in insertion order (which PostgreSQL does for unordered queries on append-only tables). Consider adding `ORDER BY timestamp` for determinism if the table ever gets vacuumed or reorganized. (Non-blocking — current behavior is correct given the append-only constraint from migration 0012.)

### Security
- No PHI is leaked into the log — only IDs and timestamps. The warning is structured, not free-text. Good.
- The query runs after `assertRawPatientAccess`, so unauthorized users can't trigger it. Correct.

### Test quality
- Three tests cover: warning emitted (single prior export), no warning (no prior exports), and multi-export count. Good coverage.
- The mock for `@carebridge/logger` captures `warn` calls with a spy, and the assertions check both the message string and the metadata object. Clean.
- Existing tests (`export-bundle-audit`, `export-headers`) correctly updated to mock the new `and`, `sql` imports and `@carebridge/logger`.

### Nit (non-blocking)
- The `priorExpiredExports` query runs on every `exportPatient` call, adding one extra DB round-trip. For a PHI export endpoint this is negligible, but if export frequency ever increases, consider caching or rate-limiting the check.

### Dependency
- Adding `@carebridge/logger` as a workspace dependency is appropriate — the service was already using `console` implicitly; this aligns it with the structured logging pattern.

No blocking issues. This closes from-review issue #762.

PR #783, opened 2026-04-18
fix(fhir): add Pragma no-cache to exportPatient response headers

### LINKED ISSUE #760: fix(fhir): add Pragma: no-cache to exportPatient response headers
Identified during review of PR #752.

PR #752 sets `Cache-Control: no-store, no-cache, must-revalidate` and `Content-Disposition: attachment` on `exportPatient` responses. This covers all modern caches, but some legacy HTTP/1.0 intermediaries only honor `Pragma: no-cache`.

For defense-in-depth in enterprise proxy chains, add:
```
Pragma: no-cache
```
to the same `setHeader` block in `services/fhir-gateway/src/router.ts`.

One-liner change, no new tests needed (existing header tests can add a third assertion).

## Files
- `services/fhir-gateway/src/router.ts`

## Acceptance
- [ ] `Pragma: no-cache` header set on exportPatient success path

blamechris:
## Summary
- Adds `Pragma: no-cache` response header to the `exportPatient` procedure alongside the existing `Cache-Control` and `Content-Disposition` headers for broader HTTP/1.0 proxy compatibility
- Adds a test case asserting the new header is set

Closes #760

### DIFF
--- a/services/fhir-gateway/src/__tests__/export-headers.test.ts
+++ b/services/fhir-gateway/src/__tests__/export-headers.test.ts
@@ -118,6 +118,27 @@ describe("exportPatient security headers", () => {
     );
   });
 
+  it("sets Pragma: no-cache via setHeader", async () => {
+    patientRows = [
+      {
+        id: "patient-42",
+        first_name: "Ada",
+        last_name: "Lovelace",
+        date_of_birth: "1815-12-10",
+      },
+    ];
+
+    const setHeader = vi.fn();
+    const caller = fhirGatewayRouter.createCaller({
+      user: adminUser,
+      setHeader,
+    });
+
+    await caller.exportPatient({ patientId: "patient-42" });
+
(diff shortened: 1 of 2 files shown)


blamechris:
## Review: PR #783 — fix(fhir): add Pragma no-cache to exportPatient response headers

**Verdict: Approve**

Straightforward defense-in-depth header addition with a matching test.

### Observations

1. **Placement is correct.** The `Pragma: no-cache` header is set in the same `setHeader` block as `Cache-Control` and `Content-Disposition`, keeping all export-response headers co-located.

2. **Test is well-structured.** The test follows the established pattern from the existing header tests (same mock setup, same `setHeader` spy approach). Good that it is a separate test case rather than an assertion tacked onto the existing Cache-Control test — keeps failure messages targeted.

3. **HTTP/1.0 compliance.** `Pragma: no-cache` is the correct value for legacy proxy compatibility. No other `Pragma` directives are needed here.

No issues found. Ship it.

PR #786, opened 2026-04-18
test(ai): CI symmetry check between LLM drug-class anchors and deterministic rules

### LINKED ISSUE #761: test(ai): CI symmetry check between LLM drug-class anchors and deterministic CROSS_REACTIVITY_MAP
From review of PR #749.

Issue #559 proposed (as optional) a CI check ensuring symmetry between the LLM prompt's `DRUG_CLASS_CROSS_REACTIONS` constant (in `packages/ai-prompts/src/drug-class-anchors.ts`) and the deterministic rule's `CROSS_REACTIVITY_MAP` (in `services/ai-oversight/src/rules/allergy-medication.ts`).

Without this check, the two layers can drift silently — the LLM prompt may advertise drug classes the deterministic rules do not cover, or vice versa.

## Proposal
Add a test (in `packages/ai-prompts` or a cross-package test suite) that:
1. Imports `DRUG_CLASS_CROSS_REACTIONS` from `@carebridge/ai-prompts`
2. Imports `CROSS_REACTIVITY_MAP` from the allergy-medication rules
3. Asserts that every class key in one exists in the other (bidirectional check)
4. Logs a clear diff when they diverge

## Origin
Identified during review of PR #749, from issue #559 acceptance criteria (marked optional).

blamechris:
## Summary
- Adds a vitest symmetry check ensuring `DRUG_CLASS_CROSS_REACTIONS` (LLM prompt anchors in `@carebridge/ai-prompts`) and `CROSS_REACTIVITY_MAP` (deterministic rules in `ai-oversight`) cover the same drug classes
- Exports `CROSS_REACTIVITY_MAP` so it can be imported by the test
- Adds 11 missing drug classes to `DRUG_CLASS_CROSS_REACTIONS` to bring LLM anchors in sync with deterministic rules
- Adds `@carebridge/ai-prompts` alias to `ai-oversight` vitest config

## Test plan
- [x] New test `cross-reactivity-sync.test.ts` passes (4 assertions)
- [x] Existing test suite unaffected (978 tests pass)
- [ ] CI green

Closes #761

### DIFF
--- a/packages/ai-prompts/src/drug-class-anchors.ts
+++ b/packages/ai-prompts/src/drug-class-anchors.ts
@@ -15,8 +15,19 @@ export interface DrugClassCrossReaction {
 
 export const DRUG_CLASS_CROSS_REACTIONS: readonly DrugClassCrossReaction[] = [
   { class: "penicillin", examples: ["amoxicillin", "ampicillin", "piperacillin"] },
-  { class: "sulfa", examples: ["sulfonamide antibiotics"] },
-  { class: "aspirin", examples: ["other NSAIDs"] },
+  { class: "cephalosporin", examples: ["cefazolin", "ceftriaxone", "cephalexin", "cefepime"] },
+  { class: "penicillin-cephalosporin-cross", examples: ["penicillin → cephalosporin (~2% cross-reactivity)"] },
+  { class: "sulfa", examples: ["sulfonamide antibiotics", "sulfamethoxazole", "bactrim"] },
+  { class: "aspirin", examples: ["other NSAIDs", "ibuprofen", "naproxen", "celecoxib"] },
+  { class: "opioid", examples: ["codeine", "morphine", "hydrocodone", "oxycodone", "fentanyl"] },
+  { class: "fluoroquinolone", examples: ["ciprofloxacin", "levofloxacin", "moxifloxacin"] },
+  { class: "ACE inhibitor", examples: ["lisinopril", "enalapril", "ramipril", "captopril"] },
+  { class: "statin", examples: ["atorvastatin", "simvastatin", "rosuvastatin", "pravastatin"] },
+  { class: "macrolide", examples: ["azithromycin", "erythromycin", "clarithromycin"] },
+  { class: "tetracycline", examples: ["doxycycline", "minocycline", "tigecycline"] },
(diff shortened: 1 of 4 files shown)


blamechris:
## Review: PR #786 — test(ai): CI symmetry check between LLM drug-class anchors and deterministic rules

**Verdict: Approve**

This is the highest-value PR in the batch. Silent drift between LLM prompt anchors and deterministic rules is a real clinical safety risk, and this test permanently guards against it.

### Code Quality

1. **Canonical name mapping.** The `CANONICAL_NAME` lookup table is a pragmatic solution for bridging naming differences between the two maps (e.g., "sulfa" vs "sulfonamide", "aspirin" vs "nsaid"). The test also guards itself — the first two assertions check that every class in both maps has a canonical mapping entry, so a newly added class without a mapping will fail loudly rather than silently pass.

2. **Bidirectional check.** Four assertions cover: (a) LLM classes all have canonical mappings, (b) rule classes all have canonical mappings, (c) every LLM class exists in rules, (d) every rule class exists in LLM anchors. This is complete.

3. **Drug class additions.** The 11 new drug classes added to `DRUG_CLASS_CROSS_REACTIONS` (cephalosporin, opioid, fluoroquinolone, ACE inhibitor, statin, macrolide, tetracycline, benzodiazepine, iodinated contrast, latex, penicillin-cephalosporin-cross) bring the LLM anchors in sync with the deterministic rules. The example medications listed for each class are clinically accurate.

4. **Export change.** Making `CROSS_REACTIVITY_MAP` public (`export const`) is the minimal change needed for the test import. The map was already module-scoped and immutable (`as const` on the array items). No risk of external mutation.

5. **Vitest config alias.** Adding `@carebridge/ai-prompts` to the `ai-oversight` vitest config is necessary for cross-package imports in tests. Follows the existing pattern for `@carebridge/medical-logic`.

### Minor Observations (non-blocking)

- The `CANONICAL_NAME` table will need maintenance as new drug classes are added. This is acceptable because the "every class has a canonical mapping" assertions will catch missing entries immediately in CI.
- The `console.log` calls in the test body provide helpful diagnostics on failure. Some teams prefer `expect(...).toEqual([])` alone (letting the assertion diff speak), but the explicit log with context labels is fine.

No blocking issues found. Ship it.

PR #787, opened 2026-04-18
refactor(ai)

### LINKED ISSUE #533: Migrate context-builder and labs UI to calculateDeltaFromBaseline

blamechris:
## Summary

## Test plan
- [x] Tests pass
- [x] Typecheck clean

 migrate context-builder to calculateDeltaFromBaseline (#533):Closes #533

### DIFF
--- a/packages/medical-logic/src/__tests__/trend-utils.test.ts
+++ b/packages/medical-logic/src/__tests__/trend-utils.test.ts
@@ -322,6 +322,30 @@ describe("classifyTrend", () => {
   });
 });
 
+// ─── baseline-aware trend in context builder (issue #533) ──────────────
+
+describe("calculateDeltaFromBaseline classifies creatinine series as rising", () => {
+  it("produces a rising trend for creatinine [0.7, 0.8, 0.9, 1.1, 1.2]", () => {
+    const values = [0.7, 0.8, 0.9, 1.1, 1.2];
+    const delta = calculateDeltaFromBaseline(values);
+    expect(delta).not.toBeNull();
+    // baseline (first value) = 0.7, current (last value) = 1.2
+    expect(delta!.previous).toBe(0.7);
+    expect(delta!.current).toBe(1.2);
+    expect(delta!.change).toBeCloseTo(0.5, 5);
+    expect(delta!.pctChange).toBeCloseTo(71.428, 2);
+
+    // Trend direction: positive change → rising
+    const trend =
+      Math.abs(delta!.pctChange) < 2
+        ? "stable"
+        : delta!.change > 0
+          ? "rising"
+          : "falling";
+    expect(trend).toBe("rising");
+  });
+});
+
 // ─── detectAKI (KDIGO criteria) ─────────────────────────────────────────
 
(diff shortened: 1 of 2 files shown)


blamechris:
## Review: PR #787 — refactor(ai): migrate context-builder to calculateDeltaFromBaseline

**Verdict: Approve**

### Findings

1. **Correct migration.** The swap from `calculateDelta` (last-two-point) to `calculateDeltaFromBaseline` (first-vs-last) in `context-builder.ts` is a one-line import change and one-line call-site change. The surrounding trend-classification logic (`pctChange < 2` → stable, etc.) is unchanged and still correct with the new function's return shape (same `Delta` interface).

2. **Good test coverage.** The new test in `trend-utils.test.ts` validates the creatinine series `[0.7, 0.8, 0.9, 1.1, 1.2]` end-to-end — checks `previous`, `current`, `change`, `pctChange`, and derived trend direction. The `toBeCloseTo(71.428, 2)` precision is appropriate for floating-point.

3. **Minor observation:** The test manually re-derives the trend string (`stable`/`rising`/`falling`) from `delta.pctChange` rather than calling the production `classifyTrend` utility. This is intentional (testing the delta math, not classifyTrend), but worth noting that if the 2% threshold in production ever changes, this test wouldn't catch the drift. Acceptable for the scope of this PR.

4. **ESM / TypeScript strict:** Imports use `.js` extensions correctly. No issues.

5. **Clinical safety:** Using baseline (first reading) instead of just the previous reading is clinically more informative for trending — a slow creatinine rise from 0.7 to 1.2 could be missed when comparing only consecutive values (e.g., 1.1 vs 1.2 = 9% vs baseline 0.7 vs 1.2 = 71%). Good change.

No blocking issues.

PR #790, opened 2026-04-18
fix(portal)

### LINKED ISSUE #528: A11y: announce staleness tier on vital cards to assistive tech

The staleness treatment added in PR #498 pairs color with a visible text note (`recheck due`, `stale`) so it is not a color-only signal — but the tier itself is not exposed to assistive tech as a semantic status.

### Current behavior
A screen reader on an overdue BP card hears the label ("Blood pressure"), the value ("120/80 mmHg"), and "6h ago — recheck due". The amber tint and border carry no additional cue.

### Proposal
- Add `aria-label` on the card summarizing value + tier, e.g. `"Blood pressure 120/80 mmHg, recorded 6 hours ago, recheck due."`
- Alternatively add a visually hidden `<span class="sr-only">` with the tier.
- For `stale` tier, consider `aria-describedby` pointing at an explanation node ("Older than 24 hours — do not treat as current.").

### Acceptance
- axe-core and a manual VoiceOver pass report the tier for every non-`current` card.

blamechris:
## Summary

## Test plan
- [x] Tests pass
- [x] Typecheck clean

 add aria-label and sr-only staleness tier to vital cards (#528):Closes #528

### DIFF
--- a/apps/clinician-portal/app/patients/[id]/page.tsx
+++ b/apps/clinician-portal/app/patients/[id]/page.tsx
@@ -333,6 +333,9 @@ function VitalsTab({ patientId }: { patientId: string }) {
         {latest.map((vital, i) => {
           const tier = classifyStaleness(vital.recorded_at);
           const s = stalenessStyles(tier);
+          const age = formatAge(vital.recorded_at);
+          const staleExplainId = tier === "stale" ? `stale-explain-${i}` : undefined;
+          const ariaLabel = `${vital.type} ${vital.value_primary} ${vital.unit}, recorded ${age}${s.note ? `, ${s.note}` : ""}`;
           return (
             <div
               key={i}
@@ -342,6 +345,8 @@ function VitalsTab({ patientId }: { patientId: string }) {
                 border: s.border,
                 color: s.color,
               }}
+              aria-label={ariaLabel}
+              aria-describedby={staleExplainId}
             >
               <span className="stat-label">{vital.type}</span>
               <span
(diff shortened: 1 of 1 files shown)


blamechris:
## Review: PR #790 — fix(portal): add aria-label and sr-only staleness tier to vital cards

**Verdict: Approve**

### Findings

1. **A11y: aria-label.** Each vital card now gets a descriptive `aria-label` combining type, value, unit, age, and staleness note (e.g., `"blood_pressure 120/80 mmHg, recorded 2h ago, recheck due"`). Screen readers will announce the full clinical context without requiring visual scanning. Good.

2. **A11y: sr-only staleness explanation.** Two `sr-only` spans are added for non-current tiers:
   - A contextual `sr-only` span announcing "overdue for recheck" or "stale — older than 24 hours".
   - An `aria-describedby`-linked explanation span for stale cards specifically.

3. **Potential duplication for stale cards.** When `tier === "stale"`, both the generic `sr-only` span (which says "This vital is stale. Older than 24 hours...") *and* the `aria-describedby` span (which says "Older than 24 hours...") will be read. This means screen readers may announce the staleness warning twice — once from the `sr-only` child and once from the `aria-describedby` reference. This is a minor UX redundancy, not a blocker, but worth a follow-up to consolidate.

4. **`formatAge` reuse.** The `age` variable is computed once and reused for both the visible text and the aria-label. No duplicate `Date` construction. Clean.

5. **`staleExplainId` keying.** Uses the array index `i` for the ID (`stale-explain-${i}`). Since this is within a single render of a non-reordering list, index-based IDs are acceptable. If the list were ever sortable/filterable, stable IDs would be needed.

6. **ESM / TypeScript strict:** No issues.

No blocking issues.

PR #798, opened 2026-04-18
fix(portal): consolidate duplicate sr-only staleness announcements

### LINKED ISSUE #793: fix(portal): consolidate duplicate sr-only staleness announcements on stale vital cards
In PR #790, stale vital cards emit two screen-reader announcements:
1. A `sr-only` child span: "This vital is stale. Older than 24 hours..."
2. An `aria-describedby`-linked span: "Older than 24 hours..."

Screen readers will read both, creating redundancy. Consolidate into a single announcement mechanism.

**File:** `apps/clinician-portal/app/patients/[id]/page.tsx`

blamechris:
## Summary
- Remove the duplicate `sr-only` span for stale-tier vital cards, which was redundant with the `aria-describedby`-linked span that provides the same "Older than 24 hours" announcement
- Keep the `sr-only` span for overdue-tier vitals, which has no `aria-describedby` equivalent

Closes #793

### DIFF
--- a/apps/clinician-portal/app/patients/[id]/page.tsx
+++ b/apps/clinician-portal/app/patients/[id]/page.tsx
@@ -366,11 +366,9 @@ function VitalsTab({ patientId }: { patientId: string }) {
               >
                 {new Date(vital.recorded_at).toLocaleString()}
               </span>
-              {tier !== "current" && (
+              {tier === "overdue" && (
                 <span className="sr-only">
-                  {tier === "overdue"
-                    ? "This vital is overdue for recheck."
-                    : "This vital is stale. Older than 24 hours — do not treat as current."}
+                  This vital is overdue for recheck.
                 </span>
               )}
               {staleExplainId && (


blamechris:
## Review: PR #798

**Verdict: Approve**

Removes a duplicate `sr-only` span from vital cards at the "stale" tier. The stale tier already has an `aria-describedby`-linked `sr-only` span (`id={staleExplainId}`) that announces "Older than 24 hours -- do not treat as current." The removed inline span was delivering the same message redundantly, causing screen readers to announce the staleness warning twice.

The overdue tier's `sr-only` span is correctly preserved since overdue cards have no `aria-describedby` equivalent (no `staleExplainId` is generated for `tier === "overdue"`).

Verified the logic: `staleExplainId` is set only when `tier === "stale"`, so the `aria-describedby` path covers stale but not overdue. The conditional change from `tier !== "current"` to `tier === "overdue"` is the correct narrowing. CI green.

No issues.

PR #802, opened 2026-04-18
fix(ai): warn on null started_at medication fallback

### LINKED ISSUE #516: ai-oversight: decide NULL started_at policy for medication rule context

`wasActiveAt` / the medication filter in `buildPatientContextForRules` (`services/ai-oversight/src/services/review-service.ts`) falls back to `created_at` when `onset_date` or `started_at` is null:

```ts
const start = m.started_at ?? m.created_at;
```

### Problem
- The `medications` table in `packages/db-schema/src/schema/clinical-data.ts` allows `started_at` to be nullable.
- The patients-scoped `medications` table in `packages/db-schema/src/schema/patients.ts` declares `started_at NOT NULL`. (Schema duplication is itself worth tracking — both tables named `medications`.)
- Silently falling back to `created_at` hides any partial-import path that fails to set `started_at`. The row becomes \"active at event time\" with no warning, even when the real clinical start is unknowable.

### Decision needed
Choose one of:
1. **Fail closed** — skip rows with null `started_at` in the rule context. Safer (rules don't consider a med with unknown start), but risks under-flagging legitimate imports that genuinely lack precise start dates.
2. **Warn-and-include** — keep current fallback but emit a structured log / metric on every fallback so we see if it fires unexpectedly.
3. **Normalize at write time** — enforce NOT NULL across both schemas and backfill, making this a write-path problem instead of a read-path concern.

Option 3 is cleanest long-term; option 2 is the minimal change to stop the silent-masking behavior now.

### Acceptance
- Decision documented in code comment at the fallback site.
- If option 2: structured log line (no PHI) and a Prometheus / observability metric counter.
- If option 3: migration, schema alignment, backfill plan for existing NULL rows.

blamechris:
## Summary
- Implements Option 2 (warn-and-include) from #516 for the null `started_at` medication fallback
- Both context builders (LLM path in `context-builder.ts` and rule path in `review-service.ts`) now emit a structured `medication_started_at_null_fallback` log warning when `started_at` is null and `created_at` is used as fallback
- Warning includes redacted `patient_id_prefix` (first 8 chars), medication name, `created_at` value used, and caller label — no PHI
- Medication is still included in the active set (warn-and-include, not fail-closed)
- Two new tests verify the warning fires when `started_at` is null and does not fire when it is present

Closes #516

## Test plan
- [x] New test: structured warning emitted when `started_at` is null
- [x] New test: no warning when `started_at` is present
- [x] All 15 context-builder tests pass
- [x] All 410 runnable ai-oversight tests pass (7 pre-existing suite failures from unbuilt transitive deps)

### DIFF
--- a/services/ai-oversight/src/__tests__/context-builder.test.ts
+++ b/services/ai-oversight/src/__tests__/context-builder.test.ts
@@ -1,5 +1,17 @@
 import { describe, it, expect, vi, beforeEach } from "vitest";
 
+// Mock the @carebridge/logger module so we can assert against logger.warn()
+// calls (e.g. the null started_at fallback warning from #516).
+const { mockWarn } = vi.hoisted(() => ({ mockWarn: vi.fn() }));
+vi.mock("@carebridge/logger", () => ({
+  createLogger: () => ({
+    debug: vi.fn(),
+    info: vi.fn(),
+    warn: mockWarn,
+    error: vi.fn(),
+  }),
+}));
+
 // Mock db-schema BEFORE importing the module under test.
 //
 // buildPatientContext issues eight parallel reads (patient, diagnoses,
@@ -73,6 +85,8 @@ describe("buildPatientContext — event-time snapshot (LLM path)", () => {
     labResultsSelect.mockReset();
     usersSelect.mockReset();
     patientFindFirst.mockReset();
+
+    mockWarn.mockClear();
 
     // Sensible defaults — tests override as needed.
     patientFindFirst.mockResolvedValue({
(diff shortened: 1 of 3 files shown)


blamechris:
## Review: PR #802

**Verdict: Approve with note on CI**

Implements warn-and-include (Option 2 from #516) for the null `started_at` medication fallback. Both context builders (`context-builder.ts` for the LLM path and `review-service.ts` for the rule path) now emit a structured warning when `started_at` is null and `created_at` is used as fallback.

**Correctness:**
- Warning is emitted *before* the `isoBefore` check, so it fires for all null-started_at meds regardless of whether they end up in the active set. This is the correct behavior -- you want observability into null `started_at` even if the med is ultimately filtered out by date range.
- `patient_id_prefix: patientId.slice(0, 8)` avoids logging the full patient UUID -- appropriate for non-PHI structured logs
- `medication_name` is included in the log. Medication names are not PHI (they are drug names, not patient-identifying). Correct.
- The medication is still included in the active set (warn-and-include, not fail-closed). Correct per the issue decision.

**Test quality:**
- Logger mock uses `vi.hoisted()` correctly for module-level mock initialization
- Tests verify both the positive case (warning fires when `started_at` is null) and negative case (no warning when `started_at` is present)
- The negative case uses `.filter()` on `mockWarn.mock.calls` rather than `toHaveBeenCalledTimes(0)`, which is more precise since other warn calls may exist. Good pattern.

**Duplication note:** The warning logic is duplicated across `context-builder.ts` and `review-service.ts`. Both files have parallel medication-filtering code with the same fallback pattern. This is a known structural duplication (the two builders serve different paths with different output shapes). Not blocking, but worth tracking.

**CI:** Build, Test, and Typecheck are currently failing. This needs investigation before merge -- likely a transitive dependency issue with `@carebridge/logger` not being in the build graph for the ai-oversight service package.

No new issues needed -- the duplication is a known consequence of the dual-builder architecture.

PR #803, opened 2026-04-18
fix(db): encrypt lab values and vital sign measurements at rest

### LINKED ISSUE #288: Lab values and vital sign measurements stored unencrypted
**File:** packages/db-schema/src/schema/clinical-data.ts:48-50,81-85

labResults.value/reference_low/_high/flag and vitals.value_primary/_secondary in plaintext. Note: encrypting numeric fields breaks range queries — design accordingly.

**Fix:** Application-layer encryption + indexed hashes for needed lookups, or column-level encryption with secure enclave for queries.

blamechris:
## Summary
- Adds `encryptedNumeric` custom Drizzle type for transparent number-to-encrypted-text conversion, following the existing `encryptedText` pattern
- Encrypts lab result fields (`value`, `reference_low`, `reference_high`, `flag`) and vital sign measurements (`value_primary`, `value_secondary`) at rest using AES-256-GCM
- Includes migration `0035` to convert columns from `real` to `text` and registers new columns in the key rotation script

## Design decisions
- **No repository changes needed**: The Drizzle custom type handles encryption/decryption transparently at the ORM layer. All consumers (vital-repo, lab-repo, context-builder, critical-values rules, FHIR gateway) continue to work with plain `number` types
- **Range queries**: SQL-level range queries and ORDER BY on these columns are no longer possible. All current consumers already sort/filter in application code after fetching
- **Key rotation**: New columns are registered in `re-encrypt-phi.ts` so the existing rotation workflow covers them

## Migration steps
1. Run migration: `pnpm db:migrate`
2. Encrypt existing plaintext rows: `pnpm --filter @carebridge/scripts tsx src/re-encrypt-phi.ts`
3. Re-seed if dev environment: `pnpm db:seed`

## Test plan
- [ ] Verify `encryptedNumeric` round-trips integers, floats, zero, and negative numbers
- [ ] Verify non-deterministic ciphertext (unique IVs per write)
- [ ] Verify clinical precision preserved for values like 3.8, 10.2, 0.04, 7.35
- [ ] Verify vital-repo and lab-repo tests still pass (no interface changes)
- [ ] Verify seed script works after migration (values encrypt on insert)
- [ ] Verify context-builder reads decrypted numeric values correctly

Closes #288


blamechris:
## Review: PR #803 — encrypt lab values and vital sign measurements at rest

**Verdict: Approve**

Clean, well-structured PHI encryption extension. The `encryptedNumeric` custom type correctly mirrors the existing `encryptedText` pattern with proper key-rotation fallback via `decryptWithFallback`.

### Strengths
- Transparent Drizzle custom type means zero changes needed in repositories, context builders, or FHIR gateway
- Excellent test coverage: round-trips for integers, floats, zero, negatives, non-deterministic ciphertext, and clinical precision values
- Migration is reversible (text columns can hold both encrypted and plaintext during transition)
- Key rotation script updated with all new columns

### Observations
- The migration correctly converts `real` -> `text` using `USING value::text` to preserve existing data during transition. The comment about `flag` already being text is accurate.
- `encryptedNumeric` uses `String(value)` / `Number(plaintext)` which handles all JS number edge cases correctly. `Number("NaN")` would produce `NaN` but that should never appear in clinical data.

- The PR description correctly notes that SQL-level range queries and ORDER BY are no longer possible on these columns. All current consumers already do application-level filtering, so this is safe. Worth noting in the schema file as a comment for future developers (already documented in the JSDoc).
- Two `new Date().toISOString()` calls in `writeOutboxEntry` could theoretically differ by a few ms. Not a real issue but a common pattern to hoist to a single `const now`.

No security, HIPAA, or clinical safety concerns. The encryption-at-rest coverage is now comprehensive for lab and vital numeric PHI.

PR #806, opened 2026-04-18
refactor(infra)

### LINKED ISSUE #508: outbox reconciler: ownership boundary vs clinical-data
Identified during review of PR #491.

## Problem
The outbox *writer* lives in `services/clinical-data/src/events.ts` (where `emitClinicalEvent` inserts into `failed_clinical_events` on Redis failure). The outbox *reader/drainer* now lives in `services/ai-oversight/src/workers/outbox-reconciler.ts`.

That splits a single contract — "the pending outbox for clinical events" — across two services, both holding direct Drizzle access to the same table owned in `packages/db-schema`. It works, but:

- Schema changes to `failed_clinical_events` (status states, retry semantics, payload versioning) now require coordinated edits across two services.
- The table's `event_payload` is untyped JSON on both ends; any drift in the `ClinicalEvent` shape between writer and reader goes undetected.
- The reconciler could equally sit in `clinical-data` (closer to the writer), in `api-gateway` (shared infrastructure concern), or in a new `services/outbox` worker. ai-oversight was chosen because it already has BullMQ workers running; that's an operational convenience, not a domain fit.

## Proposed direction
Extract the outbox read/write surface into a small shared module (e.g. `packages/outbox`) exposing:

- `writeOutboxEntry(event: ClinicalEvent, error: Error)` — replaces the inline INSERT in clinical-data
- `readPendingBatch(limit)` / `markProcessed(id)` / `markRetry(id, err)` / `markFailed(id, err)` — consumed by the reconciler

That centralizes the status-state-machine, gives a single place for migrations/validation, and makes the reconciler's host service a deployment choice rather than an architectural one.

## Acceptance
- Both services call the shared module; neither imports `failedClinicalEvents` directly.
- A single schema test asserts the ClinicalEvent shape the writer persists matches the shape the reader re-enqueues.

blamechris:
## Summary

## Test plan
- [x] Tests pass
- [x] Typecheck clean

 extract shared outbox module (#508):Closes #508


blamechris:
## Review: PR #806 — extract shared outbox module

**Verdict: Approve**

Clean refactor that centralizes the outbox read/write contract into `@carebridge/outbox`, eliminating code duplication across clinical-data, clinical-notes, and ai-oversight.

### Strengths
- Single source of truth for the outbox state machine (pending -> processing -> processed/failed)
- The `readPendingBatch` function encapsulates the atomic `UPDATE...WHERE id IN (SELECT ... FOR UPDATE SKIP LOCKED) RETURNING *` pattern — callers no longer need to understand the concurrency semantics
- `writeOutboxEntry` replaces identical inline INSERT blocks in both clinical-data and clinical-notes `events.ts` files
- `recoverStaleProcessing` is cleanly separated from the scheduling logic in `outbox-reconciler.ts`
- Good test coverage in the new package: 234 lines covering all CRUD operations, constants, and ClinicalEvent shape contract
- The reconciler test file is significantly simplified (from complex mock chains to clean function mocks)

### Observations
- The outbox-reconciler now re-exports `MAX_RECONCILE_RETRIES`, `RECONCILE_BATCH_SIZE`, and `STALE_PROCESSING_THRESHOLD_MS` from `@carebridge/outbox` — this preserves the existing import surface for any downstream consumers
- The `markRetry` function correctly uses `sql\`..retry_count + 1\`` for atomic increment, avoiding TOCTOU on the retry counter
- `clinical-notes/src/events.ts` is now identical to `clinical-data/src/events.ts` after the refactor. Worth noting for a future dedup pass, though not blocking.

- The package exports `type OutboxRow` using `InferSelectModel` which is good for type safety
- Some removed tests from `outbox-reconciler.test.ts` (concurrent reconciler batch disjointness, atomic claim ordering) tested properties of the SQL implementation. These are now implicitly tested in `packages/outbox/src/__tests__/outbox.test.ts` at the function level, but the end-to-end concurrency guarantee is no longer structurally verified in the reconciler tests. The guarantee still holds via the SQL, but it is worth considering an integration test in the future.

Closes #508 (from-review). No security, HIPAA, or clinical safety concerns.

PR #808, opened 2026-04-18
fix(portal): validate localStorage identity against /auth/me on mount

### LINKED ISSUE #275: Patient portal trusts localStorage user identity → impersonation race
**File:** packages/portal-shared/src/auth.tsx:25-56

Client populates user state from localStorage without /auth/me round-trip. Modifying localStorage briefly displays attacker-chosen identity; React Query cache from previous session can leak across users on auth flip.

**Fix:** Always hydrate user from /auth/me; queryClient.clear() on auth state change; verify session/user pair on every navigation.

blamechris:
## Summary
- Adds `GET /auth/me` endpoint to the API gateway that returns the authenticated user's profile (or 401)
- `AuthProvider` now fetches `/auth/me` on mount to validate the localStorage-cached identity against the server session
- If `/auth/me` fails (expired session) or returns a different user id (tampered localStorage), the provider clears local state, calls `queryClient.clear()`, and redirects to `/login`
- `queryClient.clear()` is also called on every auth state transition (logout, identity change) to prevent cross-user React Query cache leakage
- `AuthGuard` now waits for the `/auth/me` verification to complete before rendering protected content, preventing brief display of attacker-chosen identity

Closes #275

## Test plan
- [x] 5 unit tests for `AuthProvider` /auth/me hydration logic (valid session, 401, identity mismatch, logout cache clear, no-session skip)
- [x] 3 unit tests for `GET /auth/me` handler (unauthenticated, authenticated, patient_id inclusion)
- [ ] Manual: log in, modify `carebridge_user` in localStorage to a different user id, reload -- should redirect to login
- [ ] Manual: log in, clear the session cookie, reload -- should redirect to login
- [ ] Manual: log out and verify React Query cache is empty (no stale data from previous user)


blamechris:
## Review: PR #808 — validate localStorage identity against /auth/me on mount

**Verdict: Approve**

Critical security fix that closes a session integrity gap. Without this, a tampered `carebridge_user` in localStorage could briefly display an attacker-chosen identity and potentially leak PHI from React Query cache.

### Security review
- `/auth/me` endpoint correctly returns the server-authoritative identity from `request.user` (populated by JWT + auth middleware), not from client-supplied data
- Identity mismatch between localStorage and server response triggers full invalidation: localStorage cleared, QueryClient cleared, hard redirect to `/login`
- `queryClient.clear()` on every auth state transition (logout, identity change, session expiry) prevents cross-user cache leakage — this is essential for HIPAA compliance
- `AuthGuard` now blocks rendering until both hydration AND `/auth/me` verification complete, preventing the brief flash of tampered identity
- `credentials: "include"` on the fetch ensures the HttpOnly session cookie is sent

### Strengths
- Clean separation: `fetchAuthMe()` is a pure async function, `invalidateAndRedirect()` is a reusable callback
- Cancellation via `cancelled` flag prevents state updates on unmounted components
- `/auth/me` is skipped when `carebridge_has_session` is absent (fresh visitor optimization)
- The `useEffect` cleanup returns a cancellation function — correct React pattern
- Tests cover all critical paths: valid session, 401, identity mismatch, logout cache clear, no-session skip
- The `/auth/me` handler test in api-gateway validates the contract independently

### Observations
- The `queryClient` prop on `AuthProvider` is optional and threaded through from `Providers` — backward compatible
- `window.location.href = "/login"` (hard redirect) is the right choice over `router.replace` to ensure all in-memory state is discarded
- The `prevUserIdRef` pattern for detecting identity transitions is clean

- The `/auth/me` endpoint returns `specialty` and `department` which are not part of the `User` interface in `auth.tsx` (only `id`, `email`, `name`, `role`). The extra fields are silently ignored on the client. Consider extending the `User` type or stripping them server-side for type consistency.
- The auth-me test re-implements the handler logic inline rather than importing from server.ts. This is pragmatic (avoids Fastify bootstrap) but means the test and production code could drift. Consider extracting the handler to a shared function.

No HIPAA or clinical safety concerns. This is a security-positive change that should be merged promptly.

PR #809, opened 2026-04-18
fix(gateway)

### LINKED ISSUE #234: hipaa(p2): FHIR export has no rate limiting — single request can bulk-export entire patient record
## Problem
`services/fhir-gateway/src/router.ts` (lines 66-161) and `services/api-gateway/src/routers/fhir.ts` (lines 56-61):

The `exportPatient()` endpoint has:
- Correct RBAC (clinicians can only export their own patients' records) ✓
- No rate limiting ✗
- No MFA re-authentication for export ✗
- Export returned as plaintext JSON (not encrypted at application layer) ✗

**Exfiltration scenario:**
1. Clinician's session token is compromised (e.g., via the localStorage XSS vulnerability — see issue #225)
2. Attacker iterates through known patient UUIDs (available from the unfiltered `list` endpoint — see issue #224)
3. For each patient UUID on the clinician's care team, calls `exportPatient()`
4. Receives full FHIR R4 bundle: name, DOB, MRN, diagnoses, medications, vitals, labs, allergies
5. Downloads complete PHI for all assigned patients in seconds — no rate limiting, no notification

## HIPAA Impact
A single compromised session can result in bulk export of all PHI for a clinician's patient panel (potentially 50–200 patients) in under a minute. This is a HIPAA reportable breach if patients' records are exfiltrated.

## Fix
1. **Rate limiting:** Max 5 exports per user per hour via `@fastify/rate-limit`
2. **Re-authentication:** Require password confirmation for bulk export (like financial apps require for large transfers)
3. **Audit enrichment:** Log export volume (resource count) in the audit trail entry
4. **Notification:** Alert the exporting user via in-app notification with "You exported patient X's full record at [time]"
5. **Encryption:** Return FHIR bundle as AES-256-GCM encrypted payload with a download key

blamechris:
## Summary

## Test plan
- [x] Tests pass

 add per-user rate limit on FHIR exports (#234):Closes #234


blamechris:
## Review: PR #809 — fix(gateway): per-user rate limit on FHIR exports

**Verdict: Approve**

### Correctness
- Redis INCR + EXPIRE sliding window is correctly implemented. TTL is set only on `count === 1` (first request in window), preventing TTL resets on subsequent requests.
- Fire-and-forget audit with `.catch()` prevents audit I/O failures from blocking the 429 response. Good pattern.
- `retry-after` header sourced from `redis.ttl()` with a fallback to the full window — correct.

### TypeScript / ESM
- `.js` extensions used in all imports. Strict types throughout. `FhirExportAuditEvent` interface is well-typed.
- `crypto.randomUUID()` used for audit row IDs — consistent with project convention.

### Security
- Rate limit on FHIR export is an important data exfiltration control. 5/hour/user is a reasonable default for bulk patient exports.
- Audit trail on exceedance with IP, userId, and procedure name provides forensic value.
- The `onExceedance` injection point is clean for testing without needing real DB/Redis.

### Tests
- 261 lines of well-structured tests covering: within-limit, over-limit, per-user isolation, URL matching with query strings, TTL behavior, non-interference with other routes, and audit emission.
- Good edge case: unauthenticated request is a no-op (defensive guard).

### Nits (non-blocking)
- The `defaultAuditEmitter` constructs `new Date().toISOString()` inline. If the audit log table has a `DEFAULT now()`, this is redundant but harmless.
- Consider logging the `count` value in the audit event for trend analysis (how far over the limit was the user?).

LGTM.

PR #823, opened 2026-04-18
refactor(gateway): extract /auth/me handler for testability

### LINKED ISSUE #820: refactor(gateway): extract /auth/me handler for testability
PR #808's auth-me test re-implements the handler logic inline rather than importing from server.ts, because importing from server.ts requires bootstrapping Fastify. Consider extracting the /auth/me handler into a standalone function (e.g., in a handlers/ directory) that both server.ts and the test can import, preventing drift between test and production code.

Found during review of PR #808.

blamechris:
Closes #820

Extracts the `/auth/me` route handler from `server.ts` into a standalone
function in `services/api-gateway/src/handlers/auth-me.ts` so the unit
test and the Fastify route share a single code path, eliminating the
inline re-implementation that previously lived in the test file.

## Summary
- New `handleAuthMe(request, reply)` function exposed from `handlers/auth-me.ts` with a narrow `AuthMeRequest` / `AuthMeReply` surface — Fastify's typed `FastifyRequest` / `FastifyReply` are structurally compatible, so the route registration stays one line.
- `server.ts` now delegates the route body to the extracted handler.
- `__tests__/auth-me.test.ts` imports the handler directly (no Fastify bootstrap) and keeps the same three contract cases: 401 unauthenticated, profile echo, and `patient_id` for patient-role users.

## Test plan
- [x] `pnpm --filter @carebridge/api-gateway test` (256 passed, including 3 in `auth-me.test.ts`)
- [x] `pnpm typecheck`
- [x] `pnpm lint`


blamechris (commented):
## Review — refactor(gateway): extract /auth/me handler

**Verdict: LGTM (approve-equivalent; cannot self-approve)**

### Behavior preservation
The extracted `handleAuthMe` is byte-equivalent to the previous inline route body: same 401-on-missing-user branch, same seven-field profile payload (`id`, `email`, `name`, `role`, `specialty`, `department`, `patient_id`), same `return` semantics. `server.ts` now delegates via a one-line arrow that forwards `(request, reply)` straight through — no hidden wrapping, no accidental `preHandler` reshuffling.

### Auth gating intact
`server.addHook("preHandler", authMiddleware)` on server.ts:165 still runs globally before every route, so `/auth/me` remains gated exactly as before. The in-handler `!request.user` check is defensive and was already present inline.

### Testability win
The test now imports `handleAuthMe` directly and exercises the production code path — the previous duplicated `authMeHandler` inline in the test file (which was the whole reason #820 existed) is gone. The three contract cases (401, profile echo, patient_id propagation) are preserved and the fixtures are now typed as `User` from `@carebridge/shared-types`, which will catch future shape drift automatically.

### Type surface
Narrow `AuthMeRequest` / `AuthMeReply` interfaces are structurally compatible with Fastify's typed `FastifyRequest` / `FastifyReply` (the `code(...).send(...)` chain matches). No `as any`. Reusing `User["role"]` and `Pick<User, ...>` keeps the handler locked to the shared schema. CI typecheck passes.

### Other checks
- ESM `.js` extensions on both imports
- No unrelated changes (3 files, all directly related)
- Commit and PR body are clean
- CI green (Typecheck, Test, Build)

### Nits (non-blocking, filed as follow-up)
- The test's local `MockReply` interface duplicates the now-exported `AuthMeReply` shape. Could be reused to lock the mock to the handler contract.

Nice refactor — small, focused, eliminates the drift risk flagged in #820.

PR #825, opened 2026-04-18
fix(ai): use lab-provided H/L/critical flags and per-result reference ranges

### LINKED ISSUE #244: Missing reference ranges → critical lab values silently not flagged
**File:** services/ai-oversight/src/rules/critical-values.ts:97-111

If lab lacks reference_low/_high AND test not in COMMON_LAB_TESTS lookup, never flagged regardless of severity. 'Far out of range' check requires 2× → K+ 6.8 mEq/L may slip through.

**Fix:** Treat lab-provided H/L/critical flags as authoritative; prefer per-result reference_low/_high.

blamechris:
Closes #244

## Patient-safety summary

The critical-values deterministic rule in `services/ai-oversight/src/rules/critical-values.ts` could silently miss lab-flagged abnormal results. If a lab result was:

- not in `CRITICAL_LAB_THRESHOLDS` (no explicit rule), AND
- not in `COMMON_LAB_TESTS`, AND
- lacked `reference_low` / `reference_high`,

then even an analyzing laboratory's own `flag: "H" | "L" | "critical"` was ignored. The `"far out of range"` fallback also required a value be outside 2x the reference deviation, meaning a K+ 6.8 mEq/L reported with a 3.5–5.0 reference could slip past without an alert.

## Fix

In the heuristic-fallback path:

1. **Lab-provided flags are authoritative.** `flag: "critical"` emits a critical flag; `flag: "H"` / `flag: "L"` emit warning flags with the correct direction. The analyzing lab has more patient-specific context (baselines, institutional critical-value policy) than our static typical ranges.
2. **Per-result reference range beats `COMMON_LAB_TESTS`.** The shared typical range is only consulted when the inbound result does not carry its own `reference_low` / `reference_high`.
3. Existing severity stratification, explicit `CRITICAL_LAB_THRESHOLDS` (potassium, troponin, INR, etc.), and LOINC matching are unchanged. Warning flags use `WARNING-LAB-*` rule ids.

No schema changes were needed — `LabResult.flag` and the Zod validator already accept `"H" | "L" | "critical"`.

## Tests

Five new regression tests in `services/ai-oversight/src/__tests__/critical-values.test.ts`:

- Unknown lab with `flag: "critical"` and no reference range → critical flag.
- Unknown lab with `flag: "H"` → warning flag, high.
- Unknown lab with `flag: "L"` → warning flag, low.
- `flag: "critical"` honored even when value is within `COMMON_LAB_TESTS` typical range.
- Per-result `reference_low` / `reference_high` drive the far-outside check for labs not in `COMMON_LAB_TESTS`.
- Per-result range mid-value with no explicit flag → no flag (precedence over any mismatching `COMMON_LAB_TESTS` entry).

All 543 `@carebridge/ai-oversight` tests pass, plus `pnpm typecheck` and `pnpm lint` from repo root are clean.

### DIFF
--- a/services/ai-oversight/src/__tests__/critical-values.test.ts
+++ b/services/ai-oversight/src/__tests__/critical-values.test.ts
@@ -311,6 +311,119 @@ describe("checkCriticalValues — lab heuristic fallback", () => {
   });
 });
 
+// ─── Lab-provided flag precedence (issue #244) ───────────────────
+describe("checkCriticalValues — lab-provided flag precedence", () => {
+  it("flags unknown lab as critical when the lab reports flag='critical' and no reference range is supplied", () => {
+    // Regression: previously this silently slipped through because the lab
+    // was not in COMMON_LAB_TESTS and had no reference range. An analyzing
+    // laboratory explicitly flagging the value as critical is authoritative.
+    const flags = checkCriticalValues(
+      makeLabEvent([
+        {
+          test_name: "Obscure Marker X",
+          value: 999,
+          unit: "U/L",
+          flag: "critical",
+        },
+      ]),
+    );
+    expect(flags).toHaveLength(1);
+    expect(flags[0]!.severity).toBe("critical");
+    expect(flags[0]!.rule_id).toBe("CRITICAL-LAB-OBSCURE_MARKER_X");
+  });
+
+  it("emits warning flag when lab reports flag='H' without a reference range", () => {
+    const flags = checkCriticalValues(
+      makeLabEvent([
+        {
+          test_name: "Novel Biomarker",
+          value: 42,
+          unit: "pg/mL",
+          flag: "H",
+        },
+      ]),
+    );
+    expect(flags).toHaveLength(1);
+    expect(flags[0]!.severity).toBe("warning");
+    expect(flags[0]!.summary).toContain("High");
+  });
+
+  it("emits warning flag when lab reports flag='L' without a reference range", () => {
+    const flags = checkCriticalValues(
+      makeLabEvent([
+        {
+          test_name: "Novel Biomarker",
+          value: 0.1,
+          unit: "pg/mL",
+          flag: "L",
+        },
+      ]),
+    );
+    expect(flags).toHaveLength(1);
+    expect(flags[0]!.severity).toBe("warning");
+    expect(flags[0]!.summary).toContain("Low");
+  });
+
+  it("honors lab flag even when value is within COMMON_LAB_TESTS typical range", () => {
+    // A lab flagging a result as critical overrides typical-range fallback
+    // even when the value itself would otherwise look benign. The laboratory
+    // has more context (patient baseline, critical-value policy) than we do.
+    const flags = checkCriticalValues(
+      makeLabEvent([
+        {
+          test_name: "Hemoglobin",
+          value: 13.5, // within typical 12.0–17.5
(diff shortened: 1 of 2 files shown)


blamechris (commented):
## Inspector review — VERDICT: approve (self-authored, comment-only)

This is a tight, well-scoped patient-safety fix. The silent-drop scenario in #244 is addressed precisely: lab-provided `flag: "critical" | "H" | "L"` is now authoritative regardless of whether the test appears in `COMMON_LAB_TESTS` or carries a reference range, and per-result `reference_low/high` correctly take precedence over the shared typical range. Existing explicit thresholds (POTASSIUM >=6.0, Troponin I, INR, etc.) are unchanged because the `matchedExplicit` short-circuit still runs first — so no regression of existing deterministic coverage.

### What I verified
- `CRITICAL_LAB_THRESHOLDS` path (lines 469-486) is untouched; those thresholds still fire first and skip the heuristic fallback.
- New branches are exclusive (`if/else if`) with an `else if (low === undefined && high === undefined)` guard on the `COMMON_LAB_TESTS` fallback — per-result ranges win, matching the fix spec.
- `LabFlag` is `"H" | "L" | "critical"` (`packages/shared-types/src/clinical-data.ts:109`) and the Zod schema enforces the enum (`packages/validators/src/clinical-data.ts:86`), so the narrowing in the rule file is sound without `as any`.
- Rule ID namespace is distinguishable: `CRITICAL-LAB-*` for critical, `WARNING-LAB-*` for H/L. The flag-service dedup key is `(patient_id, rule_id, status='open')`, so warning->critical escalation on the same analyte emits a new flag (correct).
- Scope is disciplined: only `critical-values.ts` heuristic block + its test file changed. No other rules touched.
- Zero-attribution policy: clean. ESM `.js` extensions preserved.
- Six added tests cover all new branches plus the precedence edge cases. CI is green (build, typecheck, test).

### Nits (filed as follow-ups, not blocking)
1. Direction inference for `flag === "critical"` only consults `reference_low`; if a lab sends only `reference_high` with a below-range value, `direction` defaults to `"high"`. Low-impact because `summaryPrefix` is "Critical lab result" for severity=critical regardless of direction, but the `direction` value still flows into `suggested_action` wording for warnings and could matter for future consumers.
2. Unknown flag values (anything outside the `H|L|critical` enum) silently fall through to range checks. With the Zod schema in place upstream this should be unreachable, but a defensive debug log would make future FHIR-ingestion gaps easier to diagnose.
3. Broader gap (pre-existing, not a regression): a result with no `flag`, no `reference_low/high`, and not in `COMMON_LAB_TESTS` is still never flagged regardless of how extreme the value is. Worth considering a unit-aware sanity check or explicit "needs-human-review" flag for this case.
4. Rule-ID inconsistency (pre-existing): the `CRITICAL_LAB_THRESHOLDS` path emits `CRITICAL-LAB-*` rule_ids even for `severity: "warning"` (e.g., Troponin 0.05-0.4). This PR introduces `WARNING-LAB-*` only for the H/L fallback branch. Harmonizing the prefix-to-severity mapping across both paths would make downstream filtering/metrics cleaner.
5. Test gap: no regression test for an unrecognized `flag` string (e.g., `"abnormal"`, `"HH"`) to pin the fall-through behavior.

Follow-up issues filed separately; see cross-links. Patient-safety-critical fix, approving.

PR #839, opened 2026-04-18
refactor(portal): narrow User.role to UserRole literal union

### LINKED ISSUE #829: refactor(portal): narrow User.role to UserRole literal union from shared-types
## Context
Found during review of PR #824.

`packages/portal-shared/src/auth.tsx` declares:

```ts
export interface User {
  id: string;
  email: string;
  name: string;
  role: string;
  specialty?: string;
  department?: string;
  patient_id?: string;
}
```

`role: string` is wider than it needs to be. `packages/shared-types/src/auth.ts` already exports a literal union:

```ts
export type UserRole = "patient" | "nurse" | "physician" | "specialist" | "admin";
```

## Proposal
Switch `User.role` to `UserRole` so portal code gets exhaustiveness checking in `switch (user.role)` blocks and role-based UI guards can't silently accept a typo like `"physcian"`.

Optional follow-ups (lower priority, list here for context):
- `specialty?: Specialty` — would require defining an enum/union of accepted specialties. Currently free-form strings from the DB, so this may be infeasible without a catalog.
- `department?: Department` — same caveat.

## Scope
- `packages/portal-shared` must depend on `@carebridge/shared-types` (check workspace package graph).
- Touch points: any consumer that does `user.role === "..."` — should remain source-compatible since the literal union is narrower than `string`.

## Non-goals
Behavioral change, runtime validation. Pure typing refinement.

blamechris:
## Summary
- Switch `User.role` in `@carebridge/portal-shared`'s auth context from `string` to the `UserRole` literal union from `@carebridge/shared-types` (`"patient" | "nurse" | "physician" | "specialist" | "admin"`).
- Adds `@carebridge/shared-types` as a `workspace:*` dependency of `portal-shared`, matching the convention used by `validators`, `medical-logic`, `db-schema`, `ai-prompts`, and `outbox`.
- Adds a type-level test in `auth.test.tsx` that pins the narrowing via three independent probes:
  1. Direct `UserRole` assignability from `user.role`.
  2. Exhaustive `switch` over the union with a `never` fallthrough.
  3. `@ts-expect-error` on an out-of-union value (`"doctor"`).
  If the field ever re-widens to `string`, `tsc` fails; if a role is dropped, the exhaustive switch fails; if a bogus role is accepted, the expect-error probe fails.

## Rationale
Consumers of `useAuth()` previously received `user.role: string`, so role-based branching was silently tolerant of typos and never triggered exhaustiveness warnings. The canonical `UserRole` union already exists in `@carebridge/shared-types`; this widening was purely an integration gap.

## Scope
- In scope: `User.role` only.
- Out of scope (per issue #829 notes): `specialty?` and `department?` remain free-form `string` — those are harder to narrow cleanly and not urgent.

## Downstream consumer audit
Greppled `user.role` across the repo. All string comparisons (`=== "admin"`, `=== "patient"`, `=== "physician"`, `=== "specialist"`, `=== "nurse"`) are members of the union and continue to type-check. I did not find any portal-side comparison of `user.role` against an out-of-union literal (e.g. `"doctor"`). Gateway-side `(ctx.user.role as string) === "family_caregiver"` checks live on a different `User` type (server-side `shared-types` `User`, which already has `role: UserRole`) and are unaffected.

## Test plan
- [x] `pnpm typecheck` — 38/38 tasks pass
- [x] `pnpm lint` — 21/21 tasks pass (warnings pre-existing)
- [x] `pnpm --filter @carebridge/portal-shared test` — 6/6 passing (5 existing + 1 new type-pin test)
- [x] `pnpm --filter @carebridge/patient-portal test` — 9/9 passing
- [x] `pnpm --filter @carebridge/clinician-portal test` — 173/173 passing

Closes #829

### DIFF
--- a/apps/clinician-portal/app/login/page.tsx
+++ b/apps/clinician-portal/app/login/page.tsx
@@ -2,6 +2,7 @@
 
 import { useState } from "react";
 import { useRouter } from "next/navigation";
+import type { UserRole } from "@carebridge/shared-types";
 import { useAuth } from "@/lib/auth";
 import { trpcVanilla } from "@/lib/trpc";
 import { PasswordInput } from "@carebridge/portal-shared/password-input";
@@ -33,7 +34,7 @@ export default function LoginPage() {
         return;
       }
       const loginResult = result as {
-        user: { id: string; name: string; email: string; role: string };
+        user: { id: string; name: string; email: string; role: UserRole };
         session: { id: string };
       };
       setSession(loginResult.user, loginResult.session.id);
--- a/apps/patient-portal/app/login/page.tsx
+++ b/apps/patient-portal/app/login/page.tsx
@@ -2,6 +2,7 @@
 
 import { useState } from "react";
 import { useRouter } from "next/navigation";
+import type { UserRole } from "@carebridge/shared-types";
 import { useAuth } from "@/lib/auth";
 import { trpcVanilla } from "@/lib/trpc";
 import { PasswordInput } from "@carebridge/portal-shared/password-input";
@@ -33,7 +34,7 @@ export default function LoginPage() {
         return;
       }
       const loginResult = result as {
-        user: { id: string; name: string; email: string; role: string };
+        user: { id: string; name: string; email: string; role: UserRole };
         session: { id: string };
       };
       setSession(loginResult.user, loginResult.session.id);
--- a/packages/portal-shared/package.json
+++ b/packages/portal-shared/package.json
@@ -11,6 +11,7 @@
     "./password-input": "./src/password-input.tsx"
   },
   "dependencies": {
+    "@carebridge/shared-types": "workspace:*",
     "@trpc/client": "^11.0.0",
     "@trpc/react-query": "^11.0.0",
     "@trpc/server": "^11.0.0",
--- a/packages/portal-shared/src/auth.test.tsx
+++ b/packages/portal-shared/src/auth.test.tsx
@@ -3,6 +3,7 @@ import { renderHook, act, waitFor } from "@testing-library/react";
 import { QueryClient } from "@tanstack/react-query";
 import { AuthProvider, useAuth } from "./auth";
 import type { User } from "./auth";
+import type { UserRole } from "@carebridge/shared-types";
 import type { ReactNode } from "react";
 
 // ---------------------------------------------------------------------------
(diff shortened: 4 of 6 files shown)


blamechris:
## /agent-review — VERDICT: REQUEST CHANGES (comment mode, self-authored)

The narrowing itself and the merge resolution are clean. Wave 1 (#824) fields (`specialty`, `department`, `patient_id`) and the `verifying` state are preserved, the Wave 2 type-level pin is well-crafted (direct assignability + exhaustive switch + `@ts-expect-error` on `"doctor"`), and there are no `as any` escapes. But there is a hard blocker:

### Blocker 1: Build fails on both portals — login pages hand `role: string` to `setSession`

CI Build job is red on this PR head (run 24611219474). The failure:

```
./app/login/page.tsx:39:18
Type error: Argument of type '{ id: string; name: string; email: string; role: string; }'
  is not assignable to parameter of type 'User'.
  Types of property 'role' are incompatible.
    Type 'string' is not assignable to type 'UserRole'.
```

Both `apps/patient-portal/app/login/page.tsx` and `apps/clinician-portal/app/login/page.tsx` do:

```ts
const loginResult = result as {
  user: { id: string; name: string; email: string; role: string };
  session: { id: string };
};
setSession(loginResult.user, loginResult.session.id);
```

That explicit `role: string` on the local cast was compatible when `User.role: string` but is now a hard compile error against `User.role: UserRole`. The MFA path (`setSession(result.user, result.session.id)` at line 61 of both files) should also be re-checked after the fix — if tRPC's inferred return type still carries `role: string`, it will fail on the same rule.

The PR description says `pnpm typecheck` (38/38) and `pnpm lint` pass locally, but `pnpm build` invokes `next build` which runs its own stricter type gate — that is what the CI Build job catches and what is red here. The downstream audit in the PR body missed the two `login/page.tsx` callsites.

**Fix options:**
1. Drop the `role: string` from the local type annotation and let tRPC inference carry `UserRole` (requires the server-side return type to already be `UserRole`; if not, option 2).
2. Change to `role: UserRole` and import the type.
3. Cast at the call site: `setSession({ ...loginResult.user, role: loginResult.user.role as UserRole }, ...)` — least clean, avoid unless (1) and (2) are infeasible.

Then rerun CI.

### Everything else looks good (praise)

- `User` post-merge has all 7 fields (id, email, name, `role: UserRole`, specialty?, department?, patient_id?) with informative JSDoc on each.
- The type-level pin test is one of the better ones in this repo: three independent probes (direct assignment, exhaustive `never` fallthrough, `@ts-expect-error` typo guard) mean any of widening / narrowing / typo-tolerance regression breaks `tsc`.
- Both Wave 1 (specialty/department/patient_id regression tests) and Wave 2 (role pin) suites are present and coherent in the merged `auth.test.tsx`.
- `@carebridge/shared-types` added to `packages/portal-shared/package.json` with `workspace:*`, matching convention.
- No `as any`, no `@ts-ignore`, no silencing.
- Zero AI attribution anywhere (commit messages, PR body, merge commit).
- Server-side `user.role === "..."` comparisons in `services/*/src/**` use a separate `User` from `@carebridge/shared-types` directly and are unaffected. Portal-side `user?.role` reads (sidebar, settings page) are display-only and remain source-compatible.
- ESM `.js` extension not applicable here — these are `.tsx` type-only imports from a workspace package.

### Nit (filed as follow-up)
The `(ctx.user.role as string) === "family_caregiver"` pattern in `services/api-gateway/src/routers/patient-records.ts` (4 occurrences, lines 308/321/358/371) is a smell unrelated to this PR: a role exists in code but not in the union. Either add `"family_caregiver"` to `UserRole` or remove the dead branch.

Issue #836, opened 2026-04-18
refactor(ai): harmonize CRITICAL-LAB-* vs WARNING-LAB-* rule_id prefixes across lab rules


blamechris:
Found during review of PR #825.

Inconsistency observed in `services/ai-oversight/src/rules/critical-values.ts`:

- Explicit threshold path (lines 469-486) emits `rule_id: CRITICAL-LAB-${ruleKey}` even when `threshold.severity === \"warning\"` (e.g., Troponin I between 0.04 and 0.4 emits `CRITICAL-LAB-TROPONIN_I` with severity=warning; Potassium 5.1-5.9 emits `CRITICAL-LAB-POTASSIUM` with severity=warning).
- New heuristic-fallback H/L path (introduced in PR #825) emits `rule_id: WARNING-LAB-${test_name}` for severity=warning.

Downstream consumers that filter by `rule_id` prefix to infer severity will see inconsistent semantics depending on which path fired. The existing dedup key is `(patient_id, rule_id, status='open')`, so the two paths also dedup independently even for the same analyte + severity.

**Proposed:** harmonize on a single scheme — either:
- `CRITICAL-LAB-*` is a namespace (severity-independent) and downstream must read the `severity` field, OR
- rule_id prefix reflects severity everywhere: `CRITICAL-LAB-*` for critical, `WARNING-LAB-*` for warning in both paths.

Option 2 matches the direction PR #825 took for the new code; consider a small sweep to bring the explicit-threshold path in line, or document the convention in the rule file header.

**Not clinical-safety-critical** (severity is always carried on the flag record), but cleans up a footgun for anyone building dashboards / metrics keyed on rule_id.

Pre-existing; not blocking PR #825.

PR #842, opened 2026-04-18
fix(ai): lab-flag direction inference + unrecognized-flag handling (#833, #834, #837)

### LINKED ISSUE #833: fix(ai): direction inference for critical lab flag ignores reference_high-only case
Found during review of PR #825 (issue #244 fix).

In `services/ai-oversight/src/rules/critical-values.ts`, the heuristic fallback infers `direction` for `flag: \"critical\"` using only `reference_low`:

\`\`\`ts
direction =
  result.reference_low !== undefined &&
  result.value < result.reference_low
    ? \"low\"
    : \"high\";
\`\`\`

If a lab reports a value below `reference_high` and omits `reference_low` entirely (some reference sets only bound one side), direction defaults to `\"high\"` even when the value is clearly low. Impact is muted for severity=critical because the user-facing summary is \"Critical lab result\" regardless, but `direction` still influences `suggested_action` wording when extended to warnings in the future, and any downstream consumer reading the direction field would be misled.

**Proposed fix:** also check `reference_high` when `reference_low` is undefined:

\`\`\`ts
direction =
  result.reference_low !== undefined && result.value < result.reference_low
    ? \"low\"
    : result.reference_high !== undefined && result.value > result.reference_high
      ? \"high\"
      : \"high\"; // default
\`\`\`

Or infer from the midpoint when both are present.

**Why clinical-safety:** misdirected alerts (e.g., telling a clinician a panic-low magnesium is \"high\") erode trust in the deterministic layer and may delay appropriate treatment.

### LINKED ISSUE #834: feat(ai): log and optionally flag unrecognized lab-result flag values
Found during review of PR #825 (issue #244 fix).

The new heuristic branches in `critical-values.ts` match exactly on `flag === \"critical\" | \"H\" | \"L\"`. Any other string (e.g., `\"HH\"` panic-high, `\"LL\"` panic-low, `\"abnormal\"`, `\"A\"`, empty string, whitespace) falls through silently to the range checks. Upstream the Zod validator (`packages/validators/src/clinical-data.ts:86`) enforces the enum, so intra-app writes can't produce these — but results ingested from FHIR, HL7v2, or any path that bypasses the Zod validator could.

**Proposed:**
1. Emit a `logger.warn` (or observability metric) when a non-null `flag` is present but not in the recognized enum. Silent drops of an explicit lab abnormality signal are the exact failure mode #244 was filed to prevent.
2. Consider treating `\"HH\"` / `\"LL\"` / `\"panic\"` / `\"A\"` (HL7 common values) as at-least-warning pending a principled mapping.
3. Add a test case in `critical-values.test.ts` pinning the current fall-through behavior so any future change is intentional.

**Why clinical-safety:** protects against the same class of silent-drop bug that #244 addressed, at the ingestion boundary rather than the rule.

### LINKED ISSUE #837: test(ai): pin fall-through behavior for unrecognized lab flag values
Found during review of PR #825.

Test coverage gap: the new branches in `critical-values.ts` key on `flag === \"critical\" | \"H\" | \"L\"`, but there is no test that pins what happens when a lab result carries a non-enum flag value (e.g., `\"HH\"`, `\"LL\"`, `\"abnormal\"`, empty string). Current behavior is silent fall-through to the range checks. That's intentional, but without a test it's unprotected against regressions.

**Suggested addition** to `services/ai-oversight/src/__tests__/critical-values.test.ts`:

\`\`\`ts
it(\"falls through to range checks when flag is an unrecognized value\", () => {
  const flags = checkCriticalValues(
    makeLabEvent([
      { test_name: \"Obscure Marker X\", value: 999, unit: \"U/L\", flag: \"HH\" },
    ]),
  );
  // No reference range, not in COMMON_LAB_TESTS, unrecognized flag → no flag emitted.
  expect(flags).toHaveLength(0);
});
\`\`\`

See also the related clinical-safety issue about logging unrecognized flag values.

blamechris:
## Summary

Bundled fix for three clinical-safety follow-ups from PR #825 (the #244 fix).
All three issues touch `services/ai-oversight/src/rules/critical-values.ts`,
so they are bundled for coherent review. Each issue is still testable and
diffable in isolation within the change.

- **#833** — half-bounded reference-range direction inference bug: a critical
  lab with only `reference_high` (no `reference_low`) and a value clearly
  below `reference_high` was mis-tagged as `direction=high`. Misdirected
  panic-low flags erode clinician trust and could delay treatment.
- **#834** — silent drop of non-enum flag values (`"HH"`, `"LL"`,
  `"abnormal"`, `"A"`, empty) at the ingress boundary. Intra-app writes are
  Zod-guarded, but FHIR / HL7v2 / direct DB seeding paths bypass the
  validator.
- **#837** — test coverage gap for the unrecognized-flag fall-through.

### `services/ai-oversight/src/rules/critical-values.ts`
1. New `inferCriticalDirection(value, referenceLow, referenceHigh)` helper:
   - Both bounds present → direction = midpoint comparison
   - Only `reference_high` present → direction = comparison with `reference_high`
   - Only `reference_low` present → direction = comparison with `reference_low`
   - Neither present → `"high"` (documented fallback)
2. Direction now appears in the critical-severity summary
   (`"Critical low lab result"` vs `"Critical high lab result"`) so the
   #833 fix is observable to clinicians in the rendered flag.
3. Added `logger.warn("unrecognized_lab_flag", { flag, test_name,
   test_code })` whenever a non-null flag falls outside the validator
   enum. Metric-shaped payload; no PHI.
4. Added conservative HL7v2 mapping:
   - `"HH"` → warning, direction `"high"`
   - `"LL"` → warning, direction `"low"`

   **HL7 mapping decision:** I included `HH` / `LL` because their
   semantics are unambiguous in HL7v2 ("panic high" / "panic low"). I
   intentionally did **not** map `"A"` (abnormal-unspecified) or
   `"panic"` because their severity / direction mapping is ambiguous in
   clinical terms — the `logger.warn` is the signal for operators, and
   a future issue can propose a principled mapping. Chosen severity is
   `warning` (not `critical`) as a conservative floor; callers who want
   `critical` severity should normalize to the canonical `"critical"`
   enum value upstream.

### `services/ai-oversight/src/__tests__/critical-values.test.ts`
- Mocked `@carebridge/logger` so `mockWarn` can be asserted against.
- Added `describe("... direction inference (issue #833)")` with three
  cases: half-bounded low, half-bounded high, and no-bounds default.
- Added `describe("... unrecognized flag handling (issues #834, #837)")`
  with the suggested named test from #837 plus tests for: empty-string
  flag, `"abnormal"` flag, no-warn for recognized/absent flags, and the
  new `HH` / `LL` mappings.

## Clinical safety rigor

- `CRITICAL_LAB_THRESHOLDS` and every hard-coded numeric threshold are
  unchanged.
- `rule_id` naming is unchanged (issue #836 is a separate cleanup).
- All PR #825 / issue #244 regression tests pass unchanged.
- Full ai-oversight test suite (552 tests) green.
- `pnpm typecheck` and `pnpm lint` green.

## Test plan

- [x] `pnpm --filter @carebridge/ai-oversight test` — 552 passed
- [x] `pnpm lint` — green (only pre-existing portal warnings)
- [x] #244 regression-guard tests still pass unchanged
- [ ] Reviewer to confirm HL7 mapping decision (HH / LL only; A, panic deferred)
- [ ] Reviewer to confirm direction-in-summary wording is acceptable for clinician UI

Closes #833
Closes #834
Closes #837

### DIFF
--- a/services/ai-oversight/src/__tests__/critical-values.test.ts
+++ b/services/ai-oversight/src/__tests__/critical-values.test.ts
@@ -1,4 +1,16 @@
-import { describe, it, expect, vi } from "vitest";
+import { describe, it, expect, vi, beforeEach } from "vitest";
+
+// Mock @carebridge/logger so we can assert on logger.warn() calls when an
+// unrecognized lab flag is encountered (issue #834).
+const { mockWarn } = vi.hoisted(() => ({ mockWarn: vi.fn() }));
+vi.mock("@carebridge/logger", () => ({
+  createLogger: () => ({
+    debug: vi.fn(),
+    info: vi.fn(),
+    warn: mockWarn,
+    error: vi.fn(),
+  }),
+}));
 
 // Mock workspace dependencies before importing the module under test.
 vi.mock("@carebridge/shared-types", () => ({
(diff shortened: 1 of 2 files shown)


blamechris (commented):
## /agent-review — PR #842 (patient-safety rigor)

**Verdict: approve with 3 NITs filed for follow-up** (posting as comment — cannot self-approve)

### Criteria pass/fail

| # | Criterion | Result |
|---|-----------|--------|
| 1 | Zero Attribution | PASS — no Claude/Anthropic/"Generated with" in diff |
| 2 | Wave 1 #244 regression | PASS — existing assertions use `summary.toContain("High"/"Low")`, the new `"Critical high lab result"` / `"Critical low lab result"` prefixes still satisfy them |
| 3a | `reference_high`-only, value < `reference_high` → direction "low" | PASS (new `inferCriticalDirection`) |
| 3b | `reference_low`-only, value > `reference_low` → direction "high" | PASS |
| 3c | Both present → midpoint | PASS (midpoint is a principled neutral choice for in-range panic flags) |
| 3d | Neither present → default | PASS — `"high"` (documented fallback, matches prior behavior; pinned by new test) |
| 4 | HL7 HH/LL mapping choice | See NIT #853 — filed as clinical-safety discussion: HH/LL are panic flags; mapping to `warning` may under-alert vs the lab's own severity stratification. Filed for review, not blocking. |
| 5 | Logger PHI-safety | PASS — payload is `{metric, flag, test_name, test_code}`; no `value`, no `patient_id`, no `reference_low/high` |
| 6 | Unrecognized-flag fall-through test (#837) | PASS — `describe("... unrecognized flag handling")` block includes the suggested named test plus empty-string, `"abnormal"`, no-warn-for-recognized, and HH/LL positive cases |
| 7 | `CRITICAL_LAB_THRESHOLDS` untouched | PASS — diff shows no numeric threshold changes |
| 8 | `rule_id` naming preserved | PASS (issue #836 remains out-of-scope) |
| 9 | TypeScript strict, no `as any` | PASS — no `as any`, no `@ts-ignore`, no `@ts-expect-error`, no `eslint-disable` in diff |

### Key finding — not blocking, but filed as NIT

The `HH` / `LL` mapping sets `severity = "warning"` in the flag branch, but the subsequent numeric range-check block is gated on `if (severity === null)`. So once HH/LL assigns warning severity, the far-outside-range and COMMON_LAB_TESTS checks are **skipped** — making the mapping a ceiling, not the intended floor. Concrete bad case: `{flag:"LL", value:0.1, reference_low:3.5, reference_high:5.0}` would fire warning (via LL) instead of critical (via numeric check). Copilot flagged the same. Filed as **#849** (clinical-safety).

Also filed:
- **#851** — log-event-name vs `metric`-field naming divergence (matches existing `timestamp_fallback_total` pattern).
- **#853** — clinical-safety discussion on whether HH/LL should map to `critical` rather than `warning` (HL7v2 `HH`/`LL` are panic flags; single-letter `H`/`L` are simply out-of-range).

### Praise

- The new `inferCriticalDirection` helper is well-documented, the midpoint rule is a principled choice for in-range panic flags, and the fix is observable to clinicians via the new `"Critical high/low lab result"` summary prefix — good instinct to surface direction in the user-facing string so silent direction-bugs can't hide again.
- Test coverage is thorough: regression cases for all three bounds-combinations plus the unrecognized-flag pin requested in #837, plus positive HH/LL coverage, plus negative-coverage for recognized enum values.
- PR description explicitly calls out the HL7 severity-floor decision and defers `A` / `panic` mapping — the kind of intentional conservatism patient-safety code benefits from.

### Files

- `services/ai-oversight/src/rules/critical-values.ts`
- `services/ai-oversight/src/__tests__/critical-values.test.ts`

PR #845, opened 2026-04-18
test(services): migrate remaining DB mock chains to @carebridge/test-utils

### LINKED ISSUE #830: test(services): migrate remaining mockReturnValueOnce DB chains to @carebridge/test-utils
Follow-up from review of #826.

PR #826 introduces `@carebridge/test-utils` with `createMockDb()` and migrates the `patient-records` and `clinical-notes` service tests. The subagent explicitly left the remaining consumers for a follow-up. This issue tracks that migration.

## Files still using hand-rolled `mockReturnValueOnce` / nested `vi.fn(() => ({...}))` DB chain stubs

Identified via `grep` for the old patterns:

- `services/clinical-data/src/__tests__/lab-repo.test.ts`
- `services/clinical-data/src/__tests__/vital-repo.test.ts`
- `services/clinical-data/src/__tests__/medication-repo.test.ts`
- `services/notifications/src/__tests__/dispatch-worker.test.ts` (if it uses DB chain stubs)
- `services/api-gateway/src/middleware/rbac.test.ts` (only if it stubs DB; otherwise skip)
- `services/auth/src/__tests__/session-cleanup.test.ts`
- `services/auth/src/__tests__/mfa-enumeration.test.ts`
- `services/ai-oversight/src/__tests__/llm-timeout-fallback.test.ts`
- `services/ai-oversight/src/__tests__/claude-client-logging.test.ts`

## Migration checklist per file

1. Add `@carebridge/test-utils` to the service's `devDependencies` (workspace:*) if not already.
2. Add the vitest alias in `services/<svc>/vitest.config.ts`.
3. Replace per-method nested `vi.fn()` chains with a `let db: MockDb; beforeEach(() => { db = createMockDb(); });` pattern.
4. Replace `xMock.mockResolvedValueOnce(rows)` with `db.willSelect(rows)` / `willInsert()` / `willUpdate(rows)` / `willDelete()`.
5. Replace chain-method call assertions with `db.select.calls[0]?.chain.includes("where")` etc.
6. Confirm assertion count is preserved (no silently dropped coverage).
7. Run `pnpm --filter @carebridge/<svc> test` — all should still pass.

## Scope control

- Add new chain verbs to `packages/test-utils/src/mock-db.ts` `chainNames` whitelist only as needed.
- Keep migrations per-PR or batched by service; do not mix with unrelated test changes.

blamechris:
## Summary

Follow-up to PR #826, which introduced `@carebridge/test-utils` with `createMockDb()` and migrated `patient-records` and `clinical-notes`. This migrates the rest of the consumers identified in #830 to the shared fluent helper.

### Migrated

- `services/clinical-data/src/__tests__/vital-repo.test.ts`
- `services/clinical-data/src/__tests__/medication-repo.test.ts`
- `services/clinical-data/src/__tests__/lab-repo.test.ts` — keeps a standalone `transactionMock` since `db.transaction(cb)` is outside the helper's Drizzle-chain API; select chains flow through `willSelect`.
- `services/notifications/src/__tests__/dispatch-worker.test.ts` — care-team + users + per-user preferences selects flow through `willSelect`; `insert(...).values(...)` args read from `db.insert.calls[i].chainArgs`.
- `services/auth/src/__tests__/session-cleanup.test.ts` — `delete().where().returning()` and `insert().values()` chains migrated; the `where(...)` argument inspection (OR/AND drizzle shape) now reads from `chainArgs`.
- `services/auth/src/__tests__/mfa-enumeration.test.ts` — user-row lookup migrated to `willSelect`; the trivial insert/update/delete stubs the helper provides replace the one-liner `vi.fn()` stubs.

Each service now declares `@carebridge/test-utils` as a `workspace:*` devDependency and adds a matching vitest alias.

### Holdouts (follow-up issues recommended)

- `services/api-gateway/src/middleware/rbac.test.ts` — uses dual-table mock routing (`careTeamSelectMock` / `emergencySelectMock`) with per-table call-count assertions. The shared helper's single select queue would lose that distinction and the test would become less readable, not more.
- `services/ai-oversight/src/__tests__/llm-timeout-fallback.test.ts` — uses `db.query.patients.findFirst` (a prepared-query surface the helper does not model) and asserts on `update().set(...)` args by chain index; migrating would require verbose accessors without a clarity win.
- `services/ai-oversight/src/__tests__/claude-client-logging.test.ts` — does not stub DB chains (only mocks `@anthropic-ai/sdk`), skipped per the issue scope ("only if it stubs DB").

### Scope discipline

- No production code touched.
- Helper's `chainNames` whitelist was sufficient as-is — no additions needed.
- Assertion counts preserved exactly:
  - `@carebridge/clinical-data` — 47 tests (matches baseline)
  - `@carebridge/notifications` — 63 tests (matches baseline)
  - `@carebridge/auth` — 97 tests (matches baseline)

## Test plan

- [x] `pnpm --filter @carebridge/clinical-data test` — 47 tests pass
- [x] `pnpm --filter @carebridge/notifications test` — 63 tests pass
- [x] `pnpm --filter @carebridge/auth test` — 97 tests pass
- [x] `pnpm lint` — clean (only preexisting Next.js warnings unchanged)

Closes #830

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -530,6 +530,9 @@ importers:
         specifier: ^3.24.0
         version: 3.25.76
     devDependencies:
+      '@carebridge/test-utils':
+        specifier: workspace:*
+        version: link:../../packages/test-utils
       '@types/node':
         specifier: ^22.0.0
         version: 22.19.17
@@ -576,6 +579,9 @@ importers:
         specifier: ^3.24.0
         version: 3.25.76
     devDependencies:
+      '@carebridge/test-utils':
+        specifier: workspace:*
+        version: link:../../packages/test-utils
       '@types/node':
         specifier: ^22.0.0
         version: 22.19.17
@@ -736,6 +742,9 @@ importers:
         specifier: ^3.24.0
         version: 3.25.76
     devDependencies:
+      '@carebridge/test-utils':
+        specifier: workspace:*
+        version: link:../../packages/test-utils
       '@types/node':
         specifier: ^22.0.0
         version: 22.19.17
--- a/services/auth/package.json
+++ b/services/auth/package.json
@@ -36,6 +36,7 @@
     "zod": "^3.24.0"
   },
   "devDependencies": {
+    "@carebridge/test-utils": "workspace:*",
     "@types/node": "^22.0.0",
     "@types/qrcode": "^1.5.6",
     "typescript": "^5.7.0",
--- a/services/auth/src/__tests__/mfa-enumeration.test.ts
+++ b/services/auth/src/__tests__/mfa-enumeration.test.ts
@@ -1,5 +1,6 @@
 import { describe, it, expect, beforeEach, vi } from "vitest";
 import { TRPCError } from "@trpc/server";
+import { createMockDb, type MockDb } from "@carebridge/test-utils";
 
 // ---------------------------------------------------------------------------
 // Mocks — set up before importing the module under test
@@ -75,23 +76,10 @@ vi.mock("@carebridge/redis-config", () => ({
   }),
 }));
 
-// Controllable user-row result for the router's db lookup.
-let mockUserRows: unknown[] = [];
-
-const mockLimit = vi.fn(async () => mockUserRows);
-const mockWhere = vi.fn(() => ({ limit: mockLimit }));
-const mockFrom = vi.fn(() => ({ where: mockWhere }));
-const mockSelect = vi.fn(() => ({ from: mockFrom }));
+let db: MockDb;
 
 vi.mock("@carebridge/db-schema", () => ({
-  getDb: () => ({
-    select: mockSelect,
-    insert: vi.fn(() => ({ values: vi.fn(async () => undefined) })),
-    update: vi.fn(() => ({
-      set: vi.fn(() => ({ where: vi.fn(async () => undefined) })),
-    })),
-    delete: vi.fn(() => ({ where: vi.fn(async () => undefined) })),
-  }),
+  getDb: () => db,
   users: {
     id: "users.id",
     email: "users.email",
@@ -154,8 +142,8 @@ function createCaller() {
 describe("mfaCompleteLogin — user enumeration hardening (issue #278)", () => {
   beforeEach(() => {
     redisStore.clear();
-    mockUserRows = [];
     vi.clearAllMocks();
+    db = createMockDb();
   });
 
   async function callAndCaptureError(): Promise<TRPCError> {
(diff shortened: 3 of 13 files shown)


blamechris (commented):
## Review (self-review, would be approve otherwise)

Clean, disciplined Wave 2 migration. Zero Attribution clean. Coverage preserved exactly (clinical-data 47 / notifications 63 / auth 97 verified locally). Holdouts (rbac, llm-timeout-fallback, claude-client-logging) are well-reasoned. No `as any` introduced in migrated files. No helper changes — `packages/test-utils/src/mock-db.ts` untouched, as claimed. Workspace wiring and `pnpm-lock.yaml` updates consistent across all three services.

### Praise

- **`dispatch-worker.test.ts` latent-bug catch is the right move.** The old hand-rolled mock reused `mockSelectResult2` (the **users list** shape) for every select after the first, so the 3rd per-recipient `notification_preferences` select was silently receiving provider-shaped rows. Tests still passed because the worker happened not to read a distinguishing field — but the test was never actually exercising the preferences path it claimed to cover. The new `primeDispatch(assignments, providers)` helper queues exactly `1 + 1 + N` selects, so each recipient's preferences lookup now gets a real (empty → opt-out default) result. This is a **test-fidelity improvement**, not a regression.

- `lab-repo.test.ts` correctly keeps a standalone `transactionMock` rather than forcing `db.transaction(cb)` through the helper — the right call, since the helper models chains, not callback APIs.

- `session-cleanup.test.ts` extracts a small `getDeleteWhereArg()` helper that makes the OR/AND-condition drilling genuinely more readable than the old `mockWhere.mock.calls[0][0]` pattern.

### NITs (non-blocking)

1. **Surface the `primeDispatch` fix in the PR body.** The behavioural fix (3rd select now gets correct `notification_preferences` shape rather than accidental users-list shape) is buried inside a single-line parenthetical in the Migrated section. A reviewer skimming the PR body would miss it. Suggest adding an explicit bullet under a new "### Incidental fixes" subheading, e.g.:

   > The old hand-rolled mock reused the users-list shape for every select after the first, papering over a missing per-recipient `notification_preferences` select. The `primeDispatch()` helper now queues all three selects in the correct shape; no production behaviour changed, but the test now actually exercises the preferences opt-out default path.

2. **Holdouts now have issues.** Filed two follow-up issues so the rbac and llm-timeout-fallback holdouts do not drift:
   - #850 — `rbac.test.ts` dual-table routing
   - #852 — `llm-timeout-fallback.test.ts` `db.query.<table>.findFirst` + index-based `set()` assertions

   The `claude-client-logging.test.ts` holdout is correctly out of scope (no DB stubs) per #830's explicit scope guard.

3. (Style, entirely optional) `session-cleanup.test.ts` line 251 keeps the `as { op; args }` cast for one branch but the `getDeleteWhereArg()` helper already narrows to that shape — the cast could be dropped. Not worth a follow-up.

### Verification summary

- `pnpm --filter @carebridge/clinical-data test` → 47 pass
- `pnpm --filter @carebridge/notifications test` → 63 pass
- `pnpm --filter @carebridge/auth test` → 97 pass
- Diff confirms `packages/test-utils/src/mock-db.ts` not modified (helper-extension claim holds)
- Diff grep for `Claude | Anthropic | Co-Authored | Generated with` — no matches (Zero Attribution clean)
- Diff grep for `as any` in migrated files — no matches

PR #844, opened 2026-04-18
feat(portal): add role=alert to errors and aria-label to interactive elements

blamechris:
## Summary
WCAG 2.1 AA a11y fixes for the clinician- and patient-portal surfaces called out in issue #183.

- **3.3.1 Error Identification** — both login forms' error container is now `role="alert"` + `aria-live="assertive"`, so assistive tech announces failed credentials / MFA errors the moment they render.
- **4.1.2 Name, Role, Value** — the per-flag Acknowledge / Resolve / Dismiss buttons in the AI inbox get an `aria-label` that includes the flag summary, giving each a unique accessible name.
- **2.4.4 / 4.1.2** — the patients-list search input exposes `aria-label="Search by patient name or MRN"` (placeholders are not an accessible name under the ARIA accessible-name algorithm).
- **2.4.3 Focus Order** — the patient-chart tab strip sets `aria-current="true"` on the active tab (string value, not boolean) and omits the attribute on inactive tabs, so screen readers can identify the current section.

Closes #183.

## WCAG criteria addressed
- [x] 2.4.3 Focus Order — active tab has `aria-current="true"`; inactive tabs omit the attribute
- [x] 2.4.4 Link Purpose (In Context) — search input has a real accessible name
- [x] 3.3.1 Error Identification — login error containers are live regions
- [x] 4.1.2 Name, Role, Value — inbox action buttons have unique accessible names tied to their flag

## Test plan
- [x] `pnpm --filter @carebridge/clinician-portal test` — 186/186 pass, including 10 new a11y tests
- [x] `pnpm --filter @carebridge/patient-portal test` — 13/13 pass, including 1 new a11y test
- [x] `pnpm typecheck` — green
- [x] `pnpm lint` — green (only pre-existing warnings)

### New render-based tests (TDD red → green)
- `login-error-a11y.test.tsx` (clinician + patient): mount the login page, reject the login mutation, assert `role="alert"` and `aria-live="assertive"` on the error container.
- `inbox-button-a11y.test.tsx`: render the inbox with mocked flag data, assert every Acknowledge / Resolve / Dismiss button has the expected `aria-label` referencing the flag summary.
- `patients-search-a11y.test.tsx`: assert the search input is labelled "Search by patient name or MRN".
- `patient-chart-tabs-a11y.test.tsx`: render the patient chart, assert only the active tab has `aria-current="true"` (and inactive tabs omit the attribute, not `"false"`).

### Harness notes
[...]

PR #859, opened 2026-04-18
refactor(portal): extract createPermissiveTrpcMock for a11y render tests

### LINKED ISSUE #848: nit(portal): extract tRPC Proxy mock helper for a11y render tests
Follow-up from review of PR #844.

The `patient-chart-tabs-a11y.test.tsx` uses a Proxy-based deep-getter to stub arbitrary `trpc.<router>.<proc>.useQuery/useMutation` access paths:

```ts
const proxy = (path: string[]): unknown =>
  new Proxy({}, {
    get(_t, prop: string) {
      if (prop === "then") return undefined;
      if (prop === "useUtils") return () => utilsProxy();
      ...
    },
  });
```

This pattern is useful whenever a rendered page touches many tRPC procs (patient chart, inbox, etc.), but the mock logic is inline and likely to be re-implemented by hand in future a11y / render tests. Extract it to `apps/clinician-portal/src/__tests__/helpers/trpc-mock.ts` (or a shared `@carebridge/test-utils` package) with a typed factory:

```ts
export function createPermissiveTrpcMock(overrides?: Partial<TrpcOverrides>) { ... }
```

Nice-to-have, not a blocker. Original context: PR #844 closes #183 (WCAG 2.1 AA ARIA affordances).

blamechris:
## Summary

Extracts the inline Proxy-based tRPC stub from
`apps/clinician-portal/src/__tests__/patient-chart-tabs-a11y.test.tsx`
(introduced in Wave 2 PR #844) into a reusable helper at
`apps/clinician-portal/src/__tests__/helpers/trpc-mock.ts`.

The helper exports `createPermissiveTrpcMock({ overrides })` — a
typed factory that returns a deep-proxy answering any
`trpc.<router>.<...>.<proc>.useQuery(...)` or `.useMutation()` call
with a loading-but-quiet stub, plus a no-op `trpc.useUtils()` whose
`.invalidate()` always resolves. Callers pass per-route overrides
keyed by dot-separated procedure path (e.g. `patients.getById`)
when they need a real-looking value for a single query.

## Location decision — Option 1 (portal-local)

Task offered two locations:

- Option 1: `apps/clinician-portal/src/__tests__/helpers/trpc-mock.ts`
- Option 2: `packages/test-utils/src/trpc-mock.ts` (shared)

Chose Option 1. Grepping `new Proxy(` across `apps/clinician-portal`
and `apps/patient-portal` test files shows only one call site today
(the one being refactored). Patient-portal has no analogous inline
Proxy stubs, so promoting to a shared package would be speculative
scope. If a second consumer appears later, relocation is a trivial
move.

## Scope discipline

Only the one Wave 2 test using the permissive deep-proxy pattern was
migrated. The other Wave 2 a11y tests use narrower, adequate mocks
and are intentionally left alone:

- `login-error-a11y.test.tsx` — mocks `trpcVanilla.auth.login.mutate`
  directly (simple reject-on-demand).
- `inbox-button-a11y.test.tsx` — explicit `clinicalFlags.*` query
  and mutation shape only.
- `patients-search-a11y.test.tsx` — one-line `patients.list.useQuery`
  stub.

## Implementation notes

- `vi.mock` factories are hoisted above top-level imports, so the
  factory uses dynamic `await import("./helpers/trpc-mock")` and
  the per-test `patient` fixture is wrapped in `vi.hoisted(...)`.
- Helper API intentionally minimal: `defaultQuery`, `defaultMutation`,
  `overrides`. No speculative expansion.

## Test plan

- [x] `pnpm --filter @carebridge/clinician-portal test` — 186 tests pass (16 files).
- [x] `pnpm lint` — 22 tasks successful, no new warnings introduced.
- [x] Clinician-portal `tsc --noEmit` — no new type errors (pre-existing
  `layout-skip-nav.test.tsx` errors unchanged).
- [x] Grepped `new Proxy(` in both portals to justify Option 1.

Closes #848

### DIFF
--- a/apps/clinician-portal/src/__tests__/helpers/trpc-mock.ts
+++ b/apps/clinician-portal/src/__tests__/helpers/trpc-mock.ts
@@ -1,0 +1,135 @@
+/**
+ * Shared tRPC mock helper for render / a11y tests.
+ *
+ * Many tests only care about one slice of the UI (e.g. a tab strip,
+ * a button, a layout shell) but the components they render pull
+ * dozens of tRPC queries and mutations. Re-implementing a permissive
+ * stub in every test file is noisy and fragile — any new call path
+ * added downstream breaks unrelated tests.
+ *
+ * `createPermissiveTrpcMock` returns a Proxy that answers any
+ * `trpc.<router>.<...>.<proc>.useQuery(...)` or `.useMutation()` call
+ * with a loading-but-quiet stub, plus a no-op `trpc.useUtils()` whose
+ * `.invalidate()` always resolves. Callers may pass per-route
+ * overrides keyed by the dot-separated procedure path (e.g.
+ * `"patients.getById"`) when they need a real-looking value for a
+ * single query.
+ *
+ * Scope is intentionally small: only what the current caller set
+ * needs. Expand deliberately, not speculatively.
+ */
+import { vi } from "vitest";
+
+/** Minimal shape of a React Query `useQuery` result our tests read. */
+export interface StubQueryResult<T = unknown> {
+  data: T | undefined;
+  isLoading: boolean;
+  isError: boolean;
+  [key: string]: unknown;
+}
+
+/** Minimal shape of a React Query `useMutation` result our tests read. */
+export interface StubMutationResult {
+  mutate: ReturnType<typeof vi.fn>;
+  isPending: boolean;
+  [key: string]: unknown;
+}
+
+/**
+ * Per-route overrides. Keys are dot-separated procedure paths as
+ * they would appear on the tRPC client (e.g. `"patients.getById"`,
+ * `"clinicalFlags.listForPatient"`). The value is either a
+ * `useQuery`-shaped result or a factory that produces one.
+ */
+export type TrpcMockOverrides = Record<
+  string,
+  StubQueryResult | (() => StubQueryResult)
+>;
+
+export interface CreatePermissiveTrpcMockOptions {
+  /**
+   * Default shape returned by `useQuery` when no override matches.
+   * Defaults to `{ data: undefined, isLoading: true, isError: false }`.
+   */
+  defaultQuery?: StubQueryResult;
+  /**
+   * Default shape returned by `useMutation`.
+   * Defaults to `{ mutate: vi.fn(), isPending: false }`.
+   */
+  defaultMutation?: () => StubMutationResult;
+  /** Per-route `useQuery` overrides keyed by procedure path. */
+  overrides?: TrpcMockOverrides;
+}
+
+/**
+ * Build a `vi.mock("@/lib/trpc", ...)` factory return value.
+ *
+ * Usage:
+ *
+ * ```ts
+ * vi.mock("@/lib/trpc", () =>
+ *   createPermissiveTrpcMock({
+ *     overrides: {
+ *       "patients.getById": { data: patient, isLoading: false, isError: false },
+ *     },
+ *   })
+ * );
+ * ```
+ */
+export function createPermissiveTrpcMock(
+  options: CreatePermissiveTrpcMockOptions = {},
+): { trpc: unknown } {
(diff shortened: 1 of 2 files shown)


apps/clinician-portal/src/__tests__/helpers/trpc-mock.ts
```diff
@@ -0,0 +1,135 @@
...
+    new Proxy(
+      {},
+      {
+        get(_t, prop: string) {
+          if (prop === "invalidate") {
+            return () => Promise.resolve();
+          }
+          return utilsProxy();
+        },
+      },
```
blamechris, line 112: Nit (Copilot caught this too): `Proxy` `get` traps receive `prop: string | symbol` at runtime, but both handlers type `prop` as `string`. If something like `Symbol.toStringTag`, `Symbol.iterator`, or `util.inspect.custom` is ever accessed on the `utilsProxy` chain, it gets proxied back recursively. For `utilsProxy` in particular, returning `undefined` for symbol keys avoids surprise recursion and matches how most inspection code expects absent values. Not blocking — in-test coverage doesn't trip this today — but a one-line `if (typeof prop === "symbol") return undefined;` would harden the helper for future callers.

apps/clinician-portal/src/__tests__/helpers/trpc-mock.ts
```diff
@@ -0,0 +1,135 @@
...
+          if (prop === "useUtils") return () => utilsProxy();
+          if (prop === "useQuery") {
+            return (..._args: unknown[]) => resolveQuery(path.join("."));
+          }
+          if (prop === "useMutation") {
+            return () => defaultMutation();
+          }
+          return proxy([...path, prop]);
+        },
+      },
```
blamechris, line 131: Nit: same symbol-key concern as on `utilsProxy`. If a symbol key slips through, `[...path, prop]` then `path.join(".")` will coerce it to `"Symbol(foo)"` and look it up in `overrides` (harmless miss) — but strictly typed, `prop` is `string | symbol`. Tightening to `prop: string | symbol` and early-returning `undefined` for symbols would make the type annotation honest and prevent symbol strings from ever reaching `overrides`.

apps/clinician-portal/src/__tests__/helpers/trpc-mock.ts
```diff
@@ -0,0 +1,135 @@
...
+ * needs. Expand deliberately, not speculatively.
+ */
+import { vi } from "vitest";
+
+/** Minimal shape of a React Query `useQuery` result our tests read. */
+export interface StubQueryResult<T = unknown> {
+  data: T | undefined;
+  isLoading: boolean;
+  isError: boolean;
+  [key: string]: unknown;
```
blamechris, line 28: Nit: `StubQueryResult<T>` declares `data`, `isLoading`, `isError` alongside `[key: string]: unknown` — the index signature forces the named fields to satisfy `unknown`, which works, but weakens downstream inference (e.g. `result.data` widens to `unknown | T | undefined`). If callers start reading extra React Query fields, an intersection type `{ data: T | undefined; isLoading: boolean; isError: boolean } & Record<string, unknown>` keeps the named fields strongly typed while still letting the object carry extras. Non-blocking — the current shape is fine for the one caller today.

blamechris (commented):
Inspector review — verdict: approve (non-blocking nits only).

Portal-local helper extraction is clean. API surface matches the spec (`defaultQuery` / `defaultMutation` / `overrides` with dot-path keys). The `vi.hoisted` + dynamic `await import(...)` pattern is the correct workaround for `vi.mock` factory hoisting. Scope discipline is good — the other Wave 2 a11y tests use simpler adequate mocks and were correctly left alone. Location decision (option 1 portal-local) is justified by a single current consumer; relocation to `packages/test-utils` is a trivial move if a second caller appears.

Three NITs inline, all non-blocking. (Self-approve blocked by GH, so this is a COMMENT event — treat as green from Inspector's side.)

Issue #851, opened 2026-04-18
chore(ai): align unrecognized_lab_flag warn event name with metric field (naming consistency)


blamechris:
Found during review of PR #842.

In `services/ai-oversight/src/rules/critical-values.ts`:

```ts
logger.warn("unrecognized_lab_flag", {
  metric: "unrecognized_lab_flag_total",
  flag: result.flag,
  test_name: result.test_name,
  test_code: result.test_code,
});
```

The warn event name (`"unrecognized_lab_flag"`) and the `metric` field (`"unrecognized_lab_flag_total"`) diverge. Elsewhere in this service the two match — e.g. `services/ai-oversight/src/utils/validate-event-timestamp.ts`:

```ts
logger.warn("timestamp_fallback_total", {
  metric: "timestamp_fallback_total",
  ...
});
```

Align one of the two so log-search and metric-search find the same string. Either:
- `logger.warn("unrecognized_lab_flag_total", { metric: "unrecognized_lab_flag_total", ... })` (match existing `_total` convention), or
- drop the `metric` field if the event name is sufficient.

No behavior change, purely observability naming consistency.

PR #860, opened 2026-04-18
fix(ai): promote HL7v2 HH/LL to critical and flag no-flag/no-range labs (#849, #853, #835)

### LINKED ISSUE #835: feat(ai): catch wildly out-of-range labs with no flag and no reference data
Found during review of PR #825 (issue #244 fix).

PR #825 closes the silent-drop for labs with an explicit `flag` value, but a related gap remains: a result with **no `flag`**, **no `reference_low` / `reference_high`**, and **not in `COMMON_LAB_TESTS`** is never flagged by the deterministic rule regardless of how extreme the value is.

Example: an ingestion pipeline delivers `{ test_name: \"Potassium Level\" (non-canonical), value: 8.0, unit: \"mEq/L\" }` with no flag and no ref range. Because `\"Potassium Level\"` doesn't match the canonical names in `CRITICAL_LAB_THRESHOLDS.POTASSIUM` exactly or the `COMMON_LAB_TESTS` key, the result slips past the deterministic layer entirely.

**Possible mitigations (for discussion):**
1. Fuzzy matching / synonym expansion against `CRITICAL_LAB_THRESHOLDS.names` and LOINC.
2. Emit a low-severity \"unable-to-evaluate\" flag when a lab result lacks both a flag and any reference range — gives downstream LLM review a signal.
3. Explicit observability metric for \"heuristic fallback reached empty branch\" so we can quantify the gap.

**Why clinical-safety:** this is the exact spirit of #244 applied to the \"no flag, no range\" variant. Not a regression from PR #825, but worth capturing before it bites.

Pre-existing behavior; not blocking PR #825.

### LINKED ISSUE #849: fix(ai): HH/LL mapping pre-empts numeric range check — potential severity-ceiling
Found during review of PR #842.

In `services/ai-oversight/src/rules/critical-values.ts`, the new HL7v2 mapping for `HH` / `LL` sets `severity = "warning"` in the flag-handling branch. The subsequent numeric range-check block is gated on `if (severity === null)`, so once HH/LL has assigned a warning-level severity, the "far outside reference range" / COMMON_LAB_TESTS fallback is **skipped**.

Concrete scenario:
```
{ flag: "LL", value: 0.1, unit: "mEq/L", reference_low: 3.5, reference_high: 5.0 }
```
Range deviation = 1.5; value 0.1 is well below `reference_low - range = 2.0`. The numeric check would fire **critical**, but the HH/LL mapping pre-empts with **warning**. This is a severity ceiling, not a floor — and it's the same class of signal-suppression failure that issue #244 was filed to address at the ingestion boundary.

The PR description states the intent is `warning` as a floor, so the current implementation doesn't match intent. Suggested remediation (non-exhaustive):

1. Move the HL7 mapping to set `severity` **only if** `severity` is still null after the numeric check, OR
2. Upgrade to `max(severity, mappedSeverity)` after the numeric check, OR
3. Track `minimumSeverity` separately from `severity` and take the max at emission time.

**Why clinical-safety:** panic-low / panic-high labs are exactly the results where under-alerting is harmful. The PR is strictly net-positive over main (today HH/LL fall through entirely), but this ceiling should be closed in a follow-up.

Original call-out from Copilot on PR #842: "setting `severity` here prevents the later numeric 'far outside range' / COMMON_LAB_TESTS checks from running (they're gated on `severity === null`). That makes the HH/LL mapping a ceiling rather than a floor and can suppress a critical flag that would otherwise be emitted."

### LINKED ISSUE #853: discuss(ai): revisit HL7v2 HH/LL mapping — should panic flags map to critical, not warning?
Found during review of PR #842.

PR #842 introduces conservative HL7v2 mapping:
- `HH` → **warning**, direction=high
- `LL` → **warning**, direction=low

The PR rationale: "Chosen severity is warning (not critical) as a conservative floor."

**Counter-argument for clinical-safety review:** HL7v2 semantics define `HH` and `LL` as "panic high" / "panic low" — these are the lab's explicit signal that the value is above/below **panic** thresholds, which is a stronger claim than the single-letter `H` / `L` (simply out-of-reference-range). Mapping single-letter `H` / `L` to warning is correct; mapping double-letter `HH` / `LL` to the *same* severity silently compresses the lab's own severity stratification.

Common HL7v2 precedence in published mappings:
- `H`, `L` → warning / out-of-range
- `HH`, `LL` → critical / panic
- `A` → abnormal-unspecified (ambiguous, leave unmapped)

Consider promoting `HH` / `LL` to `severity: "critical"`. This would also sidestep the ceiling-vs-floor concern in issue #849 (a critical flag wouldn't under-alert a truly panic-low value even if the numeric range check is skipped).

**Why clinical-safety:** under-alerting a lab-self-declared panic value is the exact failure mode #244 was filed to prevent at the ingestion boundary.

Note: this is a deliberate design discussion, not a bug — filing per review guidance to make the severity choice explicit and revisit with clinical input. #849 addresses the code-level ceiling regardless of which severity is chosen.

blamechris:
## Summary

Three interlocking clinical-safety follow-ups from PR #842, bundled because
they all touch `services/ai-oversight/src/rules/critical-values.ts` and
share a failure class: silent or under-severity lab signals at the
deterministic rule layer.

- **#849** — HH/LL mapping pre-empted the numeric range-check branch,
  creating a severity *ceiling* when the intent was a floor. Panic-low
  values that would otherwise fire critical (e.g. `LL` with value 0.1,
  ref_low 3.5, ref_high 5.0) were compressed to warning.
- **#853** — HL7v2 precedent: single-letter `H` / `L` = out-of-range;
  double-letter `HH` / `LL` = panic. Compressing `HH` / `LL` to the same
  severity as `H` / `L` silently flattens the lab's own stratification.
- **#835** — No-flag, no-range, unknown-name labs slip past the
  deterministic layer entirely. A non-canonical "Potassium Level" with
  value 8.0 receives no signal at all.

### `services/ai-oversight/src/rules/critical-values.ts`

1. **HL7v2 mapping promoted (fixes #849 + #853 together).**
   `HH` → `{ severity: "critical", direction: "high" }`
   `LL` → `{ severity: "critical", direction: "low" }`
   Single-letter `H` / `L` remain `warning` (unchanged from PR #842).
   Because `critical` is the top severity in `FlagSeverity`, the
   numeric range-check gate (`if (severity === null)`) cannot suppress
   it — the severity-ceiling concern of #849 is resolved structurally,
   not by rewiring the gate.

2. **New "unable-to-evaluate" fallback (fixes #835).**
   When a lab result has no `flag`, no `reference_low` / `reference_high`,
   and no `COMMON_LAB_TESTS` entry, and did not match any
   `CRITICAL_LAB_THRESHOLDS` definition, we now:
   - Emit `logger.warn("lab_unevaluated", { metric, test_name, test_code, value, unit })`
     for operator observability.
   - Emit an `info`-severity `RuleFlag` so downstream LLM review has
     visibility into the silent result.

   **Severity choice — `info`.** The `FlagSeverity` enum already includes
   `"info"` (see `packages/shared-types/src/ai-flags.ts` and the Zod
   `flagSeveritySchema` in `packages/validators/src/ai-flags.ts`), so no
   schema churn. `info` is intentional — we do **not** know whether the
   value is abnormal; `critical` would be over-alerting, and `warning`
   implies a clinical judgment we cannot support from the data given.
   The signal exists to surface the result for LLM review, not to fire
   a clinical alert directly.

3. **Guard: `matchedExplicitDef`.** A new flag tracks whether the
   test name matched any `CRITICAL_LAB_THRESHOLDS` definition, even
   when `evaluate` returned `null` (normal value). This prevents the
   unevaluated fallback from re-surfacing benign canonical labs — e.g.
   `Lactate` with value 1.5 (no reference range, not in the test mock's
   `COMMON_LAB_TESTS`) now correctly stays silent.

### `services/ai-oversight/src/__tests__/critical-values.test.ts`

- Updated the two pre-existing `HH` / `LL` tests from PR #842 to
  assert `critical` severity (was `warning`).
- Added regression-guard tests for #849's concrete scenarios:
  - `LL`, value 0.1, ref_low 3.5, ref_high 5.0 → critical/low
  - `HH`, value 999, ref_low 100, ref_high 200 → critical/high
- Added regression guards pinning `H` → warning and `L` → warning
  (so a future refactor cannot accidentally promote them alongside HH/LL).
- Added a new `describe("... no-flag/no-range gap (issue #835)")` block
  with seven tests covering: the exact #835 "Potassium Level" scenario,
  the `logger.warn("lab_unevaluated")` observability signal, and four
  negative cases (lab in COMMON_LAB_TESTS, lab with reference range,
  lab with any flag, lab matching CRITICAL_LAB_THRESHOLDS, and a
  non-lab vital event) to ensure the fallback activates only for
  genuinely un-evaluable results.
- Updated the `"does not flag unknown lab with no reference..."` test
  from the heuristic-fallback describe block: it now asserts the new
  info-severity unevaluated signal (the old assertion was literally
  pinning the bug #835 was filed against).

## Clinical safety rigor

- `CRITICAL_LAB_THRESHOLDS` and every hard-coded numeric threshold are
  unchanged.
- `rule_id` naming follows existing conventions
  (`WARNING-LAB-UNEVALUATED-<name>`); rule-id harmonization is
  explicitly out of scope (tracked by #836).
- All Wave 1 (#244) and Wave 2 (#833 / #834 / #837) tests pass
  unchanged.
- Full `@carebridge/ai-oversight` suite: **622 passing** (611 baseline
  from Wave 2 + 11 new tests).
- `pnpm typecheck` green.
- `pnpm lint` green (only pre-existing portal warnings).

## Test plan

- [x] `pnpm --filter @carebridge/ai-oversight test` — 622 passed
- [x] `pnpm lint` — green (pre-existing portal warnings only)
- [x] #849 regression guard: `LL` with value far below ref range → critical
- [x] #853 HL7v2 semantics: `HH` → critical/high, `LL` → critical/low
- [x] #835 regression guard: unknown lab with no flag/range/COMMON match → info flag
- [x] Negative cases: `H` / `L` still warning; canonical labs still critical;
      labs with reference range not re-surfaced; vitals not re-surfaced
- [ ] Reviewer to confirm `info` severity is acceptable for the
      unevaluated signal (alternative was metric-only `logger.warn`)
- [ ] Reviewer to confirm rule_id `WARNING-LAB-UNEVALUATED-*` is OK
      (could be `INFO-LAB-UNEVALUATED-*`; kept `WARNING-` prefix to
      follow the existing same-level naming in this file — #836 will
      harmonize)

Closes #849
Closes #853
Closes #835

### DIFF
--- a/services/ai-oversight/src/__tests__/critical-values.test.ts
+++ b/services/ai-oversight/src/__tests__/critical-values.test.ts
@@ -313,13 +313,20 @@ describe("checkCriticalValues — lab heuristic fallback", () => {
     expect(flags[0]!.rule_id).toBe("CRITICAL-LAB-HEMOGLOBIN");
   });
 
-  it("does not flag unknown lab with no reference and no COMMON_LAB_TESTS entry", () => {
+  it("emits an info-severity 'unable-to-evaluate' signal for unknown lab with no reference and no COMMON_LAB_TESTS entry (issue #835)", () => {
+    // Issue #835: previously this result was silently dropped. The rule
+    // layer cannot assess the value against any threshold, but it also
+    // must not leave a totally silent lab result invisible to downstream
+    // LLM review. Emit info severity (not critical — we do not know the
+    // value is abnormal) to surface the result without over-alerting.
     const flags = checkCriticalValues(
       makeLabEvent([
         { test_name: "Obscure Marker X", value: 999, unit: "U/L" },
       ]),
     );
-    expect(flags).toHaveLength(0);
+    expect(flags).toHaveLength(1);
+    expect(flags[0]!.severity).toBe("info");
+    expect(flags[0]!.rule_id).toBe("WARNING-LAB-UNEVALUATED-OBSCURE_MARKER_X");
   });
 });
 
@@ -713,10 +720,12 @@ describe("checkCriticalValues — unrecognized flag handling (issues #834, #837)
     expect(unrecognizedCalls).toHaveLength(0);
   });
 
-  it("maps HL7v2 'HH' (panic high) to a warning flag with direction='high'", () => {
-    // HL7v2 abnormal-flags: "HH" means panic/critical-high. Treat as warning
-    // (not critical) to avoid over-alerting pending principled mapping —
-    // see issue #834 discussion. Warning is the floor, not the ceiling.
+  it("maps HL7v2 'HH' (panic high) to a critical flag with direction='high' (issues #849, #853)", () => {
+    // HL7v2 abnormal-flags: "HH" means panic/critical-high. Promoting HH to
+    // critical matches HL7v2 semantics (panic > out-of-range) AND naturally
+    // resolves the severity-ceiling concern of issue #849: even though the
+    // numeric range-check branch is gated on `severity === null`, the
+    // HL7 mapping now emits the top severity so nothing is being suppressed.
     const flags = checkCriticalValues(
       makeLabEvent([
         {
@@ -728,14 +737,14 @@ describe("checkCriticalValues — unrecognized flag handling (issues #834, #837)
       ]),
     );
     expect(flags).toHaveLength(1);
-    expect(flags[0]!.severity).toBe("warning");
-    expect(flags[0]!.summary).toContain("High");
+    expect(flags[0]!.severity).toBe("critical");
+    expect(flags[0]!.summary.toLowerCase()).toContain("high");
     // This mapping should still emit the warn log so operators notice
     // non-enum flags arriving through FHIR/HL7 ingress paths.
     expect(mockWarn).toHaveBeenCalled();
   });
 
-  it("maps HL7v2 'LL' (panic low) to a warning flag with direction='low'", () => {
+  it("maps HL7v2 'LL' (panic low) to a critical flag with direction='low' (issues #849, #853)", () => {
     const flags = checkCriticalValues(
       makeLabEvent([
         {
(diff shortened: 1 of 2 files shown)


blamechris:
## Agent review: approved (self-PR; posted as comment)

Clinical-safety rigor applied. 3 NITs filed as follow-ups. This PR is strictly net-positive for patient safety and correctly addresses all three tracked issues.

### Clinical-safety verification

1. **#849 severity ceiling — fully resolved.** Promoting `HH` / `LL` to `critical` is a structural fix: because `critical` is the top severity, the `if (severity === null)` range-check gate can no longer suppress a panic-flagged lab. The #849 regression-guard test (`LL`, value 0.1, ref_low 3.5) pins this.

2. **#853 HL7v2 semantics — correctly implemented.** `HH` / `LL` → `critical` matches HL7v2 panic-flag semantics. Single-letter `H` / `L` remain `warning`; regression guards pin both.

3. **#835 no-flag/no-range fallback — correctly gated.** Control flow verified end-to-end:
   - Branch 1 (CRITICAL_LAB_THRESHOLDS) runs first; on name match, sets `matchedExplicitDef=true` even when `evaluate` returns null.
   - Branch 2 (flag/range heuristic) gated on `!matchedExplicit` — cannot double-fire with branch 1.
   - Branch 3 (info-severity unevaluated) gated on `severity === null` AND `!matchedExplicitDef` AND `!hasReferenceRange` AND `!inCommonLabTests` AND `flag == null`.

4. **Specific case audited: `flag=LL` on canonical "Potassium" with normal value 4.2.** Branch 1 matches by name, `matchedExplicitDef=true`, `evaluate(4.2)` returns null → no emission. Branch 2 runs (`matchedExplicit=false`), sees flag=LL, fires `critical`. Branch 3 skipped. Correct — no double-fire, panic flag respected, severity ceiling fix propagates cleanly.

5. **`info` severity valid.** Confirmed in `packages/shared-types/src/ai-flags.ts:6` and `packages/validators/src/ai-flags.ts:3`. No schema churn.

6. **`rule_id` distinctness.** `WARNING-LAB-UNEVALUATED-<test>` cannot collide with any existing `WARNING-LAB-<test>` or `CRITICAL-LAB-<test>` because of the `UNEVALUATED-` infix.

7. **`CRITICAL_LAB_THRESHOLDS` untouched.** Verified.

8. **Test coverage.** 11 new tests, 622 passing total (611 baseline preserved). Build/Test/Typecheck CI green.

9. **Zero Attribution.** Clean across commits, PR body, diff, and code.

### NITs (filed, non-blocking)

- **#863** — `logger.warn("lab_unevaluated", { metric: "lab_unevaluated_total" })` diverges event name from metric field. Replicates the exact inconsistency #851 tracks for `unrecognized_lab_flag`. Convention per `validate-event-timestamp.ts` is both-match with `_total` suffix.
- **#864** — `rule_id: "WARNING-LAB-UNEVALUATED-*"` for `info`-severity flag is a semantic mismatch. PR explicitly defers to #836 (rule_id harmonization); filed to record the option space.
- **#867** — Clinical follow-up: should `info`-severity escalate to `warning` when unmatched value is numerically implausible (e.g. 999 for any analyte)? Current behavior strictly better than silent, but residual under-alerting risk exists.

### Praise

- **Structural #849 fix.** Resolving a severity-ceiling by promoting the top severity rather than rewiring the gate is the more robust choice — leaves less room for regression.
- **`matchedExplicitDef` tracking.** Carefully distinguishes "name matched a threshold def" from "threshold actually fired," preventing the #835 fallback from re-surfacing benign canonical labs (e.g. Lactate at 1.5 with no reference range). Subtle and correct.
- **Seven negative-case tests** in the `no-flag/no-range gap` describe block — each documenting why the branch must NOT activate. Excellent regression scaffolding.
- **Commit message.** Explicit patient-safety rationale, each of the three issues addressed separately, fallback-guard invariants enumerated, regression-guard status pinned.

PR #861, opened 2026-04-18
feat(ai): add 3 cross-specialty rules and extract getRecentPotassium helper

blamechris:
Builds on Wave 2 PR #843. Adds three high-value patient-safety rules from
issue #263, and extracts the helper requested in the #843 PR review (#855).

### 1. `DI-STATIN-GEMFIBROZIL` (critical) — drug-interactions.ts
Upgrades the prior `DI-STATIN-FIBRATE` (warning) to two separate rules based
on fibrate identity:

- **`DI-STATIN-GEMFIBROZIL` — critical.** Gemfibrozil inhibits statin
  glucuronidation and OATP1B1-mediated hepatic uptake, raising plasma statin
  levels several-fold. This combination drove the fatal rhabdomyolysis cases
  behind Pfizer's 2001 withdrawal of cerivastatin (Baycol). FDA labeling
  now contraindicates gemfibrozil with simvastatin, lovastatin, pitavastatin,
  and rosuvastatin, and recommends avoidance with other statins.
- **`DI-STATIN-FENOFIBRATE` — warning** (retained severity). Fenofibrate
  does not inhibit statin glucuronidation, so the interaction is clinically
  manageable with CK monitoring — ACC/AHA and NLA support this combination
  for mixed dyslipidemia.

### 2. `CROSS-METFORMIN-GFR-001` (critical) — cross-specialty.ts
Fires when the patient is on metformin (or any metformin-containing combo
product) **and** a recent eGFR value is `< 30 mL/min/1.73m²`. This is a
hard contraindication per FDA 2016 labeling and ADA guidance — accumulation
in advanced renal impairment drives metformin-associated lactic acidosis
(MALA), mortality 30–50%.

- Threshold is strict `< 30` to mirror the published cutoff (eGFR = 30 does
  not fire).
- Recognises `eGFR`, `GFR`, and `Estimated GFR` lab names.
- Complementary to the existing `DI-METFORMIN-CONTRAST` (which covers
  transient peri-contrast AKI).

### 3. `CROSS-THIAZIDE-HYPOK-001` (warning, escalates to critical) — cross-specialty.ts
Fires when the patient is on a thiazide (HCTZ, chlorthalidone, indapamide,
metolazone, chlorothiazide) and K+ is `< 3.5 mEq/L`. Thiazides worsen
pre-existing hypokalemia via distal-tubule K+ wasting, increasing arrhythmia,
weakness, and rhabdomyolysis risk.

[...]

PR #870, opened 2026-04-18
feat(ai): unit-aware recent_labs in rule-level PatientContext

### LINKED ISSUE #856: feat(ai): defensive unit check on recent_labs numeric values in rules
Follow-up NIT from /agent-review on #843.

The `recent_labs` entries in `PatientContext` (`services/ai-oversight/src/rules/cross-specialty.ts`) currently expose `{ name: string; value: number }` only — no `unit` field on the rule-level shape (the LLM path does carry unit, but the rule path doesn't).

For potassium (`CROSS-QT-HYPOK-001`), mEq/L and mmol/L are numerically identical (1:1), so the threshold of <3.5 is safe. But this is brittle for future electrolyte-aware or dose-aware rules where mg/dL vs. mmol/L differ by order of magnitude (e.g., glucose: 120 mg/dL = 6.7 mmol/L; creatinine: 1.0 mg/dL = 88.4 µmol/L).

**Proposal:** one of:
1. Extend the rule-level `recent_labs` shape to include `unit` and add assertions in `buildPatientContextForRules` that labs are normalized to expected units (mEq/L for K+, mg/dL for Cr, etc.).
2. Or, add a canonical-unit normalization helper in `context-builder.ts` / `review-service.ts` so all rules see values in a single, documented unit per analyte.

Would prevent a class of silently-wrong threshold comparisons in future cross-specialty / dose-aware rules.

Raised during /agent-review of #843.

blamechris:
## Summary

Extend the rule-level `PatientContext.recent_labs` shape to include a required `unit: string` field (issue #856) and add defensive unit-aware lab accessors so threshold-based rules cannot silently compare wrong-unit values.

## Why

For potassium (CROSS-QT-HYPOK-001) the current rule code compares a raw `value` against `< 3.5` — safe as long as the recorded unit is `mEq/L` or `mmol/L`, since those are 1:1 equivalent for monovalent ions. But an EHR import or data-entry error that puts `K+ 3.2 mg/dL` into `lab_results.value` would still trip the rule. For analytes like glucose or creatinine, where `mg/dL <-> mmol/L` requires molar-mass conversion, the risk is worse. This PR closes that gap at the shape level.

## Changes

- **`PatientContext.recent_labs`** — each entry now carries `unit: string`. Empty string denotes "unknown" (source row had null unit).
- **`buildPatientContextForRules`** — selects `lab_results.unit`, forwards it to the rule context, emits `lab_result_unit_missing` warn for rows with a null/empty unit.
- **`rules/lab-units.ts`** (new) — `findRecentLab(ctx, namePattern, acceptedUnits, analyte)` plus analyte-specific helpers `getRecentPotassium`, `getRecentSodium`, `getRecentChloride`, `getRecentEGFR`. Mismatched / missing units fail closed and log a structured warn.
- **All K+ / eGFR threshold rules migrated** to the unit-aware helpers. Previously only CROSS-QT-HYPOK-001 was in scope; merging main (which introduced #861's CROSS-THIAZIDE-HYPOK-001 and CROSS-METFORMIN-GFR-001) extends coverage to:
  - `CROSS-QT-HYPOK-001` (K+)
  - `CROSS-THIAZIDE-HYPOK-001` (K+)
  - `CROSS-METFORMIN-GFR-001` (eGFR)
  The local `getRecentPotassium` / `getRecentEGFR` helpers introduced by #861 in `cross-specialty.ts` are removed; both callers now import from `./lab-units.js`.

### Normalization policy

Conservative — only 1:1 numeric aliases are listed as "equivalent":

- Potassium / sodium / chloride: `mEq/L` and `mmol/L` (monovalent ions).
- eGFR: `mL/min/1.73m²` and tolerant spacing/encoding variants.

Anything requiring molar-mass conversion (mg/dL <-> mmol/L for organic compounds) is intentionally NOT aliased — that would be a real conversion and belongs in a separate, well-tested module. Filed as a follow-up only if a threshold rule ever needs it.

## Test plan

- [x] TDD — failing tests written first for:
  - `CROSS-QT-HYPOK-001` with K+ in `mEq/L`, `mmol/L`, `mg/dL`, empty string, case/whitespace variants.
  - `buildPatientContextForRules` passes `unit` through the shape and maps null -> empty string.
  - Direct helper coverage in `lab-units.test.ts` (17 cases).
- [x] #861 rule tests (`CROSS-THIAZIDE-HYPOK-001`, `CROSS-METFORMIN-GFR-001`) migrated to supply `unit: "mEq/L"` / `unit: "mL/min/1.73m²"` in recent_labs fixtures after the merge.
- [x] `pnpm --filter @carebridge/ai-oversight test` — 28 files / 674 tests pass.
- [x] `pnpm typecheck` — clean.
- [x] `pnpm lint` — clean (pre-existing clinician-portal warnings unrelated).

Closes #856

### DIFF
--- a/services/ai-oversight/src/__tests__/build-patient-context.test.ts
+++ b/services/ai-oversight/src/__tests__/build-patient-context.test.ts
@@ -40,6 +40,7 @@ vi.mock("@carebridge/db-schema", () => ({
     created_at: "created_at",
     test_name: "test_name",
     value: "value",
+    unit: "unit",
   },
   reviewJobs: {},
 }));
@@ -95,7 +96,7 @@ describe("buildPatientContextForRules — recent_labs wiring", () => {
     primeSelectMocks();
   });
 
-  it("populates recent_labs from the joined lab_results query", async () => {
+  it("populates recent_labs from the joined lab_results query, including unit (#856)", async () => {
     diagnosesSelect.mockResolvedValue([
       { description: "Breast cancer", icd10_code: "C50.9", status: "active" },
     ]);
@@ -103,16 +104,16 @@ describe("buildPatientContextForRules — recent_labs wiring", () => {
       { name: "Cisplatin", status: "active" },
     ]);
     labsSelect.mockResolvedValue([
-      { test_name: "ANC", value: 800, created_at: nowIso },
-      { test_name: "WBC", value: 2.1, created_at: nowIso },
+      { test_name: "ANC", value: 800, unit: "cells/µL", created_at: nowIso },
+      { test_name: "WBC", value: 2.1, unit: "10^3/µL", created_at: nowIso },
     ]);
 
     const ctx = await buildPatientContextForRules("p-1", stubEvent);
 
     expect(ctx.recent_labs).toBeDefined();
     expect(ctx.recent_labs).toEqual([
-      { name: "ANC", value: 800 },
-      { name: "WBC", value: 2.1 },
+      { name: "ANC", value: 800, unit: "cells/µL" },
+      { name: "WBC", value: 2.1, unit: "10^3/µL" },
     ]);
     expect(ctx.event_timestamp).toBe(stubEvent.timestamp);
     expect(ctx.active_diagnoses).toContain("Breast cancer");
@@ -124,13 +125,29 @@ describe("buildPatientContextForRules — recent_labs wiring", () => {
     medicationsSelect.mockResolvedValue([]);
     // Ordered desc by created_at — first row wins.
     labsSelect.mockResolvedValue([
-      { test_name: "ANC", value: 500, created_at: nowIso },
-      { test_name: "ANC", value: 1800, created_at: "2026-04-05T00:00:00Z" },
+      { test_name: "ANC", value: 500, unit: "cells/µL", created_at: nowIso },
+      { test_name: "ANC", value: 1800, unit: "cells/µL", created_at: "2026-04-05T00:00:00Z" },
     ]);
 
     const ctx = await buildPatientContextForRules("p-1", stubEvent);
 
-    expect(ctx.recent_labs).toEqual([{ name: "ANC", value: 500 }]);
+    expect(ctx.recent_labs).toEqual([{ name: "ANC", value: 500, unit: "cells/µL" }]);
+  });
+
+  it("populates unit='' (empty string) when the lab row has null/missing unit, and logs a warn (#856)", async () => {
+    // Defensive: a lab row with null unit must not crash and must not be
+    // silently treated as if it had a canonical unit. The shape should
+    // carry an empty string, and downstream unit-aware helpers refuse to
+    // match it against a specific expected unit.
+    diagnosesSelect.mockResolvedValue([]);
+    medicationsSelect.mockResolvedValue([]);
+    labsSelect.mockResolvedValue([
+      { test_name: "Potassium", value: 3.2, unit: null, created_at: nowIso },
+    ]);
+
+    const ctx = await buildPatientContextForRules("p-1", stubEvent);
+
+    expect(ctx.recent_labs).toEqual([{ name: "Potassium", value: 3.2, unit: "" }]);
   });
 
   it("returns recent_labs as undefined when no recent lab rows exist", async () => {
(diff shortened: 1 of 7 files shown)


blamechris (commented):
## Review — unit-aware recent_labs (#870)

Thorough, well-scoped PR with TDD coverage and conservative alias policy. The overall
direction (fail-closed unit check, observability warns, monovalent-ion 1:1 aliasing only)
is exactly right for a patient-safety surface. However, there are two blocking items tied
to rules that already exist on `main` via the merged PR #861.

### BLOCKING

1. **Branch is stale — must rebase onto `main` and migrate CROSS-METFORMIN-GFR-001
   and CROSS-THIAZIDE-HYPOK-001.** `main` now contains two rules not yet on this
   branch:
   - `CROSS-METFORMIN-GFR-001` — uses `getRecentEGFR(ctx)` to compare eGFR `< 30`.
   - `CROSS-THIAZIDE-HYPOK-001` — uses `getRecentPotassium(ctx)` to compare K+ `< 3.5`
     and escalates to critical `< 3.0`.

   Both use module-local helpers in `cross-specialty.ts` with signature
   `(ctx) => number | undefined`. This PR introduces `lab-units.ts` with helpers of
   the **same names** but signature `(ctx) => RecentLab | undefined`. After rebase:
   - Either the main-branch local helpers remain and the new `lab-units.ts` exports
     collide (rebase will fail to compile), or
   - The new `lab-units.ts` helpers win and the two call sites in
     `CROSS-METFORMIN-GFR-001` / `CROSS-THIAZIDE-HYPOK-001` break (comparing a
     `RecentLab` object against a number, always false in TS-strict).

   More importantly — the whole point of this PR is that K+ and eGFR comparisons must
   be unit-aware. Leaving CROSS-THIAZIDE-HYPOK-001 (K+) and CROSS-METFORMIN-GFR-001
   (eGFR) on the old numeric-only path means a mg/dL potassium or raw mL/min creatinine
   clearance can still silently trip those rules. Migrate them in this PR, update
   their tests to include `unit: "mEq/L"` / `unit: "mL/min/1.73m²"` positive cases and
   mg/dL / ml/min-only / empty-unit negative cases, and verify severity escalation
   (K+ `< 3.0` critical) still passes.

   PR description currently says "634/634 tests pass (up from 611)" — this is
   inconsistent with main's baseline of 640 after #861, which confirms the branch has
   not caught up.

2. **Drop `ml/min` from `EGFR_ACCEPTED_UNITS`.** The module comment says "All listed
   values are numerically equivalent — we only tolerate formatting variance, not a real
   unit mismatch," but `ml/min` is included:

   ```ts
   const EGFR_ACCEPTED_UNITS: ReadonlySet<string> = new Set([
     "ml/min/1.73m2",
     "ml/min/1.73m²",
     "ml/min/1.73 m2",
     "ml/min/1.73 m²",
     "ml/min",   // ← not equivalent to mL/min/1.73m²
   ]);
   ```

   `mL/min` (raw, non-BSA-indexed — typical for Cockcroft-Gault creatinine clearance)
   is NOT numerically equivalent to `mL/min/1.73m²` (BSA-indexed — MDRD/CKD-EPI eGFR);
   they can differ by 20–30% in patients with non-average BSA. For CROSS-METFORMIN-GFR-001
   with a hard `< 30` threshold, accepting `mL/min` as if it were eGFR is exactly the
   class of silent wrong-unit comparison this PR is supposed to prevent. Remove it and
   add a negative test (`unit: "mL/min"` must NOT match `getRecentEGFR`).

### NITs (non-blocking, safety-relevant)

3. **eGFR name regex doesn't tolerate "e?gfr" variant.** `/^(egfr|estimated gfr|gfr)$/i`
   matches only exact "egfr". Main's prior regex is `/^(e?gfr|estimated\s*gfr)$/i` —
   tolerates "GFR" alone and "Estimated GFR" with arbitrary whitespace. Consider
   aligning so the migration is a strict superset.

4. **Chloride regex `/^(chloride|cl-?)$/i`** — accepts "Cl" and "Cl-" but not "Cl−"
   (U+2212 minus). EHRs usually use ASCII, probably fine, but worth a note.

5. **`lab_result_unit_missing` warn shape** includes `patient_id_prefix: patientId.slice(0, 8)`
   — good HIPAA-conscious truncation. Worth spot-checking whether the logger redacts
   PHI automatically elsewhere; consistent with other sites in `review-service.ts`.

### PRAISE

- **Fail-closed semantics are correct.** `findRecentLab` skips (continues loop) on
  mismatched or missing unit and emits a structured warn with `metric:` tag —
  observable without silently comparing wrong-unit values. Test coverage includes
  the "skip mismatched, accept later matching" case which is a subtle but important
  behaviour.
- **Conservative alias list is exactly right.** mEq/L ↔ mmol/L for monovalent ions
  only; explicit rejection of molar-mass conversions in the module doc. This is
  the hard-to-resist-scope-creep design decision this feature needed.
- **Existing CROSS-QT-HYPOK-001 tests preserved and extended** with unit positive/negative
  (mEq/L fires, mmol/L fires, mg/dL does NOT fire, empty does NOT fire, case/whitespace
  tolerant). Severity-escalation test (K+ `< 3.0` critical) retained.
- **Observability coverage:** three distinct structured warns (`rule_lab_unit_mismatch`,
  `rule_lab_unit_missing`, `lab_result_unit_missing`) fire at the expected boundaries.
- **Source-gap handling:** null unit at source degrades to empty string + warn; no
  crash. Test `populates unit='' (empty string) when the lab row has null/missing unit`
  is exactly the right defensive contract.
- **INR / ANC intentionally unchanged** — correct: INR is unitless, ANC is a cell
  count (cells/µL). Worth a short comment in `cross-specialty.ts` next to the INR/ANC
  `.find(...)` call sites explaining why they're exempt, so a future reader doesn't
  "helpfully" migrate them and trip the accepted-units check.
- **Helper symmetry (K, Na, Cl, eGFR)** — consistent signatures, consistent warn tags.

### Verdict

CHANGES REQUESTED — land the rebase + migration of CROSS-THIAZIDE-HYPOK-001 and
CROSS-METFORMIN-GFR-001 to unit-aware helpers, drop `ml/min` from eGFR accepted set.
Everything else is approve-level.

PR #869, opened 2026-04-18
refactor(ai): harmonize lab rule_id prefixes and observability event names

blamechris:
## Summary

Bundled naming harmonization across the ai-oversight service's critical-lab path.
All changes are pure renaming / observability-convention alignment — no rule logic, severity values, or threshold numerics are touched.

After merging `main` (which brought in PR #860's HL7 panic promotion and the new info-severity `lab_unevaluated` branch), this PR now covers **all four** original issues:

- **#836 — rule_id prefix now matches severity.** Introduces a single helper `buildLabRuleId(severity, ruleKey)` returning `${SEVERITY}-LAB-${ruleKey}`. Previously, the explicit-threshold path hard-coded a `CRITICAL-LAB-*` prefix regardless of threshold severity (e.g. Troponin I 0.04–0.4 ng/mL is warning severity but emitted `CRITICAL-LAB-TROPONIN_I`), while the heuristic-fallback path already varied prefix by severity. Downstream consumers filtering on rule_id prefix now see consistent semantics. All warning-severity tests updated symmetrically.
- **#851 — logger event name matches metric field.** Renamed the `logger.warn` event string from `"unrecognized_lab_flag"` to `"unrecognized_lab_flag_total"` so it equals the `metric` field value, aligning with the convention in `utils/validate-event-timestamp.ts` (`timestamp_fallback_total`).
- **#863 — lab_unevaluated event name alignment.** Renamed the `logger.warn` event string from `"lab_unevaluated"` to `"lab_unevaluated_total"` in the issue #835 unevaluated branch (introduced by PR #860 and brought in via the main merge). Same convention as #851.
- **#864 — INFO-LAB-UNEVALUATED-* rename.** The info-severity unevaluated emission site now calls `buildLabRuleId("info", "UNEVALUATED-${name}")`, which produces the correct `INFO-LAB-UNEVALUATED-*` prefix (previously hard-coded as `WARNING-LAB-UNEVALUATED-*` despite the flag being info severity). The helper already supports all three severities (critical / warning / info).

## Changes

- `services/ai-oversight/src/rules/critical-values.ts`
  - Added `buildLabRuleId(severity, ruleKey)` helper with docstring explaining #836.
  - Explicit-threshold emission site now calls `buildLabRuleId(threshold.severity, ruleKey)`.
  - Heuristic-fallback emission site now calls `buildLabRuleId(severity, test_name_normalized)` (previous inline ternary `severity === "critical" ? "CRITICAL" : "WARNING"` is gone).
[...]

Issue #863, opened 2026-04-18
chore(ai): align lab_unevaluated warn event name with metric field (naming consistency)


blamechris:
Found during review of PR #860.

In `services/ai-oversight/src/rules/critical-values.ts`, the new no-flag/no-range fallback emits:

```ts
logger.warn("lab_unevaluated", {
  metric: "lab_unevaluated_total",
  test_name: result.test_name,
  test_code: result.test_code,
  value: result.value,
  unit: result.unit,
});
```

The warn event name (`"lab_unevaluated"`) and the `metric` field (`"lab_unevaluated_total"`) diverge. This replicates the exact inconsistency issue #851 was filed to address for `unrecognized_lab_flag`.

For reference, the established convention (see `services/ai-oversight/src/utils/validate-event-timestamp.ts`):

```ts
logger.warn("timestamp_fallback_total", {
  metric: "timestamp_fallback_total",
  ...
});
```

**Suggested remediation:** change the warn event name from `"lab_unevaluated"` to `"lab_unevaluated_total"` so it aligns with the `metric` field and the pattern established elsewhere in this service. Related: #851 tracks the same fix for `unrecognized_lab_flag`; both could land in a single PR.

**Why this matters:** downstream log-to-metric scrapers often key on either the event name or the metric field. Divergence means half the tooling will miss signals.

Non-blocking NIT for PR #860; filing as a follow-up.

PR #875, opened 2026-04-18
feat(ai): escalate info-severity unevaluable labs with numerically extreme values

### LINKED ISSUE #867: discuss(ai): consider severity escalation for numerically extreme unevaluable labs
Found during review of PR #860.

PR #860 closed the no-flag/no-range gap (issue #835) by emitting an `info`-severity "unable-to-evaluate" flag when a lab has no flag, no reference range, and no `COMMON_LAB_TESTS` entry. The severity choice (`info`) is deliberately conservative: we do not know whether the value is abnormal, so firing `critical` would be over-alerting.

**Possible follow-up for discussion:** when the value itself is numerically extreme in absolute terms, should severity escalate even without a reference?

**Concrete scenario (hypothetical):**
```
{ test_name: "Potassium Levl" (typo of "Level"), value: 999, unit: "mEq/L" }
```

No flag, no range, no canonical match. Physiologically impossible for potassium (normal ~3.5–5.0 mEq/L, "critical" is ~>6.5). Today this fires `info`. LLM review may or may not catch it, and in the worst case it is batched behind other flags.

**Options:**

1. Keep current behavior — information-asymmetry is real and `critical` is not justifiable without a reference.
2. Emit `warning` (not `info`) when value is outside a coarse "any-analyte-implausible" range (e.g., negative, zero, or >1000 for most units) — clinician-review-worthy without claiming clinical judgment.
3. Fuzzy-match the test name against canonical synonyms / LOINC before falling through to the info-severity branch — reduces the rate of reaching this fallback at all. (This is also listed as option 1 in issue #835.)

**Why clinical-safety:** the entire reason this branch exists is that a silent drop of a clinically wild value is harmful. `info` is a strict improvement over silent, but leaves residual risk for cases where the value is unambiguously implausible.

Non-blocking; filing as a follow-up for PR #860.

blamechris:
## Summary

PR #860 introduced the info-severity "unable-to-evaluate" fallback (issue #835) for labs with no flag, no reference range, and no entry in `COMMON_LAB_TESTS`. The reviewer raised issue #867: when the value itself is numerically extreme (e.g., 99999 or 0.001 for any analyte), `info` is plausibly under-alerting — the raw magnitude alone suggests data corruption, a decimal/unit error, or a genuine out-of-range reading worth clinician review.

This PR adds a coarse magnitude-based escalation so extreme unevaluable values surface at `warning` instead of `info`.

## Chosen thresholds (signal-over-noise hints, NOT clinical thresholds)

| Condition | Severity | Rationale |
| --- | --- | --- |
| `value > 10000` | `warning` | Extreme magnitude; covers misordered units / order-of-magnitude errors. Most physiological analytes rarely exceed a few thousand in their reported units. |
| `value < 0.01` (includes negatives and zero) | `warning` | Near-zero or sub-thresholdable; covers decimal misplacement and lost precision. Negatives are almost always data errors for concentration/count analytes, so the same predicate covers them. |
| `value` in `[0.01, 10000]` | `info` (unchanged) | Genuinely unevaluable; base case from PR #860. |

**Why not `critical`?** Magnitude alone does not justify an analyte-specific clinical claim without a reference. `warning` is the ceiling here.

**Why not analyte-specific?** That is what `CRITICAL_LAB_THRESHOLDS` is for. This escalation is strictly a magnitude heuristic for the unevaluable branch.

The escalation is **additive**: values in the plausible window continue to emit `info` exactly as introduced by PR #860.

## Changes

- `services/ai-oversight/src/rules/critical-values.ts`
  - Added `isExtremeNumericValue(value)` helper with `EXTREME_VALUE_HIGH` (10000) / `EXTREME_VALUE_LOW` (0.01) constants and a docstring explaining the chosen thresholds and why escalation stays at `warning`.
  - Unevaluated-fallback emission site now branches severity on the helper and threads `severity` through `buildLabRuleId`, so the rule_id prefix flips `INFO-LAB-UNEVALUATED-*` → `WARNING-LAB-UNEVALUATED-*` in lockstep with the severity (consistent with #836 harmonization landed in #869).
  - Rationale text now explains the magnitude escalation when it fires; the plain "info" rationale is unchanged for the in-window case.
  - `logger.warn("lab_unevaluated_total", ...)` event name and metric shape are unchanged (#863 rename preserved). Observability is additive.
- `services/ai-oversight/src/__tests__/critical-values.test.ts`
  - New `describe` block "extreme-value escalation on unevaluated fallback (issue #867)" with 7 tests:
    - value 99999 → warning, `WARNING-LAB-UNEVALUATED-OBSCURE_MARKER_Y`.
    - value 0.001 → warning.
    - value -5 → warning (negative is data error; captured by `value < 0.01`).
    - value 10.5 → info (baseline unchanged).
    - Escalated flag still emits `lab_unevaluated_total` logger warn with the correct metric / test_name / value / unit fields.
    - Boundary cases `value === 10000` and `value === 0.01` remain info (strict inequality in the predicate).

## Non-goals (explicitly out of scope)

- No changes to `CRITICAL_LAB_THRESHOLDS` numeric values.
- No changes to HL7v2 flag handling (`HH` / `LL` → critical, `H` / `L` → warning).
- No changes to per-result reference range handling.
- No changes to the explicit-threshold path (#836 `buildLabRuleId` harmonization from #869 is preserved).
- No escalation to `critical`.
- No analyte-specific escalation.
- No change to the `logger.warn` event name (#863 has already landed the rename).

## Test plan

- [x] `pnpm --filter @carebridge/ai-oversight test` — 690 passed, 0 failed (+7 new tests for #867).
- [x] `pnpm typecheck` — green.
- [x] `pnpm lint` — green (pre-existing portal warnings unrelated to this change).
- [x] TDD — wrote failing tests first, confirmed 3 expected-escalation tests failed with `info`, then implemented and verified all 7 new tests pass.

Closes #867

### DIFF
--- a/services/ai-oversight/src/rules/critical-values.ts
+++ b/services/ai-oversight/src/rules/critical-values.ts
@@ -366,6 +366,43 @@ function buildLabRuleId(severity: FlagSeverity, ruleKey: string): string {
 }
 
 /**
+ * Coarse "any-analyte-implausible" numeric sanity window for the issue
+ * #835 unevaluated-lab fallback.
+ *
+ * Issue #867 (reviewer follow-up on PR #860): when the unevaluable
+ * fallback fires, the info-severity default is plausibly under-alerting
+ * for values whose magnitude alone suggests data corruption or a
+ * decimal / sign / unit error. Without a reference range or canonical
+ * mapping we cannot claim a clinical threshold, but we CAN observe that
+ * common human-scale analyte results almost never fall outside
+ * (0.01, 10000) in their own units. Values outside that window are much
+ * more likely to be clinician-review-worthy than values inside it.
+ *
+ * Chosen thresholds (signal-over-noise hints, NOT clinical thresholds):
+ *   - value > 10000  → extreme magnitude (misordered units / order-of-
+ *     magnitude error). Most physiological analytes rarely exceed a few
+ *     thousand in their reported units.
+ *   - value < 0.01   → near-zero or sub-thresholdable (decimal misplacement
+ *     or lost precision). For positive analytes this usually means the
+ *     reading is either missing or encoded with the wrong multiplier.
+ *   - value < 0      → negative concentrations / counts are almost always
+ *     data errors; captured by `value < 0.01` above (negatives satisfy
+ *     the same predicate).
+ *
+ * Escalation stays at `warning` (not `critical`) because the signal is
+ * magnitude-based, not analyte-specific: `critical` requires a clinical
+ * claim we cannot make without a reference. The escalation is additive —
+ * the base case (value in the window) continues to emit info.
+ */
+const EXTREME_VALUE_HIGH = 10000;
+const EXTREME_VALUE_LOW = 0.01;
+
+function isExtremeNumericValue(value: number): boolean {
+  if (!Number.isFinite(value)) return false;
+  return value > EXTREME_VALUE_HIGH || value < EXTREME_VALUE_LOW;
+}
+
+/**
  * Infer the direction ("low" | "high") of a laboratory-flagged critical
  * result when the lab itself did not encode direction in the flag string.
  *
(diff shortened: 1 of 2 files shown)


blamechris (commented):
Inspector review (self-auth — cannot formally approve own PR, logging verdict here).

**VERDICT: approved** (pending human/Copilot re-review sign-off).

Escalation logic is correctly scoped, threads severity through `buildLabRuleId` per #869, and test coverage is thorough (positive extreme 99999, small positive 0.001, negative -5, baseline 10.5 in-window, and both strict-inequality boundaries at exactly 10000 and 0.01). `CRITICAL_LAB_THRESHOLDS`, HL7v2 flag paths, and the `lab_unevaluated_total` logger event are untouched. Commit and diff are attribution-clean.

The decision to ceiling at `warning` (not `critical`) is the right call — magnitude alone cannot justify an analyte-specific clinical claim without a reference.

One non-blocking NIT filed inline re: non-finite handling.

services/ai-oversight/src/rules/critical-values.ts
```diff
@@ -365,6 +365,43 @@ function buildLabRuleId(severity: FlagSeverity, ruleKey: string): string {
...
+ * Escalation stays at `warning` (not `critical`) because the signal is
+ * magnitude-based, not analyte-specific: `critical` requires a clinical
+ * claim we cannot make without a reference. The escalation is additive —
+ * the base case (value in the window) continues to emit info.
+ */
+const EXTREME_VALUE_HIGH = 10000;
+const EXTREME_VALUE_LOW = 0.01;
+
+function isExtremeNumericValue(value: number): boolean {
+  if (!Number.isFinite(value)) return false;
```
blamechris, line 401: NIT (non-blocking): `!Number.isFinite(value)` short-circuits for both NaN and +/-Infinity, so Infinity / -Infinity inputs currently stay at info severity. That's arguably inconsistent — +Infinity is strictly more extreme than 99999, and -Infinity is strictly more extreme than -5. In practice Zod validation at event intake should prevent non-finite floats from reaching this helper, and NaN-as-info is defensible (NaN magnitude is undefined). If you want to tighten without changing NaN behavior: `if (Number.isNaN(value)) return false; if (!Number.isFinite(value)) return true;` before the window check. Fine to leave as-is — the window-check is the primary signal and this is a defensibility comment, not a behavior bug.

PR #879, opened 2026-04-18
feat(notes): cosigning and amendment workflow with version history

blamechris:
## Summary

Implements the `notes.cosign`, `notes.amend`, and `notes.getVersionHistory` procedures over the previously-unused `cosigned_by` / `cosigned_at` columns and `note_versions` table. Every state-changing note operation (sign, cosign, amend) now archives an immutable version row so the audit trail has a full chronological spine.

### Cosigning
- [x] `notes.cosign` mutation added to the clinical-notes router and RBAC-enforced api-gateway router
- [x] Physician / specialist / admin role required — nurse, patient, family_caregiver all rejected with FORBIDDEN
- [x] Only `signed` notes can be cosigned; draft, already-cosigned, and amended notes raise `NoteStateError` → CONFLICT
- [x] Records cosigner id (always `ctx.user.id`, never client-supplied) and ISO 8601 `cosigned_at`
- [x] Status transitions `signed` → `cosigned`
- [x] Emits a `note.cosigned` clinical event and a structured audit_log row (action=cosign, resource_type=clinical_note)

### Amendments
- [x] `notes.amend` mutation creates a new version with updated sections and snapshots the pre-amendment state in `note_versions`
- [x] Amendment reason required (non-empty, trimmed, ≤ 2000 chars) — Zod validation rejects empty and whitespace-only values
- [x] Status transitions `signed` / `cosigned` / `amended` → `amended`; draft notes refused
- [x] Version number bumps on every amendment so the history chain is monotonically ordered
- [x] Emits a `note.amended` clinical event carrying `{ resourceId, amendedBy, reason, previousVersion, newVersion }` and writes an explicit audit_log row with the reason for HIPAA amendment-trail semantics

### Version history
- [x] `notes.getVersionHistory` query returns all versions for a note ordered by version ascending
- [x] Version row written on every state-change (sign, cosign, amend) in addition to the existing `updateNote` archive path
- [x] `note_versions` rows are insert-only at the application layer — no code path issues UPDATE / DELETE

### Non-goals (explicitly out of scope)
- [x] No schema migration — tables and columns already existed
- [x] No multi-party cosign chains (single cosigner per spec)
- [x] No amendment path for drafts — `updateNote` remains the draft edit path
- [x] No notifications on cosign/amend (follow-up epic)

## Test plan

- [x] `pnpm --filter @carebridge/validators test` — 37 tests pass including new cosign/amend schema cases
- [x] `pnpm --filter @carebridge/clinical-notes test` — 41 tests pass (was 25), covering cosign/amend positive, cosign wrong-state (draft, already-cosigned, amended), amend wrong-state (draft), amend with empty reason, version archival on sign/cosign/amend, `getVersionHistory` ordering
- [x] `pnpm --filter @carebridge/api-gateway test` — 276 tests pass (22 files) including 22 clinical-notes RBAC cases for sign/cosign/amend/getVersionHistory
- [x] `pnpm turbo typecheck` — all workspace packages type-clean

### DIFF
--- a/packages/db-schema/drizzle/0036_note_versions_lifecycle_event.sql
+++ b/packages/db-schema/drizzle/0036_note_versions_lifecycle_event.sql
@@ -1,0 +1,23 @@
+-- Add `lifecycle_event` column to note_versions so the version history can
+-- distinguish archive rows that share the same `version` number.
+--
+-- Background: signNote and cosignNote both call archiveVersion() with the
+-- existing `clinical_notes.version` (they do not bump it — only amend does),
+-- so a create -> sign -> cosign sequence produces two note_versions rows that
+-- are identical except for saved_at/saved_by. `getVersionHistory` ordered by
+-- `version` alone returned them in an unstable order with no way to label
+-- which snapshot represented which state transition.
+--
+-- Column values used by application code:
+--   "draft"    — draft edit archived by updateNote
+--   "signed"   — archived at sign time by signNote
+--   "cosigned" — archived at cosign time by cosignNote
+--   "amended"  — archived at amend time by amendNote
+--   "unknown"  — backfill default for any rows inserted before this column
+--                existed; new rows always receive an explicit event.
+--
+-- Non-destructive: column is NOT NULL with a DEFAULT, so existing rows are
+-- backfilled atomically. No separate data-migration script is needed.
+
+ALTER TABLE "note_versions"
+  ADD COLUMN IF NOT EXISTS "lifecycle_event" text NOT NULL DEFAULT 'unknown';
--- a/packages/db-schema/drizzle/meta/_journal.json
+++ b/packages/db-schema/drizzle/meta/_journal.json
@@ -253,6 +253,13 @@
       "when": 1746720000000,
       "tag": "0033_review_jobs_trigger_event_index",
       "breakpoints": true
+    },
+    {
+      "idx": 36,
+      "version": "7",
+      "when": 1746806400000,
+      "tag": "0036_note_versions_lifecycle_event",
+      "breakpoints": true
     }
   ]
 }
--- a/packages/db-schema/src/schema/notes.ts
+++ b/packages/db-schema/src/schema/notes.ts
@@ -36,6 +36,15 @@ export const noteVersions = pgTable("note_versions", {
   sections: encryptedSections("sections").notNull(),
   saved_at: text("saved_at").notNull(),
   saved_by: text("saved_by").notNull(),
+  // Labels which state transition produced this archive row. Distinguishes
+  // otherwise-identical snapshots taken at the same `version` number —
+  // specifically sign/cosign both archive at `existing.version` without
+  // bumping it, so without this column a create→sign→cosign trail cannot
+  // be told apart in `getVersionHistory`.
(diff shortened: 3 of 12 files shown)


blamechris:
## Agent review — changes requested (clinical-workflow rigor)

Strong PR overall: state machine is correctly guarded, spoofing defense at the gateway is tight, `note_versions` is insert-only, role gating on cosign/amend matches `sign`, every new procedure runs `enforcePatientAccess`, amendment-reason validation covers empty/whitespace/oversize, and the clinical-event union is extended cleanly. Amending a cosigned note also preserves `cosigned_by` / `cosigned_at` on the live row — clinically correct.

One blocking concern:

### BLOCKING — `note_versions` collisions break the chronological spine

`signNote` and `cosignNote` do not increment `clinical_notes.version`, but `archiveVersion()` is called with `existing.version` in both paths plus the `amendNote` path (which archives pre-increment). So a realistic lifecycle:

- create (v1 live) — no archive
- sign (v1 live) — archive row at `version=1`
- cosign (v1 live) — archive row at `version=1`
- amend -> v2 live — archive row at `version=1`
- amend -> v3 live — archive row at `version=2`

`getVersionHistory` does `orderBy(asc(noteVersions.version))`, so three snapshots labeled v1 come back in an unstable order. There is no `state` / `event_type` column on `note_versions` and no secondary sort on `saved_at`, so the caller cannot distinguish the sign-time snapshot from the cosign-time snapshot from the first amend's pre-snapshot. For a PR whose body advertises \"a full chronological spine\" used for \"audit timelines\", this is a correctness gap, not just a cosmetic one.

Minimum fix options:
1. Add a secondary `orderBy(asc(noteVersions.saved_at))` so duplicate-version rows at least sort chronologically. (Cheap; doesn't need a migration.)
2. Also record the lifecycle event (`sign | cosign | amend | update`) on the archive row — requires a migration but gives the audit timeline its semantic labels.

At least option 1 should land before merge; option 2 is worth following up as its own issue.

### NITs (filed separately, non-blocking)

- `services/clinical-notes/src/router.ts` cosign/amend procedures accept `cosigned_by` / `amended_by` as client input via `.extend(...)`. This router isn't currently HTTP-mounted, but it's inconsistent with the hardened pattern used by `sign` (schema's `signed_by` deliberately ignored at the gateway). If this router is ever mounted directly, spoofing becomes possible.
- Gateway `amend` audit-row `details` captures `new_version` but not `old_status` -> `new_status`. Amend's source state is variable (signed | cosigned | amended) and clinically meaningful; recording it would make the HIPAA amendment trail self-contained.
- `ip_address: \"\"` in the explicit audit inserts — the Fastify hook captures the real IP, but the structured row has a blank. Either drop the field or pull it from the tRPC meta / Fastify request.
- Consider archiving the initial v1 on note creation so the history table reflects the note's birth. (Ties to the spine-ordering concern above.)

### Praise

- Separate `NoteStateError` vs `NoteConflictError` -> clean CONFLICT mapping at the gateway without losing the optimistic-lock semantics.
- `archiveVersion` helper keeps the version-row shape consistent across sign/cosign/amend.
- Test coverage of the wrong-state paths (cosign draft/cosigned/amended, amend draft) is thorough.
- Cosigner trail preservation on amend (cosigned_at / cosigned_by not zeroed) is the right clinical call.

Issue #331, opened 2026-04-10
feat(portal): implement clinician daily/weekly schedule view


blamechris:
## Context

Clinicians need to see their daily schedule at a glance — who's coming in, when, and why.

## Scope

`apps/clinician-portal`

## Requirements

- New `/schedule` page with daily and weekly view toggle
- Daily view: time-slot grid showing appointments with patient name, type, reason, status
- Weekly view: condensed calendar showing appointment density per day
- Click appointment to see patient details (link to patient chart)
- Current day highlighted, ability to navigate between dates
- Show schedule blocks (vacation, meetings) as unavailable slots

## Acceptance Criteria

- [ ] `/schedule` page renders daily view by default
- [ ] Appointments shown in correct time slots with patient info
- [ ] Weekly view shows appointment distribution
- [ ] Navigation between days/weeks works
- [ ] Schedule blocks shown as unavailable
- [ ] `pnpm typecheck && pnpm lint` passes

## Dependencies

Depends on: scheduling service (#330)

blamechris:
Implemented in PR #354 (merged). Closing — the PR body used `Refs #331` rather than `Closes #331`, so GitHub never auto-closed this. Verified via code audit on current main.

Issue #182, opened 2026-04-07
feat(portal): add skip navigation links and visible focus indicators — WCAG 2.1 AA


blamechris:
## Problem
Both portals fail two critical WCAG 2.1 AA criteria:

1. **2.4.1 Bypass Blocks** — No skip navigation link exists in either `apps/clinician-portal/app/layout.tsx` or `apps/patient-portal/app/layout.tsx`. Keyboard users must tab through the entire sidebar on every page before reaching main content.

2. **2.4.7 Focus Visible** — No `:focus-visible` CSS is applied to interactive elements (buttons, links, inputs) beyond the search input. Keyboard users cannot see where focus is.

### Skip nav link (both layouts):
```jsx
<a href="#main-content" className="skip-link">Skip to main content</a>
// ...
<main id="main-content">
```

```css
.skip-link {
  position: absolute;
  top: -40px;
  left: 0;
  z-index: 100;
}
.skip-link:focus { top: 0; }
```

### Focus indicators (global CSS):
```css
:focus-visible {
  outline: 2px solid var(--accent);
  outline-offset: 2px;
}
```

**Severity: CRITICAL — WCAG 2.1 AA failure**

PR #890, opened 2026-04-18
feat(portal): patient appointment booking, cancel, and reschedule

blamechris:
## Summary
Adds the patient-facing appointment UI that was missing after scheduling service (#330) and clinician schedule view (#331) shipped. A new `/appointments` page gives patients tabs for upcoming + past visits, a multi-step book wizard, and a cancel/reschedule flow with prep-instruction detail views.

## Acceptance criteria
- [x] `/appointments` page in patient-portal with Upcoming + Past tabs
- [x] Upcoming row shows date/time, provider, type, location (or "Telehealth"), status badge, Cancel + Reschedule buttons
- [x] Past row shows date/time, provider, type, status, no actions
- [x] Book flow: provider (from care team) -> type -> slot -> confirm
- [x] Cancel flow: required non-empty reason -> confirm
- [x] Reschedule: cancel (with reason) then auto-open book modal
- [x] Appointment detail view with prep instructions
- [x] Dashboard card linking to `/appointments`
- [x] TDD: 33 tests for list rendering, book wizard, cancel validation, reschedule flow — all passing
- [x] `pnpm --filter @carebridge/patient-portal test` green
- [x] `pnpm turbo typecheck lint --filter=@carebridge/patient-portal` green
- [x] No clinician-portal changes, no scheduling backend changes

## Prep-instructions source
Prep instructions are currently a **static map** keyed by appointment type in `src/components/appointments/prep-instructions.ts` (four types: follow_up, new_patient, procedure, telehealth). This is flagged inline as a TODO — a follow-up issue should move this to a server-side content table so clinicians can maintain per-type prep text without a frontend deploy.

## Server-side gaps flagged for follow-up
The issue brief asked us to flag gaps rather than paper over them client-side. Two gaps observed in the existing scheduling procedures:

1. **No server-side care-team check on book.** `scheduling.appointments.create` lets any authenticated patient book with any provider. The UI correctly restricts the provider picker to the patient's care team (via `patients.careTeam.getByPatient`), but a malicious or buggy client could bypass this. **Recommend a new issue** to add a care-team membership assertion inside the `create` procedure.
2. **Reschedule is not atomic.** `scheduling.appointments.reschedule` exists on the raw service router but is **not** exposed by the RBAC router (`services/api-gateway/src/routers/scheduling.ts`), and even in the raw service it cancels the original appointment *before* checking for conflicts on the new slot. The client-side reschedule flow in this PR therefore does cancel-then-book as two separate RBAC-checked calls. **Recommend a new issue** to expose a single atomic `reschedule` RBAC procedure that performs check-new-slot -> insert-new -> cancel-old in a transaction.

Neither was fixed in this PR to stay scoped to the patient UI (as the issue asked). Double-booking prevention on fresh `create` is already correctly enforced by the existing transaction in the create proc, so the UI relies on that rather than client-side locking.

## Diff size
13 files, 1173 insertions. Over the 900 LOC guideline — the scope (book wizard + cancel modal + reschedule coordinator + detail view + prep-instructions map + TDD suite) made further trimming counter-productive without dropping tests or accessibility features. Components already share a `styles.tsx` helper and the coordinator pattern keeps the page shell small.

## Test plan
- [ ] `pnpm --filter @carebridge/patient-portal test` — 33 tests pass
- [ ] Manual: log in as patient@carebridge.dev, click Appointments on the dashboard
- [ ] Manual: book a new appointment with a care-team provider, verify it appears in Upcoming
- [ ] Manual: cancel an upcoming appointment — verify reason is required and the row moves to Past as Cancelled
- [ ] Manual: reschedule an appointment — verify cancel reason is captured and the book modal auto-opens afterward
- [ ] Manual: click View details on any appointment — verify prep instructions match the appointment type
- [ ] Manual: verify Telehealth appointments show "Telehealth" as the location label

### DIFF
--- a/apps/patient-portal/app/appointments/page.tsx
+++ b/apps/patient-portal/app/appointments/page.tsx
@@ -1,0 +1,112 @@
+"use client";
+
+import { useCallback } from "react";
+import { useRouter } from "next/navigation";
+import { useAuth } from "@/lib/auth";
+import { trpc } from "@/lib/trpc";
+import { useMyPatientRecord } from "@/lib/use-my-patient";
+import { AppointmentsPageInner } from "@/components/appointments/appointments-page-inner";
+import type { CareTeamProvider, SlotOption, BookPayload } from "@/components/appointments/book-appointment-modal";
+import type { AppointmentRow } from "@/components/appointments/appointment-list";
+
+const capitalize = (s: string) => (s ? s.charAt(0).toUpperCase() + s.slice(1) : s);
+
+export default function AppointmentsPage() {
+  const { user, hydrated, isAuthenticated } = useAuth();
+  const router = useRouter();
+  const utils = trpc.useUtils();
+
+  const { patient: myRecord, isLoading: patientLoading, isUnlinked } = useMyPatientRecord();
+
+  const appointmentsQuery = trpc.scheduling.appointments.listByPatient.useQuery(
+    { patientId: myRecord?.id ?? "" }, { enabled: !!myRecord },
+  );
+  const careTeamQuery = trpc.patients.careTeam.getByPatient.useQuery(
+    { patientId: myRecord?.id ?? "" }, { enabled: !!myRecord },
+  );
+
+  const bookMutation = trpc.scheduling.appointments.create.useMutation({
+    onSuccess: () => utils.scheduling.appointments.listByPatient.invalidate(),
+  });
+  const cancelMutation = trpc.scheduling.appointments.cancel.useMutation({
+    onSuccess: () => utils.scheduling.appointments.listByPatient.invalidate(),
+  });
+
+  const loadSlots = useCallback(async (a: { providerId: string; date: string }): Promise<SlotOption[]> => {
+    return (await utils.scheduling.schedule.availability.fetch(a)).slots;
+  }, [utils]);
+
+  const handleBook = useCallback(
+    async (a: BookPayload & { patientId: string }) => { await bookMutation.mutateAsync(a); },
+    [bookMutation],
+  );
+  const handleCancel = useCallback(
+    async (a: { appointmentId: string; reason: string }) => { await cancelMutation.mutateAsync(a); },
+    [cancelMutation],
+  );
+
+  if (!hydrated) return <p style={{ color: "#999" }}>Loading...</p>;
+  if (!isAuthenticated) { router.replace("/login"); return null; }
+  if (patientLoading) return <p style={{ color: "#999" }}>Loading your record...</p>;
+  if (isUnlinked || !myRecord) {
+    return (
+      <div role="alert" style={{
+        backgroundColor: "#7f1d1d", border: "1px solid #ef4444", borderRadius: 8,
+        padding: "1.25rem", color: "#fca5a5",
+      }}>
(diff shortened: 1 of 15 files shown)


blamechris (commented):
## Verdict: CHANGES REQUESTED (cannot formally request-changes on own PR — posting as review comment with the same force)

Strong implementation on the patient UI itself — wizard state machine, coordinator pattern, 33 tests asserting real behavior (not just render), Zero-Attribution clean. The reason this is blocked is one of the two "server-side gaps" flagged in the PR body turns out to be more serious than the description suggests.

### BLOCKER — Issue #891

`scheduling.appointments.create` in `services/api-gateway/src/routers/scheduling.ts` is NOT just "no care-team check." It has no patient-ownership check at all. Any authenticated user can call the raw tRPC with an arbitrary `patientId` and a `providerId` of their choosing — i.e. write an appointment row into a different patient's chart (PHI write on another patient's record) AND/OR book any provider regardless of care-team membership. The neighboring `listByPatient` in the same file (lines 30-48) gets this right — the pattern was intentional but omitted on the mutation. `cancel` has the same flaw (any appointment ID → status=cancelled, no ownership check).

This PR ships the UI that relies on these procs without those checks, so merging ships the exploit path. Filed as **#891 (p1, security)**. That fix plus this PR should merge together — or this PR should be held until #891 is fixed.

### FUNCTIONAL — Issue #892 (non-blocking)

Client-side cancel-then-book for reschedule has a loss-of-slot race window on disconnect, and the existing raw `scheduling.reschedule` in `services/scheduling/src/router.ts` isn't atomic and isn't even exposed via RBAC. Filed #892 (p2 enhancement) to expose an atomic transaction-wrapped `reschedule` RBAC proc. Not blocking this PR.

### NITs (non-blocking, filed as issues)

- **#893** — `cancel.reason` is `z.string()` on both routers; empty strings pass server-side. Client validation correctly rejects empty reasons but the server should match so non-UI callers can't defeat the requirement.
- **#894** — Modals set `role="dialog"` + `aria-modal` + `aria-labelledby` correctly, but there's no initial focus on open, no focus trap, and no Escape-to-close. To meet the #182/#183 a11y bar please add a shared dialog hook.
- **#895** — `AppointmentType` literal union is duplicated between `prep-instructions.ts` and both router Zod enums. Derive from a shared `@carebridge/types` const so a new type can't drift across the stack.

### Praise

- Coordinator component (`appointments-page-inner.tsx`) injects data + mutations as props, which is why the 116-line reschedule-flow test can exercise the full cancel→book transition without mocking tRPC. Clean separation.
- `isUpcoming` correctly classifies cancelled/completed/no_show as past regardless of date, and the list test asserts that specifically.
- Double-booking IS server-enforced via the transaction in `create` — UI only presents availability, server is authoritative. Verified `listAvailableSlots` filters on `ne(appointments.status, "cancelled")` + time overlap; no client-side filter.
- Telehealth location label override is test-covered and consistent between `appointment-list.tsx` and `appointment-detail-view.tsx`.
- Diff size at 1173 LOC is justified — wizard + cancel modal + detail view + coordinator + shared styles + 4 test files with real assertions. Fine.

### Note

Build check is failing on CI (unrelated to security concerns). Resolve before merge.

**Please address #891 (merge together or land first), then this PR is ready to go.**

PR #340, opened 2026-04-10
fix(notifications): encrypt notification payloads at rest (#318)

blamechris:
## Summary

- Encrypt notification `title` and `body` columns at rest using `encryptedText` (AES-256-GCM)
- Add `summary_safe` column for PHI-free text suitable for lock-screen/push display
- Migration 0016 adds the new column
- Transparent decrypt via Drizzle custom type — no changes to read paths

Addresses #285 (unencrypted notification storage) and #289 (lock-screen PHI disclosure)

Refs #318

## Test Plan

- [ ] pnpm typecheck passes
- [ ] Migration runs without error
- [ ] New notifications stored with encrypted title/body
- [ ] getByUser returns decrypted values transparently
- [ ] summary_safe populated on new notifications

PR #339, opened 2026-04-10
feat(notifications): implement BullMQ dispatch worker (#314)

blamechris:
## Summary

- Add BullMQ notification dispatch worker consuming from `notifications` queue
- Worker finds care team recipients by matching specialty to flag's `notify_specialties`
- Creates in-app notification records with proper linking to clinical flags
- DLQ pattern for failed jobs (3 retries with exponential backoff)
- Health check endpoint (port 4002) with Redis connectivity probe
- Wire `ai-oversight` flag creation to emit notification events automatically

Refs #314

## Test Plan

- [ ] pnpm typecheck passes
- [ ] Worker starts and connects to Redis
- [ ] Creating a clinical flag emits a notification job
- [ ] Notification records created for matching specialty providers
- [ ] Fallback to all care team members when no specialty match
- [ ] DLQ receives exhausted jobs

PR #897, opened 2026-04-18
feat(scheduling): configurable appointment reminder notifications (closes #333)

### LINKED ISSUE #333: feat(scheduling): implement configurable appointment reminder notifications
## Context

Missed appointments waste clinical resources and delay patient care. Automated reminders reduce no-show rates.

## Scope

`services/scheduling` + `services/notifications`

## Requirements

- Configurable reminder schedule per appointment (default: 24h and 2h before)
- BullMQ delayed jobs created when appointment is booked
- Reminder cancellation when appointment is cancelled
- Reminder content includes: date, time, provider name, location, preparation instructions
- Integrates with notification delivery system for channel routing

## Acceptance Criteria

- [ ] Reminder jobs scheduled when appointment created
- [ ] Reminders fire at correct times (24h and 2h before)
- [ ] Cancelled appointments have their reminders removed
- [ ] Reminder content is informative and PHI-safe for push notifications
- [ ] `pnpm typecheck && pnpm lint` passes

## Dependencies

Depends on: scheduling service (#330), notification dispatch worker

blamechris:
## Summary

Implements issue #333 — configurable appointment reminder notifications. When an appointment is booked, two delayed BullMQ jobs (24 h and 2 h before start) are queued on a new `appointment-reminders` queue. When they fire, the worker emits a `NotificationEvent` through the existing notifications dispatch chain (PRs #339 / #340), which owns DB persistence (PHI-encrypted body), preference evaluation, and SSE. Cancelling an appointment removes both queued jobs best-effort.

## Acceptance criteria

- [x] Configurable reminder schedule per appointment (default: 24 h and 2 h before)
- [x] BullMQ **delayed** jobs created when appointment is booked
- [x] Reminder cancellation when appointment is cancelled
- [x] Reminder content: date, time, provider name, location, prep instructions
- [x] PHI-safe push/lock-screen payload via the existing `summary_safe` pattern
- [x] Integrates with the notifications dispatch system (no duplicate delivery logic)

## Architecture decisions

Per the issue prompt, picked the simpler of the two options for each lever:

- **Option A** — separate `appointment-reminders` queue + dedicated worker rather than mixing delayed jobs into the instant `notifications` queue. Operators can observe/tune the two workloads independently.
- **Option X** — two nullable `reminder_24h_job_id` / `reminder_2h_job_id` columns on `appointments` rather than a separate `appointment_reminders` table. Two fixed offsets per appointment → no need for a row-per-offset schema.

## Safety rigor

- `suggested_action` (the safe-summary carrier on the reminder event) is a static template with no digits, no provider name, no times. The dispatch-worker further renders the lock-screen label from a whitelisted category (`patient-reported`), not from the event summary.
- `scheduleReminders` failure is non-fatal — wrapped in try/catch with `logger.warn`, so a Redis blip cannot block an appointment booking.
- `cancelReminders` is best-effort — null/undefined ids are no-ops; `getJob`/`remove` errors are swallowed. The worker also double-checks `appointment.status === "cancelled"` before emitting, covering the race where a job already entered the active state.
- Deleted appointments (worker finds no row) are logged at `warn` and skipped, not thrown.
- `scheduleReminders` skips offsets with non-positive delay — past-dated appointments, or appointments closer than the offset window, produce null job IDs (asserted in tests).

## Architecture

```
scheduling.appointments.create ─► scheduleReminders(appointment)
                                     │
                                     ▼
                           appointment-reminders queue (delayed)
                                     │ 24h / 2h later
                                     ▼
                           reminder-worker ─► emitNotificationEvent
                                                        │
                                                        ▼
                                             notifications dispatch chain
                                             (DB write, encryption, SSE)
```

## Test plan

- [x] `pnpm --filter @carebridge/scheduling test` — 41 tests pass (reminders + worker + router hooks)
- [x] `pnpm --filter @carebridge/notifications test` — 63 tests still pass (no regression)
- [x] `pnpm typecheck` — all 40 tasks green
- [x] `pnpm lint` — clean (only pre-existing portal warnings)
- [ ] Manual: boot docker-compose, book an appointment 48 h out, observe BullMQ job IDs on the row, cancel, observe jobs removed
- [ ] Manual: boot docker-compose, book an appointment 90 min out, observe no reminder scheduled (both delays negative)

## Non-goals (explicitly out of scope per issue #333)

- No per-user reminder preferences (#316 exists; appointment-reminder preference wiring deferred)
- No SMS/email channels — in-app only via existing dispatch
- No custom reminder intervals — offsets hardcoded at 24 h and 2 h
- No notifications chain refactor

### DIFF
--- a/packages/db-schema/drizzle/0037_appointment_reminder_job_ids.sql
+++ b/packages/db-schema/drizzle/0037_appointment_reminder_job_ids.sql
@@ -1,0 +1,21 @@
+-- Issue #333: configurable appointment reminder notifications.
+--
+-- Adds two nullable columns to `appointments` that hold the BullMQ job IDs
+-- for the 24h-before and 2h-before reminder jobs. The IDs are written by
+-- `services/scheduling/src/reminders.ts :: scheduleReminders` immediately
+-- after an appointment row is committed, and read again by the cancel path
+-- so it can call `job.remove()` to prevent a reminder from firing for a
+-- cancelled appointment.
+--
+-- Two fixed offsets are hardcoded (see issue #333 non-goals — no custom
+-- intervals yet), so a separate `appointment_reminders` table would be
+-- over-engineered for two rows per appointment.
+--
+-- Non-destructive: nullable columns, no default, no backfill needed. Existing
+-- rows simply have NULL reminder IDs which the cancel path treats as a no-op.
+
+ALTER TABLE "appointments"
+  ADD COLUMN IF NOT EXISTS "reminder_24h_job_id" text;
+
+ALTER TABLE "appointments"
+  ADD COLUMN IF NOT EXISTS "reminder_2h_job_id" text;
--- a/packages/db-schema/src/schema/scheduling.ts
+++ b/packages/db-schema/src/schema/scheduling.ts
@@ -24,6 +24,12 @@ export const appointments = pgTable("appointments", {
   cancelled_at: text("cancelled_at"),
   cancelled_by: text("cancelled_by"),
   cancel_reason: text("cancel_reason"),
+  // BullMQ job IDs for scheduled reminder delivery. Nullable — populated when
+  // a future-dated appointment is booked and cleared when the appointment is
+  // cancelled. Two fixed reminder offsets (24 h and 2 h before start_time)
+  // are hardcoded per issue #333. See services/scheduling/src/reminders.ts.
+  reminder_24h_job_id: text("reminder_24h_job_id"),
+  reminder_2h_job_id: text("reminder_2h_job_id"),
   created_at: text("created_at").notNull(),
   updated_at: text("updated_at").notNull(),
 }, (table) => [
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -64,7 +64,7 @@ importers:
         version: 19.2.5(react@19.2.5)
       recharts:
         specifier: ^3.8.1
-        version: 3.8.1(@types/react@19.2.14)(react-dom@19.2.5(react@19.2.5))(react-is@16.13.1)(react@19.2.5)(redux@5.0.1)
+        version: 3.8.1(@types/react@19.2.14)(react-dom@19.2.5(react@19.2.5))(react-is@17.0.2)(react@19.2.5)(redux@5.0.1)
       zod:
         specifier: ^3.24.0
         version: 3.25.76
@@ -820,12 +820,24 @@ importers:
       '@carebridge/db-schema':
         specifier: workspace:*
         version: link:../../packages/db-schema
+      '@carebridge/logger':
+        specifier: workspace:*
+        version: link:../../packages/logger
+      '@carebridge/notifications':
+        specifier: workspace:*
+        version: link:../notifications
+      '@carebridge/redis-config':
+        specifier: workspace:*
+        version: link:../../packages/redis-config
       '@carebridge/shared-types':
         specifier: workspace:*
         version: link:../../packages/shared-types
       '@trpc/server':
         specifier: ^11.0.0
         version: 11.16.0(typescript@5.9.3)
+      bullmq:
+        specifier: ^5.30.0
+        version: 5.74.1
       drizzle-orm:
         specifier: ^0.38.0
         version: 0.38.4(@types/react@19.2.14)(postgres@3.4.9)(react@19.2.5)
@@ -833,12 +845,18 @@ importers:
         specifier: ^3.24.0
         version: 3.25.76
     devDependencies:
+      '@carebridge/test-utils':
+        specifier: workspace:*
+        version: link:../../packages/test-utils
       '@types/node':
         specifier: ^22.0.0
         version: 22.19.17
       typescript:
         specifier: ^5.7.0
         version: 5.9.3
+      vitest:
+        specifier: ^3.0.0
+        version: 3.2.4(@types/node@22.19.17)(jsdom@29.0.2(@noble/hashes@2.0.1))(tsx@4.21.0)
 
   tooling/scripts:
     dependencies:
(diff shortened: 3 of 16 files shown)


blamechris (commented):
## Review — `/agent-review` on PR #897 (VERDICT: changes-requested — posted as comment because GH blocks self-review state changes)

Thanks for the thorough TDD and the clear issue-decision narrative in the PR body. The scheduling/worker layer itself (queue, delay math, cancel idempotency, PHI-free `suggested_action` template) is well-built and the 41 tests cover the happy path and most edge cases I wanted to see. Two blockers, though, both at the integration seam with the existing notifications dispatch chain.

### BLOCKING

**B1 — Recipient routing delivers patient reminders to the patient's care team, not the patient.**

`emitNotificationEvent` → `notifications` queue → `dispatch-worker.processNotificationJob` → `findNotificationRecipients(event.patient_id, …)`. That function resolves recipients from `careTeamAssignments` (`services/notifications/src/workers/dispatch-worker.ts:123-160`), i.e. providers assigned to the patient. There is no "deliver to the patient" path in the current dispatch chain.

Consequences:
1. The patient never receives their own reminder — the core AC of issue #333 ("Reminders fire at correct times" to the recipient) is unmet.
2. The full `event.summary` is persisted on `notifications.body` (encrypted at rest) for every clinician on the care team and then rendered in the authenticated portal fetch. That means provider name + time + reason for a patient's appointment leaks to the whole care team on each reminder fire — information they may already have access to, but not something we should broadcast via a reminder system designed to address patients.

None of the 41 tests catch this because the router-level tests mock `scheduleReminders`/`cancelReminders` and the worker-level tests mock `emitNotificationEvent` — nothing exercises the actual dispatch.

Options:
- Add a patient-recipient branch to `dispatch-worker` keyed on `source === "scheduling.reminder"` (or a new `audience: "patient" | "care-team"` field on `NotificationEvent`) so the dispatch worker treats `patient_id` as the recipient `user_id` rather than a lookup key.
- Or skip `emitNotificationEvent` and write a direct-to-patient notification path.

This needs resolution before merge.

**B2 — Notification title/label renders reminders as clinical flags.**

`buildNotificationTitle` (dispatch-worker.ts:183) always produces `"<severity>: Clinical flag — <categoryLabel>"`. With `category: "patient-reported"` + `severity: "info"` the recipient (per B1, the care team) sees:

> **Info: Clinical flag — Patient-reported concern**
> Clinical flag — Patient-reported concern. Open the portal to view details.

For an appointment reminder this is misleading on both axes. A reminder isn't a "clinical flag" and it isn't a "patient-reported concern" either (it's system-initiated). Fixing B1 with a new recipient branch is the natural place to also branch on title/summary rendering.

### NITS (non-blocking)

**N1 — `SAFE_SUMMARY` constant in `reminder-worker.ts` is dead code.** The PR description, test assertions (`suggested_action` PHI-safety block), and inline comment all imply `suggested_action` carries the lock-screen payload. It doesn't — `dispatch-worker.buildSafeSummary(categoryLabel)` derives `summary_safe` purely from the whitelisted category label and ignores `event.suggested_action` entirely. The net lock-screen string is still PHI-free by construction (good), but the code + tests + PR copy all mislead a future reader about where that safety comes from. Either wire `suggested_action` into `buildSafeSummary` (with a whitelist check) or drop the dead constant and tighten the tests.

**N2 — Category choice.** Prompt explicitly permits this as a nit: `"patient-reported"` is reused from the clinical-flag taxonomy (`packages/shared-types/ai-flags.ts`) where it means "concern reported *by* the patient". An appointment reminder is system-initiated. Adding an `"appointment-reminder": "Appointment reminder"` entry to `CATEGORY_LABELS` would produce the accurate lock-screen label `"Clinical flag — Appointment reminder. Open the portal to view details."` — still PHI-free by the same whitelist contract. (Separate typing issue: `FlagCategory` would need the new value too.)

**N3 — Reschedule path is not transactional.** The `reschedule` hook does cancel-old → `db.insert(newAppt)` → `scheduleReminders(newAppt)` → `db.update(... job_ids)` as four separate DB calls. If the `db.insert` fails after `cancelReminders` succeeds, the old reminders are already gone and the new appointment row never exists. `create` uses `db.transaction`; `reschedule` should too. Pre-existing style, but reminders widen the blast radius.

**N4 — `scheduleReminders(now)` default.** `scheduleReminders(appointment)` defaults `now = new Date()` when the router calls it. Tests pass an explicit `now`, so they're deterministic; production is not. Accepting this for a non-critical reminder system, but a frozen-clock contract would let you assert past-dated edge cases in an integration test too.

### PRAISE

- **P1 — TDD coverage.** 41 tests is proportionate; the reminders.test delay-math cases (past, <2h, <24h, exactly-at-offset, invalid ISO, defensive null-id) are precisely the edge cases I'd have asked you to add.
- **P2 — Best-effort non-blocking hooks.** `try/catch + log.warn` around both schedule and cancel paths is the right call for a nice-to-have reminder, and the router-reminders.test `"does not block booking when scheduleReminders throws"` case nails the acceptance check.
- **P3 — Migration is cleanly reversible.** Two nullable columns, no backfill needed, `IF NOT EXISTS` idempotent ALTERs, doc comment pointing back at `reminders.ts`. Good schema hygiene.
- **P4 — Double-check for cancelled status in the worker.** The belt-and-suspenders check in `processReminderJob` covers the race where `cancelReminders` raced with a job entering the active state. Nice.
- **P5 — Redis blip isolation.** `cancelReminders` swallowing `getJob`/`remove` errors is exactly the right posture for a cancel path that must never block a user-initiated cancel.

### Verdict: changes-requested

Fix B1 + B2 (they're the same integration seam). N1 is cheap cleanup while you're there. N2/N3/N4 can follow-up. Happy to re-review once the dispatch path supports the patient as a first-class recipient.

Issue #896, opened 2026-04-18
Wire family_relationships.access_scopes into enforcePatientAccess


blamechris:
Follow-up from #889 review.

## Context

`family_relationships.access_scopes` is a `jsonb` column populated when a patient grants access (tokens: `read_only`, `view_summary`, `view_appointments`, `view_medications`, `view_labs`, `view_notes`, `view_and_message`). See `packages/db-schema/src/schema/family-access.ts:60-69`.

As of #889 a caregiver with any active relationship gets all-or-nothing read access to the patient record. The granular scope tokens the patient selected during the invite flow are currently ignored.

## Problem

This is a latent security/privacy gap:
- Patients who invite a caregiver with `view_appointments` only currently get the same effective permissions as one invited with `view_and_message`.
- The patient-perceived control (\"I only shared labs, not notes\") is not actually enforced.

## Proposed fix

1. In `enforcePatientAccess` (services/api-gateway/src/routers/patient-records.ts + clinical-data.ts), extend the caregiver branch to also select `access_scopes` from the relationship row.
2. Add per-resource scope predicates:
   - `diagnoses.getByPatient` / `allergies.getByPatient` / `patients.getById` → require `view_summary` (or `read_only` as a blanket permit).
   - `observations.getByPatient` → require `view_summary`.
   - `medications.*` read → require `view_medications`.
   - `labs.*` read → require `view_labs`.
   - `notes.*` read → require `view_notes`.
   - `messages.*` → require `view_and_message`.
3. Update `getMyPatients` to surface the scope set so the UI can optionally grey out sections the caregiver lacks access to.
4. Tests: matrix test covering each scope token against each resource.

## Acceptance criteria

- [ ] Caregiver with `view_summary` only can see `patients.getById` but NOT `medications.getByPatient`.
- [ ] Caregiver with `view_and_message` can read everything readable.
- [ ] Denials come back as `FORBIDDEN` with a message naming the missing scope (without leaking PHI).
- [ ] `getMyPatients` returns the scope set alongside id/name/mrn/relationship.

## Related

- #889 (introduced the all-or-nothing caregiver read path)
- #305-313 (original Phase B3 scope work)

PR #910, opened 2026-04-19
feat(gateway): enforce view_and_message scope on messaging router

### LINKED ISSUE #909: enhancement(gateway): wire view_and_message scope into messaging router
## Context

Follow-up nit on #900. The issue body for #896 explicitly lists:

> \`messages.*\` → require \`view_and_message\`

and the PR body documents \`view_and_message\` as a superset that \"grants every read scope above **plus messaging participation**\". However, #900 does not touch \`services/api-gateway/src/routers/messaging.ts\` — messaging procedures still only gate via \`assertConversationAccess\` (participant membership), and there is no scope-level gate.

## Why this matters

The scope system is patient-delegated: a patient who invites a caregiver with \`view_summary\` only (or \`read_only\`) is telling the system \"I do not want this caregiver to read or send clinical messages on my behalf.\" Today, once a caregiver is added as a conversation participant, they can read and reply in the conversation regardless of scope. That contradicts the patient's expressed intent.

## Reality check

It is possible the current intent is \"conversation participant list is the authoritative gate, scopes do not apply\" — in which case the issue body / PR body should be updated to remove the \`view_and_message → messages.*\` mapping claim, so future readers do not think the control exists. Either way, the docs and the code should agree.

## Suggested fix

Two reasonable options; pick one:

1. **Wire the scope.** In \`messagingRbacRouter\`, resolve the patient the conversation is about (join via \`conversations.patient_id\` or the participant list), call the shared \`enforcePatientAccess\` helper (or a messaging-local variant) with \`requiredScope: \"view_and_message\"\` on \`listMessages\`, \`sendMessage\`, and any other per-conversation procedures when the caller is a family_caregiver. Admin / patient / clinician branches continue to bypass.

2. **Document the carve-out.** Update the resource→scope table in the PR description and in the \`ScopeToken\` JSDoc to say that \`view_and_message\`'s messaging semantics apply only to conversation *participant addition* (whichever invite flow grants it), not to per-message procedures.

Option 1 matches the spirit of the issue better — a caregiver removed from \`view_and_message\` should lose messaging access even if their participant row is still there.

## Tests

Add to the scope matrix or a dedicated messaging test:
- caregiver with \`view_summary\` is a participant → listMessages FORBIDDEN naming \`view_and_message\`.
- caregiver with \`view_and_message\` → succeeds.
- patient sending to their own conversation → bypass.

## Priority

Non-blocking for #900 — the existing \`assertConversationAccess\` still prevents unrelated caregivers from reading any conversation. This is a latent control surface that did not regress, but the PR/issue claim that \`view_and_message\` gates messages is incorrect as written.

blamechris:
## Summary

Wave 7 (#896 / PR #900) wired per-resource `access_scopes` enforcement into patient-records, clinical-data, clinical-notes, and scheduling, but the `messages.* -> view_and_message` line from the spec wasn't actually wired into `messaging.ts`. Caregiver access to conversations relied only on participant membership — and caregivers are never direct participants — so scope-granted caregivers were silently denied regardless of `view_and_message`.

This PR wires the messaging router to match the other four routers.

## Changes

`services/api-gateway/src/routers/messaging.ts`
- Added `findActiveFamilyRelationship` (same join-through-users pattern as patient-records/scheduling).
- Added `assertCaregiverMessagingScope` that loads the conversation to resolve `patient_id`, verifies an active family_relationship, and requires `view_and_message` via `hasScope`/`normaliseScopes`.
- Added `assertMessagingReadAccess` dispatcher: caregivers go through the scope path, every other role stays on `assertConversationAccess` (unchanged regression behaviour).
- Applied to read procedures: `listConversations`, `getConversation`, `listMessages`, `markRead`.
- `listConversations` for caregivers enumerates patients they hold `view_and_message` on (via `family_relationships` + `users.patient_id` join) and returns every conversation for those patients.

Write-side role block (HIPAA minimum-necessary):
- `sendMessage` and `createConversation` return FORBIDDEN for `family_caregiver` regardless of scope. Mirrors the `observations.create` block — caregivers read messages but never post on behalf of the patient (authorship provenance is clinically load-bearing).

## Acceptance criteria

- [x] Caregiver with `view_and_message` can list represented patient's conversations, open a conversation, and read messages.
- [x] Caregiver without `view_and_message` (including `view_summary`, `view_notes`, `read_only`, null, empty array) gets FORBIDDEN on every read procedure.
- [x] Caregiver cannot `sendMessage` or `createConversation` — FORBIDDEN at the role check before any DB lookup.
- [x] Regression: clinician on care team (participant) still reads + writes conversations as before.
- [x] Regression: patient (participant) still reads + writes their own conversations.
- [x] FORBIDDEN messages name the missing scope and do NOT leak patient, user, or conversation identifiers.
- [x] Default-deny: missing conversation -> FORBIDDEN (not NOT_FOUND), null/empty access_scopes -> read_only default -> deny.
- [x] Diff under 500 LOC (router +204/-6; test file +526 new).

## Test plan
- [x] New `messaging-rbac.test.ts`: 21 tests covering scope grant/deny, write-side block, error-message hygiene, markRead, and participant-path regression.
- [x] `pnpm --filter @carebridge/api-gateway test` — all 447 tests pass.
- [x] `pnpm typecheck` — green.
- [x] `pnpm lint` — green.

Closes #909

### DIFF
--- a/services/api-gateway/src/__tests__/messaging-rbac.test.ts
+++ b/services/api-gateway/src/__tests__/messaging-rbac.test.ts
@@ -1,0 +1,526 @@
+/**
+ * RBAC tests for the messaging router scope enforcement (issue #909).
+ *
+ * Issue #909 wires the `view_and_message` family caregiver scope into the
+ * messaging router. Wave 7 (#896 / PR #900) already wired the other four
+ * patient-scoped resource routers but the spec line
+ * `messages.* -> view_and_message` was missed. This test file locks in the
+ * following contract:
+ *
+ *  - A family_caregiver with `view_and_message` scope on the represented
+ *    patient can list conversations, open a conversation, and read messages.
+ *  - A caregiver WITHOUT `view_and_message` (any other scope token) is
+ *    denied with FORBIDDEN — messaging is default-deny.
+ *  - Caregivers are blocked from writes (sendMessage, createConversation)
+ *    at the role level — HIPAA requires they read messages but never post
+ *    on behalf of the patient.
+ *  - Regression coverage: clinician participants and the patient themselves
+ *    continue to go through the existing participant-based path.
+ *
+ * The tests exercise `messagingRbacRouter.createCaller(ctx)` directly so the
+ * middleware + procedure code runs without an HTTP transport.
+ */
+
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import type { User, ScopeToken } from "@carebridge/shared-types";
+
+const CAREGIVER_ID = "77777777-7777-4777-8777-777777777777";
+const PATIENT_USER_ID = "11111111-1111-4111-8111-111111111111";
+const PATIENT_RECORD_ID = "aaaa1111-1111-4111-8111-111111111111";
+const OTHER_PATIENT_RECORD_ID = "bbbb2222-2222-4222-8222-222222222222";
+const CLINICIAN_ID = "22222222-2222-4222-8222-222222222222";
+const CONVERSATION_ID = "cccc3333-3333-4333-8333-333333333333";
+const OTHER_CONVERSATION_ID = "dddd4444-4444-4444-8444-444444444444";
+const MESSAGE_ID = "eeee5555-5555-4555-8555-555555555555";
+
+// ---------------------------------------------------------------------------
+// Hoisted mocks
+// ---------------------------------------------------------------------------
+
+const mocks = vi.hoisted(() => {
+  let queue: unknown[][] = [];
+
+  function nextResult(): unknown[] {
+    return queue.shift() ?? [];
+  }
+
+  function makeChain() {
+    const chain: Record<string, unknown> = {};
+    chain.from = vi.fn(() => chain);
+    chain.innerJoin = vi.fn(() => chain);
+    chain.where = vi.fn(() => {
+      const next = nextResult();
+      const wrapper: Record<string, unknown> = {
+        then: (resolve: (v: unknown) => void) => {
+          resolve(next);
+          return wrapper;
+        },
+        // orderBy returns a thenable that also exposes .limit(), so chains
+        // of the form `.where().orderBy().limit()` work.
+        orderBy: vi.fn(() => {
+          const inner: Record<string, unknown> = {
+            then: (resolve: (v: unknown) => void) => {
+              resolve(next);
+              return inner;
+            },
+            limit: vi.fn(async () => next),
+          };
+          return inner;
+        }),
+        limit: vi.fn(async () => next),
+      };
+      return wrapper;
+    });
+    chain.orderBy = vi.fn(async () => nextResult());
+    chain.limit = vi.fn(async () => nextResult());
+    return chain;
+  }
+
+  const mockDb = {
+    select: vi.fn(() => makeChain()),
+    insert: vi.fn(() => ({ values: vi.fn(async () => undefined) })),
+    update: vi.fn(() => ({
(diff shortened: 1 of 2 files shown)


blamechris (commented):
## VERDICT: APPROVE (posted as comment — cannot self-approve)

Clean, scope-aligned fix for the spec gap in #896/#900. The two-path access model (participant vs delegated-caregiver) is factored correctly through `assertMessagingReadAccess`, and the write-side caregiver block is placed before any DB lookup — no enumeration surface. All seven review criteria verified in the diff:

1. Zero Attribution — no Claude/Anthropic strings in diff.
2. Caregiver read gated by `view_and_message` via `hasScope(normaliseScopes(...))` — `assertCaregiverMessagingScope` line ~674.
3. Caregiver `sendMessage` + `createConversation` return FORBIDDEN at the role check before any DB call. Matches the `observations.create` precedent.
4. Non-caregiver regression safe — `assertMessagingReadAccess` dispatches to `assertConversationAccess` for every non-caregiver role; participant-path tests cover clinician + patient read AND write.
5. Missing conversation lookup returns FORBIDDEN with a neutral message, not NOT_FOUND — enumeration-probe safe.
6. Null + empty `access_scopes` both flow through `normaliseScopes` and deny, with explicit tests for each.
7. 21 new tests confirmed (positive + negative scope combinations, wrong-scope variants `view_summary`/`view_notes`, missing relationship, missing conversation, null, empty, write-side block, markRead, and the clinician/patient participant regressions).

## Blockers
None.

## Nits (non-blocking)
- `listConversations` caregiver branch does the `hasScope` filter in-memory after fetching all active relationships. Fine for the expected cardinality (a caregiver has few relationships) but worth noting; if a caregiver ever accrues hundreds of relationships this becomes an N over-fetch. Could be pushed into the `where` with an `access_scopes` filter in a follow-up.
- No test for the "mixed scopes across multiple patients" case — a caregiver with `view_and_message` on patient A and `view_summary` on patient B should see only A's conversations. Covered implicitly by the in-memory filter but not asserted. Low value; the per-patient read tests already prove the semantics.
- The FORBIDDEN message in `assertCaregiverMessagingScope` for "no active family relationship" doesn't name the missing scope the way the scope-denial branch does. Minor UX polish — the UI gets a specific string either way.

## Praise
- Module-level JSDoc documenting the two-path access model + HIPAA write-side policy is exactly the kind of note future maintainers need; it prevents this exact bug from recurring.
- Default-deny on missing conversation with the explicit "enumeration probe" rationale in the comment is the right call and rare to see spelled out.
- Error-message-hygiene test asserting the FORBIDDEN string does NOT contain `PATIENT_RECORD_ID` / `PATIENT_USER_ID` / `CONVERSATION_ID` is a strong PHI-leak guardrail.
- The `findActiveFamilyRelationship` helper correctly mirrors the pattern from patient-records/scheduling (join through `users` because `family_relationships.patient_id` references `users.id`, not `patients.id`) — consistency across Wave 7 routers preserved.

## Key findings
- Router change is +204/-6 and regression-surface-minimal: pre-existing `assertConversationAccess` is unchanged, only dispatched through a new helper for caregivers.
- Write-side block is placed BEFORE `assertConversationAccess` in `sendMessage`, so no DB work happens on the denied path.
- `markRead` correctly uses `assertMessagingReadAccess` (not a write-side block) — treating read-receipt as a read-state mutation is defensible and the test locks it in.
- `listConversations` caregiver path correctly uses `inArray` for the user-id and patient-id lookups, avoiding N+1.

PR #912, opened 2026-04-19
feat(scheduling): atomic reschedule + server-side validation + modal a11y + prep enum

### LINKED ISSUE #892: enhancement(scheduling): expose atomic reschedule RBAC procedure
## Context
Filed from `/agent-review` of PR #890.

## Problem
PR #890 implements reschedule client-side as cancel-then-book (two separate RBAC-checked calls). If the user is disconnected after cancel but before book, they lose their slot with no recovery path. Concurrent bookers can also grab the slot during the gap.

## Underlying gaps
1. `services/api-gateway/src/routers/scheduling.ts` does not expose a `reschedule` procedure — the RBAC router is missing it entirely.
2. Even the raw `services/scheduling/src/router.ts` `reschedule` proc (lines 124-184) cancels the original **before** checking for conflicts on the new slot, and the whole thing is not wrapped in a transaction. That design would also lose the slot on conflict.

## Fix
Expose an RBAC `scheduling.appointments.reschedule` procedure that:
1. Runs inside a single DB transaction.
2. Loads the original appointment and enforces ownership/care-team checks (same guardrails as issue #<TBD — security issue for create/cancel>).
3. Checks the new slot is free (`ne status cancelled` + overlap query).
4. Inserts the new appointment row.
5. Cancels the original with `cancel_reason = "Rescheduled"`.
6. Rolls back entirely if any step fails so the patient keeps their original slot.

Then update the patient-portal `AppointmentsPageInner` reschedule flow (apps/patient-portal/src/components/appointments/appointments-page-inner.tsx) to call the new atomic proc instead of sequential cancel + create.

## Test plan
- Atomic: reschedule fails at insert-new due to conflict — original stays `scheduled` (not cancelled).
- Happy path: original → `cancelled` with reason `Rescheduled`, new row in `scheduled`.
- Ownership: patient A cannot reschedule patient B's appointment.

### LINKED ISSUE #893: nit(scheduling): cancel reason is not validated non-empty on server
## Context
Filed from `/agent-review` of PR #890.

## Issue
PR #890 correctly validates a non-empty cancel reason on the client (\`cancel-appointment-modal.tsx\` trims + disables submit). The server proc \`scheduling.appointments.cancel\` in \`services/api-gateway/src/routers/scheduling.ts\` accepts \`reason: z.string()\` — empty string is allowed. A non-UI caller can set an empty cancel_reason, defeating the product requirement that all cancellations carry a reason for clinical follow-up.

## Fix
Change the input schema in both:
- \`services/api-gateway/src/routers/scheduling.ts\` (RBAC wrapper)
- \`services/scheduling/src/router.ts\` (raw service router)

From \`reason: z.string()\` to \`reason: z.string().trim().min(1, "Cancel reason is required")\`.

Small, can ship as part of the ownership-check security fix (#891) since those files are already being touched.

### LINKED ISSUE #894: nit(portal): appointment booking wizard focus-trap + initial focus on modal open
## Context
Filed from `/agent-review` of PR #890.

## Issue
The three new modals in PR #890 (\`BookAppointmentModal\`, \`CancelAppointmentModal\`, \`AppointmentDetailView\`) use \`role="dialog"\` + \`aria-modal="true"\` + \`aria-labelledby\` correctly, and the form inputs are properly labeled. However, to meet the WCAG 2.1 AA bar established by #182/#183:

1. **Initial focus** — none of the modals move focus to the first interactive element (heading, first radio, or reason textarea) on open. Screen-reader users hear the heading only if focus happens to be on the right element.
2. **Focus trap** — Tab can escape the modal to the underlying appointments list.
3. **Escape-to-close** — no \`keydown\` handler to close on Escape.

## Fix
Either:
- Adopt an existing a11y-safe dialog primitive in the monorepo (check if one was added for #182/#183 notifications modals), or
- Add a shared \`useModalFocus(dialogRef, onClose)\` hook in \`apps/patient-portal/src/components/appointments/styles.tsx\` (or a new \`hooks/\` folder) that:
  - Focuses the dialog node on mount
  - Captures Tab/Shift+Tab within the dialog
  - Calls \`onClose()\` on Escape
  - Restores focus to the opener on unmount

## Tests to add
- Book + Cancel modals: opening focuses the dialog; Escape closes.
- Tab from last element wraps to first.

### LINKED ISSUE #895: nit(portal): prep-instructions appointment type enum should derive from DB schema
## Context
Filed from `/agent-review` of PR #890.

## Issue
\`apps/patient-portal/src/components/appointments/prep-instructions.ts\` hand-codes the \`AppointmentType\` literal union as \`"follow_up" | "new_patient" | "procedure" | "telehealth"\`. This matches today's scheduling Zod enum in \`services/api-gateway/src/routers/scheduling.ts\` and \`services/scheduling/src/router.ts\`, but there is no shared source of truth — adding a new type (e.g. \`"annual_physical"\`, \`"lab_draw"\`) means updating three files, and the existing code falls through to a generic message without a type error.

PR #890 already flags (in its body + inline TODO in the file) that prep text should move server-side so clinicians can edit it without a deploy — this NIT is adjacent: even before that move, export the enum from \`packages/types\` or \`packages/db-schema\` and have both the frontend AND the Zod input share it so missing-type drift gets caught at compile time.

## Fix
1. Add \`export const APPOINTMENT_TYPES = ["follow_up", "new_patient", "procedure", "telehealth"] as const\` to \`@carebridge/types\`.
2. Use \`z.enum(APPOINTMENT_TYPES)\` in both router files.
3. Re-export the type in \`prep-instructions.ts\` instead of redeclaring it.
4. Optionally: make \`PREP_INSTRUCTIONS\` a \`Record<(typeof APPOINTMENT_TYPES)[number], string>\` so TS fails the build when a type is added without prep text.

Combine with the server-side prep-text content table work noted in PR #890's TODO.

blamechris:
## Summary

Wave 8 bundled appointments polish:

- **#892 — atomic reschedule.** New `scheduling.appointments.reschedule` tRPC procedure wraps cancel+book in a single `db.transaction`. A conflict on the new slot rolls back the cancel so the patient is never left without an appointment on a mid-call failure. Target `patient_id` is read from the existing row (not trusted from the wire) so the same `enforcePatientAccess` check used by create+cancel applies. Patient-portal coordinator branches on whether `onReschedule` is provided so legacy pages keep working during rollout.
- **#893 — server-side cancel reason validation.** `cancelReasonSchema = z.string().trim().min(1)` lifted into `@carebridge/validators`. Empty or whitespace-only reasons now return BAD_REQUEST regardless of what the client sends. Same schema gates the reschedule procedure's reason.
- **#894 — modal focus trap.** New `useModalFocusTrap` hook (mirrors the Wave 7 sidebar trap) applied to the appointment booking, cancel, and detail modals. Behaviour: initial focus moves into the modal on open, Tab/Shift+Tab wrap within the modal, Escape closes, focus returns to the opener on close.
- **#895 — prep-instructions enum from the canonical set.** `appointmentTypeSchema` + `AppointmentType` literal union live in `@carebridge/validators` so the prep-instructions map is now `Record<AppointmentType, string>` and the booking wizard renders `APPOINTMENT_TYPES` (derived from the Zod enum). Adding a new type in one place surfaces a compile error everywhere until the map is updated. The DB column is still plain `text` today — a follow-up to promote it to a `pgEnum` remains (called out in code comments).

## Checklist

- [x] #892 — atomic `scheduling.appointments.reschedule` procedure with same RBAC as create+cancel
- [x] #893 — Zod `trim().min(1)` cancel-reason validator; server enforces BAD_REQUEST
- [x] #894 — focus trap + Escape + initial focus + focus restore on book / cancel / detail modals
- [x] #895 — `AppointmentType` literal union from `@carebridge/validators` drives the prep-instructions map + booking wizard

## Test plan

- [x] `pnpm --filter @carebridge/scheduling test` (42/42)
- [x] `pnpm --filter @carebridge/api-gateway test` (437/437) — includes new `scheduling-reschedule-rbac.test.ts` covering happy path, conflict rollback, NOT_FOUND, FORBIDDEN for cross-patient / off-care-team, BAD_REQUEST for empty reason
- [x] `pnpm --filter @carebridge/patient-portal test` (65/65) — includes new `use-modal-focus-trap.test.tsx` and extended `appointments-reschedule-flow.test.tsx` with both legacy cancel-then-book and new atomic reschedule paths
- [x] `pnpm --filter @carebridge/validators test` (183/183)
- [x] `pnpm typecheck`
- [x] `pnpm lint`

Closes #892
Closes #893
Closes #894
Closes #895


blamechris (commented):
## Review — Wave 8 appointments polish (#892 / #893 / #894 / #895)

### VERDICT: Approve with one tracked follow-up

All four issues land cleanly. RBAC, Zod validation, a11y hook, and the enum plumbing are all in good shape. One functional gap called out below is pre-existing in this router surface (api-gateway) and should be tracked but does not block.

### Blockers
None.

### Follow-ups (non-blocking)

**Reminder integration divergence between routers.** The new `schedulingRbacRouter.appointments.reschedule` in `services/api-gateway/src/routers/scheduling.ts` does not call `scheduleReminders` / `cancelReminders`. By contrast, `services/scheduling/src/router.ts` reschedule path (lines ~200–278) both cancels old reminder jobs and schedules fresh ones after the new appointment is inserted, writing the returned `reminder_24h_job_id` / `reminder_2h_job_id` back to the row.

This is consistent with the existing api-gateway `create` / `cancel` (neither of those calls reminders either), so the new procedure matches the local pattern. But from a product correctness standpoint, an atomic reschedule through this router silently drops the two BullMQ reminders for the new slot and leaves stale jobs in the queue for the cancelled original (they'll no-op via the status check in the worker, but still). Since the PR description does not claim this is wired up, filing as a Wave 9 follow-up rather than a blocker. Suggest: add a `scheduleReminders` + `cancelReminders` wrapper after `db.transaction` returns, mirroring `services/scheduling/src/router.ts#182-279`, and write the job IDs back via a second update outside the transaction.

### Key findings

**#892 atomic reschedule — solid.**
- Single tRPC procedure wraps cancel + insert in `db.transaction(...)`.
- `patient_id` is resolved from the existing row (not the wire) before `enforcePatientAccess` — matches the existing `cancel` pattern and defeats cross-patient tampering.
- Overlap check inside the tx excludes the appointment being cancelled (`ne(appointments.id, input.appointmentId)`), which correctly prevents a self-reschedule onto an overlapping slot from racing against its own row.
- Conflict throws `Error("...conflicts...")` inside the tx callback → the db driver aborts the UPDATE. Test `conflict path: rolls back when the new slot is taken` asserts `insertFn` is never called, which is the correct rollback assertion for this mock shape.
- 8 tests on `scheduling-reschedule-rbac.test.ts`: unauthenticated, NOT_FOUND, cross-patient FORBIDDEN, BAD_REQUEST empty reason, happy path, conflict rollback, admin cross-patient, physician off-care-team FORBIDDEN. Plus 3 on cancel (#893). That covers the stated matrix.

**#893 server cancel-reason — correct.**
- `cancelReasonSchema = z.string().trim().min(1, "Cancel reason is required")` lives in `@carebridge/validators/src/scheduling.ts`.
- Applied to both `cancel` and `reschedule` input schemas in api-gateway. Validator package tests (10 cases) cover empty, whitespace, non-string, and trimming-preserves-data. BAD_REQUEST flows are exercised in the api-gateway RBAC test.

**#894 modal focus trap — mirrors the Wave 7 sidebar pattern cleanly.**
- `apps/patient-portal/src/lib/use-modal-focus-trap.ts` covers initial focus, container fallback when no focusables, Tab wrap, Shift+Tab wrap, Escape, and focus restore on close.
- Applied to `BookAppointmentModal`, `CancelAppointmentModal`, and `AppointmentDetailView` (all three get `ref` + `tabIndex={-1}`).
- 6 hook tests + 2 extended reschedule-flow tests asserting the legacy vs atomic branch both reach their respective confirm buttons. The \"wraps Tab at the last focusable\" test asserts on `document.activeElement` after a synthetic `keyDown` on `document`, which matches how the hook attaches its listener.
- Comment notes the decision to keep this portal-local rather than hoisting to a shared package since there is no third consumer — reasonable.

**#895 prep-enum — type guard is tight.**
- `appointmentTypeSchema = z.enum([...])` in validators exports `AppointmentType = z.infer<...>`.
- `PREP_INSTRUCTIONS: Record<AppointmentType, string>` and `APPOINTMENT_TYPE_LABELS: Record<AppointmentType, string>` are both pinned. Adding a fifth variant would force compile errors in both maps.
- The booking wizard now renders `APPOINTMENT_TYPES` derived from `appointmentTypeSchema.options` so a new variant automatically appears in the picker.
- The `prepInstructionsFor` and `appointmentTypeLabel` functions now route unknown strings through `isAppointmentType` (type guard) before indexing, removing the `as AppointmentType` cast. Nice tightening.
- DB column remains `text` — called out in the validator file comment with a pointer to `packages/db-schema/src/schema/scheduling.ts`. Agreed this should be a separate migration.

**Client update.**
- `AppointmentsPageInner` introduces a `Mode` discriminated union with a `reschedule` phase (`reason` → `slot`). `onReschedule` is optional; when omitted the coordinator falls back to the legacy cancel-then-book flow. Existing consumers keep working during rollout.
- The cancel modal's Confirm label is rewritten to \"Continue\" when invoked from the reschedule flow so the user knows nothing has been committed yet — subtle UX win.
- Legacy test (`legacy reschedule flow`) asserts the old two-call sequence still fires; new test (`atomic reschedule flow (#892)`) asserts a single `onReschedule` call with both halves and that `onCancel` / `onBook` are never called. Conflict-surfaces-in-slot-picker test catches UI regressions on failure.

### Nits (not blocking)

1. `scheduling-reschedule-rbac.test.ts` — the \"happy path\" test asserts `result.id !== APPOINTMENT_ID` but does not assert that `updateFn` was called with the `cancelled_by: ctx.user.id` field set to the caller's id. Adding `expect(mocks.updateFn).toHaveBeenCalledWith(...)` on the `.set({...})` payload would catch regressions where the cancellation actor is misattributed in audit.
2. `services/api-gateway/src/routers/scheduling.ts` — the reschedule procedure throws `new Error(\"New time slot conflicts with an existing appointment\")` inside the transaction. Consider wrapping as `TRPCError({ code: \"CONFLICT\" })` so clients get a typed 409 rather than an INTERNAL_SERVER_ERROR. The test uses `.rejects.toThrow(/conflict/i)` so this isn't covered by the tests today.
3. `useModalFocusTrap` — the Escape handler listens on `document`. If two modals are ever stacked (e.g. an unlikely confirm-inside-detail case), both will close on one Escape. Low priority because the current UX only opens one modal at a time.

### Praise

- The `Mode` discriminated union with a phase marker is a clean way to keep the cancel modal reusable across cancel and reschedule flows without leaking reschedule-specific state into the cancel modal props.
- Excellent pre-transaction lookup pattern for NOT_FOUND: bail early without opening a transaction, then re-select inside the tx for the write lock. Explained in the comment.
- The validator-layer `appointmentTypeSchema.options` → `APPOINTMENT_TYPES` → `Record<AppointmentType, string>` chain is the tightest single-source-of-truth for a frontend enum I have seen in this monorepo. The \"adding a new variant in one place surfaces a compile error everywhere\" guarantee is real.
- Docstrings are load-bearing — every non-obvious decision (pre-tx lookup, why the DB column is still text, why the focus trap is portal-local, why the confirm label changes) is explicitly justified.
- Zero-attribution policy clean on commit message + PR body + diff.

### Test counts verified from PR description
- api-gateway 437/437 — includes new `scheduling-reschedule-rbac.test.ts` (11 cases between cancel + reschedule)
- patient-portal 65/65 — includes `use-modal-focus-trap.test.tsx` (6) + extended `appointments-reschedule-flow.test.tsx` (3)
- validators 183/183 — includes new `scheduling.test.ts` (10)
- scheduling 42/42 — untouched by this PR; reminder integration in that package's router is unchanged

CI status on the branch shows \"no checks reported\" — worth a retrigger via `gh run rerun` before merge.

PR #914, opened 2026-04-19
refactor(clinical): care team idempotency, role CHECK, atomic RBAC audit

### LINKED ISSUE #881: careTeam mutations: define idempotency semantics for add/grant/remove/revoke
From review of #880.

## Problem
`careTeamRbacRouter` mutations have undefined idempotency behavior:

- **`assignments.grant`** blindly inserts — calling twice for the same `(user_id, patient_id)` produces duplicate active RBAC rows (no unique index). `assertCareTeamAccess` would still work, but audit and cleanup get noisy.
- **`addMember`** same pattern — a provider can be added twice to the same patient's roster with two active rows.
- **`removeMember`** on an already-removed member overwrites `is_active=false, ended_at=<new timestamp>` and writes another audit entry, obscuring the original removal time.
- **`assignments.revoke`** correctly rejects already-revoked via `isNull(removed_at)` filter → NOT_FOUND. This one is fine.

## Proposal
Pick one:
1. **Idempotent / no-op**: detect existing active row, return existing ID without inserting; still write an audit entry with `action: care_team_<op>_noop`.
2. **Error on duplicate**: return `CONFLICT` (tRPC code) when an active row already exists.

Option 2 is safer for surfacing integration bugs. Either way, document in the router JSDoc and add a unique partial index: `UNIQUE (user_id, patient_id) WHERE removed_at IS NULL` for `care_team_assignments`, and `UNIQUE (patient_id, provider_id) WHERE is_active = true` for `care_team_members`.

## Acceptance
- Policy documented in router file
- DB unique partial indexes added via migration
- Test cases for duplicate-grant / duplicate-add / remove-already-removed

### LINKED ISSUE #882: care_team_members.role: promote text column to CHECK or enum to prevent drift
From review of #880.

## Problem
`careTeamMembers.role` is `text`. The Zod enum (`primary|specialist|nurse|coordinator`) is the only guardrail. Seed uses those values and the new `addMember`/`updateRole` procedures validate against them, but any future code path that bypasses Zod (direct SQL, a new router, data migration) can insert arbitrary strings with no DB-level backstop.

## Proposal
Add a migration introducing either:
- `CHECK (role IN ('primary','specialist','nurse','coordinator'))`, or
- a PostgreSQL `ENUM` type `care_team_role` and migrate the column.

Same applies to `careTeamAssignments.role` (`attending|consulting|nursing|covering`).

## Acceptance
- Migration adds DB constraint
- Drizzle schema reflects the constraint
- Seed values verified to pass constraint

### LINKED ISSUE #883: audit_log: capture RBAC assignment id in care_team_add_member atomic path
From review of #880.

## Problem
When `careTeam.addMember` runs with `assignment_role` set, two rows are inserted atomically: a `careTeamMembers` row and a `careTeamAssignments` row. The audit entry uses the `member_id` as `resource_id` and captures `assignment_role` in `new_value` but not the new assignment row's `id`. An auditor reconstructing the event cannot link the audit row back to the specific RBAC record without inferring from `(patient_id, user_id, assigned_at)`.

## Proposal
Add `assignment_id` to `new_value` (and surface it in the return value so the client has the same visibility).

## Acceptance
- `new_value.assignment_id` captured on the atomic path
- Return type includes `assignment_id` (optional, present only when `assignment_role` was set)
- Test assertion added

blamechris:
## Summary

Bundled polish for the Wave 5 care-team procedures. Each sub-change has a single motivating issue and is small enough to review on its own; they ship together to amortise one migration.

### #881 — Idempotency semantics

**Decision: return-existing on add/grant, 200-no-op on remove/revoke.**

Rationale:
- Clients retrying after a transient network failure shouldn't be punished with a 409 on the second call — they didn't cause the duplicate, the network did. Return-existing keeps retries safe-to-call.
- `removeMember` / `assignments.revoke` align with RFC 9110 idempotent DELETE — a second remove on a tombstoned resource is a 200 no-op, not an error.
- Critically, the remove/revoke paths now short-circuit BEFORE the UPDATE, so the original `ended_at` / `removed_at` is preserved. Auditors need to trust WHEN the event actually happened; a stale retry must not rewrite that history.
- A partial `UNIQUE ... WHERE is_active = true` index backs the semantic at the DB level — if two concurrent writers race past the app-level check, the second one gets a clean unique-violation rather than silent duplication.

Alternative considered (explicit 409 CONFLICT): rejected because idempotent retries are the common failure mode and a CONFLICT surfaces a "real" error to monitoring dashboards.

### #882 — Role CHECK constraint

`care_team_members.role` and `care_team_assignments.role` now carry DB CHECK constraints mirroring the Zod enums in `packages/validators/src/care-team.ts`. Zod remains the source of truth (app layer), the CHECK catches direct SQL writes (seeds, manual fixes, future services). A parity test in `packages/validators/src/__tests__/care-team.test.ts` fails if either side drifts.

Chose plain CHECK over `pgEnum` — simpler, additive, and the DROP/ADD migration pattern matches `0026_family_access_constraints.sql`.

### #883 — Audit RBAC assignment_id in atomic path

When `addMember` inserts both a `care_team_members` row and a `care_team_assignments` row in the same transaction, the audit row's `details.new_value` now includes the `assignment_id`. Without this, an auditor could see the roster change but had no way to correlate it with the access grant it triggered.

## Non-goals

- No RBAC middleware refactor.
- No role-enum expansion.
- No touches to allergy, appointments, or notes code.

## Test plan

- [x] `pnpm --filter @carebridge/api-gateway test` (492 passed — 35 care-team-rbac)
- [x] `pnpm --filter @carebridge/validators test` (199 passed — 21 new care-team parity)
- [x] `pnpm typecheck` (all packages green)
- [x] `pnpm lint` (no new warnings)
- [x] Full monorepo `pnpm test` passes
- [x] Migration `0038_care_team_constraints.sql` is additive; existing seed rows (primary / specialist / nurse / attending / consulting / nursing) conform to both CHECK sets
- [x] Partial UNIQUE indexes exclude soft-deleted / revoked rows so HIPAA history is retained

## Acceptance criteria

- [x] `addMember` returns the existing row on duplicate without writing a new audit entry
- [x] `assignments.grant` same
- [x] `removeMember` no-ops on already-inactive member (ended_at preserved)
- [x] `assignments.revoke` no-ops on already-revoked assignment (removed_at preserved)
- [x] `care_team_members.role` rejects out-of-enum values at the DB level
- [x] `care_team_assignments.role` rejects out-of-enum values at the DB level
- [x] `audit_log.details.new_value.assignment_id` present when the atomic path inserted one
- [x] Wave 5 transactional rollback contract preserved (covered by existing `rolls back the team insert when the atomic RBAC grant fails` test)

Closes #881
Closes #882
Closes #883

### DIFF
--- a/packages/db-schema/drizzle/0038_care_team_constraints.sql
+++ b/packages/db-schema/drizzle/0038_care_team_constraints.sql
@@ -1,0 +1,53 @@
+-- Care-team polish: idempotency indexes + role CHECK constraints.
+-- Closes issues #881 (idempotency) and #882 (role CHECK parity).
+--
+--  1. CHECK constraints mirror the Zod enums in
+--     `packages/validators/src/care-team.ts`. Zod is the source of truth
+--     (app layer); this DB guardrail catches direct SQL writes (seeds,
+--     manual fixes, future services) that could otherwise bypass it.
+--     The `packages/validators/src/__tests__/care-team.test.ts` tests
+--     break if either side drifts.
+--
+--  2. Partial UNIQUE indexes prevent duplicate ACTIVE rows for a given
+--     (patient, provider/user) pair — the DB-level enforcement of the
+--     idempotency semantics implemented in services/api-gateway/src/
+--     routers/care-team.ts. Soft-deleted rows are retained for HIPAA
+--     history and are explicitly excluded from the uniqueness check.
+--
+-- Non-destructive: existing rows already conform to both sets.
+
+BEGIN;
+
+-- ----------------------------------------------------------------------------
+-- 1. CHECK constraints on role columns (#882)
+-- ----------------------------------------------------------------------------
+
+ALTER TABLE "care_team_members"
+  DROP CONSTRAINT IF EXISTS "care_team_members_role_check";
+ALTER TABLE "care_team_members"
+  ADD CONSTRAINT "care_team_members_role_check"
+  CHECK (role IN ('primary', 'specialist', 'nurse', 'coordinator'));
+
+ALTER TABLE "care_team_assignments"
+  DROP CONSTRAINT IF EXISTS "care_team_assignments_role_check";
+ALTER TABLE "care_team_assignments"
+  ADD CONSTRAINT "care_team_assignments_role_check"
+  CHECK (role IN ('attending', 'consulting', 'nursing', 'covering'));
+
+-- ----------------------------------------------------------------------------
+-- 2. Partial UNIQUE indexes on active rows (#881)
+-- ----------------------------------------------------------------------------
+
+-- Clinical roster: one active row per (provider, patient). Soft-deleted
+-- rows (is_active = false) are excluded — we retain them for audit.
+CREATE UNIQUE INDEX IF NOT EXISTS "idx_care_team_members_active_unique"
+  ON "care_team_members" ("provider_id", "patient_id")
+  WHERE is_active = true;
+
+-- RBAC mapping: one active grant per (user, patient). Revoked rows
+-- (removed_at IS NOT NULL) are excluded.
+CREATE UNIQUE INDEX IF NOT EXISTS "idx_care_team_assignments_active_unique"
+  ON "care_team_assignments" ("user_id", "patient_id")
+  WHERE removed_at IS NULL;
+
+COMMIT;
--- a/packages/db-schema/drizzle/meta/_journal.json
+++ b/packages/db-schema/drizzle/meta/_journal.json
@@ -267,6 +267,13 @@
       "when": 1746892800000,
       "tag": "0037_allergy_overrides",
       "breakpoints": true
+    },
+    {
+      "idx": 38,
+      "version": "7",
+      "when": 1746979200000,
+      "tag": "0038_care_team_constraints",
+      "breakpoints": true
     }
   ]
 }
--- a/packages/db-schema/src/schema/patients.ts
+++ b/packages/db-schema/src/schema/patients.ts
@@ -1,5 +1,6 @@
 import crypto from "node:crypto";
-import { pgTable, text, boolean, real, index } from "drizzle-orm/pg-core";
+import { sql } from "drizzle-orm";
+import { pgTable, text, boolean, real, index, uniqueIndex, check } from "drizzle-orm/pg-core";
 import { encryptedText } from "../encryption.js";
 
 export const patients = pgTable("patients", {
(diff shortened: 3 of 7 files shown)


blamechris (commented):
## VERDICT: APPROVE (comment — author self-review)

Tight, well-scoped bundle. Idempotency, CHECK parity, and the `assignment_id` audit field all behave as advertised. Defence-in-depth (app-level short-circuit + partial UNIQUE index + Zod/CHECK parity test) is exactly the right posture for HIPAA-audited mutations.

### Criteria check

1. **Zero Attribution** — clean, no Claude/AI mentions anywhere.
2. **Idempotency semantics** — `addMember` / `assignments.grant` return the existing row before entering the transaction; `removeMember` / `assignments.revoke` return before any UPDATE runs. Tests explicitly assert `memberRows`, `assignmentRows`, and `auditRows` are all empty on every idempotent path. No spurious audit rows.
3. **Transaction integrity** — the idempotency SELECT is a pre-transaction short-circuit. The original atomic `tx.insert(members) + tx.insert(assignments) + audit` sequence is untouched, and the "rolls back team insert when atomic RBAC grant fails" test still guards it.
4. **Partial UNIQUE indexes** — migration 0038 creates both (`idx_care_team_members_active_unique WHERE is_active = true`, `idx_care_team_assignments_active_unique WHERE removed_at IS NULL`). Drizzle schema mirrors them via `uniqueIndex(...).where(...)`.
5. **CHECK constraint parity** — migration literals match the Zod enum options one-for-one (`primary/specialist/nurse/coordinator` and `attending/consulting/nursing/covering`). Parity test compares sorted `.options` so ordering isn't load-bearing — drift breaks the test.
6. **`assignment_id` in audit** — `let assignmentId: string | undefined` is set only inside `if (input.assignment_role)`, and the audit row uses `...(assignmentId !== undefined && { assignment_id: assignmentId })`. Both the "present when paired" and "absent when unpaired" cases have dedicated tests, and the present case asserts it equals the actually-inserted row's id.
7. **Revoke NOT_FOUND vs already-revoked** — SELECT drops the `isNull(removed_at)` filter; missing id throws `NOT_FOUND`, already-revoked returns the no-op. Error message updated to drop the now-misleading "or already revoked" suffix. Good call — that string would have been a lie after this change.
8. **Test coverage** — 7 new care-team-rbac cases (addMember idempotency x2, audit assignment_id present/absent x2, removeMember idempotency x1, grant idempotency x1, revoke idempotency x1) plus 21 validator parity tests. `audit-client-ip` was correctly updated to feed a `null` selectRow so the idempotency pre-check falls through to the insert path.

### Subtleties worth noting

- `removeMember`'s SELECT is keyed on `id` only (no `is_active` filter), so the `!existing.is_active` branch is actually reachable — matches the tests.
- Pre-transaction SELECT is by design a TOCTOU window; the partial UNIQUE index is the safety net, and the PR body calls this out explicitly. Correct layering.
- Rationale for "return existing" vs 409 is thoughtfully argued in the PR description and matches RFC 9110 for DELETE semantics.

### Nits (non-blocking)

- None worth gating on. The inline comments explaining the audit-trail preservation motive on both `removeMember` and `assignments.revoke` short-circuits are genuinely excellent and should be preserved in any future refactor.

### Praise

- The "don't overwrite `ended_at` / `removed_at` on retry" reasoning is the right HIPAA instinct, and the tests encode it rather than trusting a comment.
- The parity-via-`.options` test is the correct shape — set equality, no ordering dependency, and the failure message will point directly at the drift.
- Keeping the old error-message text honest ("not found" vs "not found or already revoked") after changing the SELECT semantics shows care.

LGTM.

PR #915, opened 2026-04-19
feat(clinical): harden allergy-override storage, internal notes router, and note version-history completeness

### LINKED ISSUE #884: Harden internal clinical-notes router against cosigned_by / amended_by spoofing
From review of #879.

`services/clinical-notes/src/router.ts` currently exposes `cosign` and `amend` procedures whose input schemas extend the validator schemas with client-supplied user ids:

```ts
.input(cosignNoteSchema.extend({ cosigned_by: z.string().uuid() }))
...
.input(amendNoteSchema.extend({ amended_by: z.string().uuid() }))
```

This mirrors the pre-fix shape of `signNote` in issue #271. The RBAC gateway router in `services/api-gateway/src/routers/clinical-notes.ts` already does the right thing — it takes the cosigner/amender id from `ctx.user.id` and ignores any client-supplied value — but the internal router is the one exported as `clinicalNotesRouter` from `@carebridge/clinical-notes`.

Today that router is not mounted at any HTTP boundary (verified via grep). However:

1. It's a latent spoofing vector if it's ever mounted directly (e.g. for an internal service-to-service path).
2. It's inconsistent with the hardened pattern used by `sign` in the same file, which intentionally does not merge `signNoteSchema` for exactly this reason.

### Fix
Drop the `.extend({ cosigned_by })` / `.extend({ amended_by })` from the internal router. If the internal router needs a caller identity for non-gateway tests / service callers, take it from a tRPC `ctx` injected by those callers rather than from client input, matching the gateway pattern.

### Context
- [PR #879 review](https://github.com/blamechris/carebridge/pull/879#issuecomment-4274648951)
- Related hardening: issue #271 (signature spoofing)

### LINKED ISSUE #886: Record old_status / new_status on amend audit_log row
From review of #879.

The explicit audit row emitted by `clinicalNotes.amend` captures `new_version` but not the pre-amendment status. Because amend accepts three source states (`signed`, `cosigned`, `amended`), the HIPAA amendment trail currently can't distinguish \"amended a signed note\" from \"amended an already-amended note\" without joining the clinical event log or re-reading the prior version row.

Current shape:

```ts
details: JSON.stringify({
  amended_by: ctx.user.id,
  reason: input.reason,
  new_version: result.version,
}),
```

### Suggested shape

```ts
details: JSON.stringify({
  amended_by: ctx.user.id,
  reason: input.reason,
  old_status: existing.status,      // add
  new_status: \"amended\",            // add
  previous_version: existing.version,// add
  new_version: result.version,
}),
```

The gateway already holds `existing` from the NOT_FOUND / enforcePatientAccess lookup, so no extra DB read is needed. A matching addition on the `cosign` audit row (`old_status: \"signed\"`, `new_status: \"cosigned\"`) would make both entries self-contained for downstream HIPAA amendment-review tooling.

### Context
- [PR #879 review](https://github.com/blamechris/carebridge/pull/879#issuecomment-4274648951)

### LINKED ISSUE #888: Archive initial v1 on note creation so version history reflects birth
From review of #879.

Per issue #398's \"archive on every state change\" rule, `createNote` could archive the initial v1 sections into `note_versions`. Today it does not — v1 only appears in `note_versions` once sign/cosign/amend fire (and even then with the [collision caveat](https://github.com/blamechris/carebridge/pull/879#issuecomment-4274648951)).

For a freshly-created but unsigned note, `getVersionHistory` returns `[]`, while the live `clinical_notes` row exists. That's defensible (the content is in the live row) but makes the \"full chronological spine\" advertised in #398 slightly frayed at the start.

### Suggested fix
Call `archiveVersion({ noteId, version: 1, sections, savedBy: provider_id, savedAt: now })` at the end of `createNote`. Would also fix the asymmetry where draft edits via `updateNote` produce history rows but the original draft itself does not.

### Context
- [PR #879 review](https://github.com/blamechris/carebridge/pull/879#issuecomment-4274648951)
- Should be tackled alongside / after the blocking fix for the note_versions version-collision ordering in #879.

### LINKED ISSUE #905: NIT: allergy_overrides — store medication_name/allergen_name on row instead of regex-parsing flag summary
Spotted during review of PR #901.

### Context
`buildPatientContextForRules` (services/ai-oversight/src/services/review-service.ts) populates `resolved_overrides` by left-joining `allergy_overrides` with `clinical_flags` and regex-extracting allergen + medication from the flag's `summary`:

```
function extractMedicationFromFlagSummary(summary: string | null | undefined): string | null {
  if (!summary) return null;
  const m = summary.match(/^Medication "([^"]+)"/);
  return m ? m[1]! : null;
}
```

### Problem
The regex is coupled to the canonical flag summary string produced by `checkAllergyMedication`. Any future change to that string (translation, rewording, "Drug" vs "Medication", locale, etc.) silently breaks `resolved_overrides` population and under-suppresses — a clinician will see a flag re-fire for an allergy-drug pair they've already formally cleared, despite this PR's goal.

### Suggested fix
Denormalise the allergen and medication names onto the `allergy_overrides` row at write time. The override procedure already has both in scope (from the flag + allergy lookup):

- Add `medication_name text` and `allergen_name text` columns to `allergy_overrides`.
- Populate at INSERT time from the clinical_flags.summary OR (preferred) from direct inputs on the override mutation.
- Drop the flag-summary regex from `buildPatientContextForRules`.

### Impact
UNDER-suppression is a safety-neutral failure mode (flag re-fires, clinician re-reviews). But the suppression feature is the entire rule-layer integration in this PR, so coupling it to a parseable string is fragile.

### Labels
enhancement, clinical-safety, from-review

### LINKED ISSUE #906: NIT: enforce allergy_overrides immutability at DB level (REVOKE DELETE/UPDATE)
Spotted during review of PR #901.

### Context
Migration 0037 creates the `allergy_overrides` table with comments stating rows are PERMANENT and never deleted. App-layer enforcement is satisfied (no code path calls DELETE/UPDATE on the table), but the DB does not prevent it.

### Precedent
Migration 0012 (`0012_audit_log_immutability.sql`) enforces audit_log append-only at the DB level. `allergy_overrides` is morally similar — it's a HIPAA-grade audit trail of clinical decisions — but lacks the same guarantee.

### Suggested fix
Add a follow-up migration mirroring 0012 for allergy_overrides:
- REVOKE DELETE, UPDATE ON allergy_overrides FROM the app role.
- Optionally add an audit trigger on DELETE attempts.

### Labels
enhancement, security, from-review

blamechris:
## Summary

Five small hardening fixes bundled into one PR — all touch the same audit-and-history surface and share test infrastructure.

### #905 — denormalise `medication_name` / `allergen_name` onto `allergy_overrides`
- [x] Migration adds nullable `medication_name` and `allergen_name` columns.
- [x] `allergies.override` mutation populates both from the structured allergy row + flag summary.
- [x] `buildPatientContextForRules` selects the new columns and prefers them over the legacy `flag_summary` parse.
- [x] Legacy rows (both columns NULL) still resolve via the summary-parse fallback.
- [x] Suppression tests unchanged; new loader-preference tests added.

### #906 — DB-level append-only for `allergy_overrides`
- [x] Migration installs `BEFORE UPDATE` / `BEFORE DELETE` triggers that raise, mirroring the `0012_audit_log_immutability.sql` precedent.
- [x] Non-destructive (additive). Triggers are idempotent via `DROP TRIGGER IF EXISTS`.

### #884 — remove spoofing vector on internal clinical-notes router
- [x] `cosigned_by` / `amended_by` / `signed_by` removed from Zod input schemas on the internal `services/clinical-notes/src/router.ts`.
- [x] Actor id now sourced from `ctx.actorId`; missing actor raises `UNAUTHORIZED`.
- [x] Gateway router unaffected (already used `ctx.user.id`).
- [x] New regression tests in `services/clinical-notes/src/__tests__/router-actor-spoof.test.ts` confirm stripped-input + ctx-override behaviour and the empty-actor failure path.

### #886 — HIPAA-complete amend audit details
- [x] Gateway amend procedure now reads pre-amend `status` + `version` before the service call.
- [x] Audit `details` JSON includes `old_status`, `new_status`, `previous_version`, `new_version` in addition to the pre-existing `amended_by` / `reason`.
- [x] New tests in `services/api-gateway/src/__tests__/clinical-notes-amend-audit.test.ts` cover signed→amended, cosigned→amended, and amended→amended transitions.

### #888 — archive v1 on note creation
- [x] `createNote` calls `archiveVersion` with `lifecycle_event="draft"`, `version=1` immediately after the note insert.
- [x] `create→sign→cosign` now yields `[draft, signed, cosigned]` (was `[signed, cosigned]`).
- [x] `create→sign→amend→amend` now yields `[draft, signed, amended, amended]` (was 3 rows).
- [x] `create→getVersionHistory` now returns `[draft v1]` (was `[]`).
- [x] Existing `note-service.test.ts` regression tests updated; new test pins the draft-only case.

## Schema migration

- **`packages/db-schema/drizzle/0040_allergy_overrides_hardening.sql`** — additive columns + immutability triggers. Both changes are non-destructive (nullable columns; triggers rely on the pre-existing no-UPDATE/no-DELETE contract at the app layer).

## Test plan

- [x] `pnpm --filter @carebridge/ai-oversight test` — 799 green
- [x] `pnpm --filter @carebridge/clinical-notes test` — 55 green
- [x] `pnpm --filter @carebridge/api-gateway test` — 491 green
- [x] `pnpm --filter @carebridge/validators test` — 178 green
- [x] `pnpm lint` — green (only pre-existing warnings in patient/clinician portal)

Closes #884
Closes #886
Closes #888
Closes #905
Closes #906

### DIFF
--- a/packages/db-schema/drizzle/0040_allergy_overrides_hardening.sql
+++ b/packages/db-schema/drizzle/0040_allergy_overrides_hardening.sql
@@ -1,0 +1,58 @@
+-- Harden the allergy_overrides audit trail (#905, #906).
+--
+-- Two related changes rolled into one non-destructive migration:
+--
+-- 1. Denormalise the allergen / medication strings that the override
+--    pertains to (issue #905).
+--
+--    Prior to this change, the ai-oversight rule layer recovered those
+--    strings by regex-parsing `clinical_flags.summary` — which is fragile
+--    (summary wording is a UX concern, not a schema contract) and couples
+--    the rule-layer suppression decision to the exact phrasing of the
+--    flag's human-readable message. Persisting the allergen and medication
+--    as explicit columns lets the suppression query read them directly,
+--    decouples the rule layer from the flag-summary text, and gives HIPAA
+--    reviewers a queryable field rather than a substring match.
+--
+--    Both columns are nullable to preserve backward compatibility for
+--    rows written prior to this migration; the rule layer falls back to
+--    the legacy parsing-from-summary path when either column is NULL,
+--    and the suppression tests cover both legacy and denormalised rows.
+--
+-- 2. Enforce immutability at the database level (issue #906).
+--
+--    The application layer already treats allergy_overrides as permanent —
+--    `allergies.override` only ever INSERTs rows — but there was no
+--    database-level safety net. Following the precedent in
+--    0012_audit_log_immutability.sql, we install BEFORE UPDATE / BEFORE
+--    DELETE triggers that raise an exception. This closes the residual
+--    risk that a future code path (or a misconfigured admin migration)
+--    could mutate or delete override records, which would undermine the
+--    audit-trail guarantee the table exists to provide.
+--
+--    Triggers are used instead of a REVOKE of table privileges so the
+--    constraint travels with the schema and does not depend on how roles
+--    are provisioned in each deployment.
+
+-- (1) Denormalised allergen / medication strings (nullable for backfill).
+ALTER TABLE "allergy_overrides"
+  ADD COLUMN IF NOT EXISTS "medication_name" text,
+  ADD COLUMN IF NOT EXISTS "allergen_name" text;
+
+-- (2) Append-only immutability (trigger-based, mirrors 0012_audit_log_*).
+CREATE OR REPLACE FUNCTION prevent_allergy_override_modification()
+RETURNS TRIGGER AS $$
+BEGIN
+  RAISE EXCEPTION 'allergy_overrides is append-only; UPDATE/DELETE not permitted';
+END;
+$$ LANGUAGE plpgsql;
+
+DROP TRIGGER IF EXISTS allergy_overrides_no_update ON "allergy_overrides";
+CREATE TRIGGER allergy_overrides_no_update
+  BEFORE UPDATE ON "allergy_overrides"
+  FOR EACH ROW EXECUTE FUNCTION prevent_allergy_override_modification();
+
+DROP TRIGGER IF EXISTS allergy_overrides_no_delete ON "allergy_overrides";
+CREATE TRIGGER allergy_overrides_no_delete
+  BEFORE DELETE ON "allergy_overrides"
+  FOR EACH ROW EXECUTE FUNCTION prevent_allergy_override_modification();
--- a/packages/db-schema/drizzle/meta/_journal.json
+++ b/packages/db-schema/drizzle/meta/_journal.json
@@ -267,6 +267,13 @@
       "when": 1746892800000,
       "tag": "0037_allergy_overrides",
       "breakpoints": true
+    },
+    {
+      "idx": 38,
+      "version": "7",
+      "when": 1747152000000,
+      "tag": "0040_allergy_overrides_hardening",
+      "breakpoints": true
     }
   ]
 }
--- a/packages/db-schema/src/schema/allergy-overrides.ts
+++ b/packages/db-schema/src/schema/allergy-overrides.ts
@@ -20,6 +20,9 @@ import { users } from "./auth.js";
  *
  * Related:
  *  - migration 0037_allergy_overrides.sql — schema + reason CHECK constraint
+ *  - migration 0040_allergy_overrides_hardening.sql — denormalised
+ *    medication_name / allergen_name columns (#905) + DB-level
+ *    append-only triggers (#906)
  *  - services/ai-oversight/src/rules/allergy-medication.ts — suppresses
  *    flags for allergy-drug pairs with an existing override
  *  - services/api-gateway/src/routers/patient-records.ts — `allergies.override`
@@ -49,6 +52,16 @@ export const allergyOverrides = pgTable(
     overridden_at: text("overridden_at")
       .notNull()
       .$defaultFn(() => new Date().toISOString()),
+    /**
+     * Denormalised medication / allergen strings (issue #905). Added in
+     * migration 0040 so the rule-layer suppression (`checkAllergyMedication`)
+     * can read them directly rather than regex-parsing the triggering flag's
+     * `summary` column. Nullable to preserve compatibility with rows written
+     * before 0040 — the review-service override loader falls back to the
+     * legacy flag-summary parse when either column is NULL.
+     */
+    medication_name: text("medication_name"),
+    allergen_name: text("allergen_name"),
   },
   (table) => [
     index("idx_allergy_overrides_patient").on(
--- a/services/ai-oversight/src/__tests__/build-patient-context-age.test.ts
+++ b/services/ai-oversight/src/__tests__/build-patient-context-age.test.ts
@@ -49,6 +49,8 @@ vi.mock("@carebridge/db-schema", () => ({
     flag_id: "flag_id",
     override_reason: "override_reason",
     overridden_at: "overridden_at",
+    medication_name: "medication_name",
+    allergen_name: "allergen_name",
   },
   clinicalFlags: {
     id: "id",
--- a/services/ai-oversight/src/__tests__/build-patient-context.test.ts
+++ b/services/ai-oversight/src/__tests__/build-patient-context.test.ts
@@ -39,6 +39,8 @@ vi.mock("@carebridge/db-schema", () => ({
     flag_id: "flag_id",
     override_reason: "override_reason",
     overridden_at: "overridden_at",
+    medication_name: "medication_name",
+    allergen_name: "allergen_name",
   },
   clinicalFlags: {
     id: "id",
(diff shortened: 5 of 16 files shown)


blamechris (commented):
## VERDICT: APPROVE (self-review, posting as comment)

Five tightly-scoped hardening fixes that all land cleanly. Migration 0040 is additive and correct; the internal-router spoofing fix is the most load-bearing piece and is well-defended. The allergy-override denormalisation preserves suppression semantics while decoupling the rule layer from flag-summary text.

## Blockers
None.

## Nits (non-blocking)

1. **`createNote` v1 archive is not transactional with the note insert.** If `archiveVersion` fails after the `clinical_notes` row is inserted, we end up with a note that has no v1 snapshot — exactly the state #888 was trying to eliminate. Existing `updateNote/signNote/amendNote` share the same pattern, so this is not a regression. A follow-up wrapping `db.insert(clinicalNotes)` + `archiveVersion(v1)` in a transaction would close the residual window.

2. **Amend audit capture race window.** The pre-amend `status`/`version` read and the `amendNote` service call are not co-transactional. Comment in the diff says the service-side amend is ordered by DB row state, but the gateway's read of `existing.status`/`existing.version` is snapshot-at-read; two concurrent amenders could both observe `version=3` and write inconsistent `previous_version` values in their respective audit rows. The underlying `amendNote` service is also non-transactional, so fixing this properly requires a repeatable-read txn spanning both reads. Not a regression introduced here — flag for the audit-integrity follow-up.

3. **Allergen denormalisation preference.** `allergen_name` prefers the structured `allergies.allergen` row over the flag-summary parse. When those differ (e.g. structured row says "Amoxicillin" but the flag summary said `allergy to "Penicillin"`), a future override-suppression lookup that string-matches on the flag's allergen would miss. The existing suppression pathway uses `resolved_overrides[].allergen` directly so this is consistent end-to-end, but the preference is a behavioural change worth documenting in `docs/` for HIPAA reviewers — the second test in `clinical-notes-amend-audit.test.ts` pins the intent but doesn't explain the "why".

4. **Migration numbering skip.** Journal jumps from idx 37 (tag `0037_allergy_overrides`) to idx 38 (tag `0040_allergy_overrides_hardening`) — the `0038`/`0039` SQL files must exist on the PR branch for drizzle-kit to accept the journal. Worth confirming `pnpm db:migrate` runs cleanly against a fresh DB.

5. **`ctx.actorId` typing.** `ClinicalNotesRouterContext` leaves `actorId` optional at the type level; `getActorId` enforces presence at runtime. A non-optional typed variant for the mutation subrouter (e.g. a `t.middleware` that narrows the type) would let the compiler catch mis-wired internal callers instead of relying on the UNAUTHORIZED runtime path. Not required — the runtime guard + tests are already strong.

## Praise

- **The internal-router fix is exactly right.** Dropping `signed_by`/`cosigned_by`/`amended_by` from Zod input and sourcing from `ctx.actorId` is the correct design. The explanatory comment ("this internal router is the last line of defence") is the kind of prose that saves the next maintainer 20 minutes. Triple tests (stripped-input, empty-string actor, missing actor → UNAUTHORIZED) close all three spoof vectors.
- **Migration 0040 mirrors `0012_audit_log_immutability.sql` exactly.** Same trigger function pattern, same `RAISE EXCEPTION` wording, same BEFORE-UPDATE/DELETE guards, `DROP TRIGGER IF EXISTS` for idempotency, and a header comment explaining the "why not REVOKE" choice.
- **#905 denormalisation is decoupled properly.** The review-service loader (`o.allergen_name ?? extractAllergenFromFlagSummary(o.flag_summary)`) keeps legacy rows working byte-identically — zero suppression regression risk for pre-migration data. The three new tests (prefer-denorm / fall-back / both-null) cover the decision tree exhaustively.
- **Version-history tests updated correctly for #888.** `[draft, signed, cosigned]` = 3 rows, `[draft, signed, amended, amended]` = 4 rows, and the new draft-only regression guard all land. The comment on why all three lifecycle rows share `version=1` is load-bearing for the next reader.
- **Amend audit details JSON now carries the full HIPAA triple** (`old_status` / `new_status` / `previous_version` / `new_version`) with explicit test coverage for signed→amended, cosigned→amended, and amended→amended transitions.
- Zero-attribution clean across the diff and commit bodies.

## Key findings

- Gateway router (`services/api-gateway/src/routers/clinical-notes.ts`) imports `noteService` directly, not the internal tRPC router — so the #884 spoofing fix has zero gateway-flow impact, which is the right blast radius.
- The internal router's sign/cosign/amend inputs have been narrowed to note-id-only (sign) or the base validator schemas (cosign/amend); the previous `.extend({ cosigned_by: z.string().uuid() })` attack surface is gone. `@ts-expect-error` annotations in the test file prove the types also refuse spoofed input at compile time.
- `createNote` writes the v1 archive with `saved_at = now` (same timestamp as the note insert) and `lifecycle_event="draft"`; `getVersionHistory` for a never-transitioned note now returns `[{version: 1, lifecycle_event: "draft"}]` instead of `[]`.
- Allergen/medication regexes in the gateway's `allergies.override` (`/^Medication "([^"]+)"/` and `/allergy to "([^"]+)"/`) exactly match both summary formats emitted by `checkAllergyMedication` (direct match + cross-reactivity). A parse miss leaves the columns NULL and the loader falls back — no data loss.
- CI green: Build / Test / Typecheck all passing.

Issue #818, opened 2026-04-18
test(ai): add integration test for concurrent outbox reconciler batch disjointness


blamechris:
PR #806 simplified the outbox-reconciler tests by mocking `@carebridge/outbox` functions directly. The previous structural tests for concurrent reconciler batch disjointness (verifying that two pods claim disjoint batches via FOR UPDATE SKIP LOCKED) were removed. The guarantee still holds via the SQL in `readPendingBatch`, but there is no longer a test that verifies this property end-to-end.

Consider adding an integration test (against a real Postgres instance) that runs two concurrent `readPendingBatch` calls and asserts the returned row sets are disjoint.

Found during review of PR #806.

PR #953, opened 2026-04-19
fix(db): add clinical_flags.rule_id column in migration 0024 (#925)

### LINKED ISSUE #925: bug(db): migration 0024 creates index on clinical_flags.rule_id but no prior migration adds the column
## Context

Running \`pnpm db:migrate\` against a fresh Postgres (as part of wiring the #917 integration test into CI) fails with:

\`\`\`
Migration failed: PostgresError: column \"rule_id\" does not exist
\`\`\`

## Root cause

\`packages/db-schema/drizzle/0024_flag_dedup_constraint.sql\` creates a unique index referencing \`clinical_flags.rule_id\` and \`clinical_flags.category\`/\`severity\`, but no migration between \`0000_initial_schema.sql\` and \`0024\` ever runs \`ALTER TABLE clinical_flags ADD COLUMN rule_id text\`. The column is defined in the Drizzle schema (\`packages/db-schema/src/schema/ai-flags.ts\`), so ORM-mediated reads work, but a plain \`pnpm db:migrate\` against an empty DB hits the gap.

Repro:
\`\`\`
docker run --rm -d -e POSTGRES_PASSWORD=x -p 5432:5432 postgres:16-alpine
DATABASE_URL=postgresql://postgres:x@localhost:5432/postgres pnpm db:migrate
\`\`\`

## Fix

Add a migration (e.g. \`0023a_clinical_flags_rule_id.sql\`) containing \`ALTER TABLE clinical_flags ADD COLUMN IF NOT EXISTS rule_id text;\` and re-run \`drizzle-kit generate\` to capture any other drift. Verify \`pnpm db:migrate\` runs to completion against a fresh Postgres before closing.

## Blocked work

- #818 / PR #917: the integration test is gated on \`TEST_DATABASE_URL\` and runs fine locally when you apply the fix inline, but CI can't run it until this is fixed.

blamechris:
## Summary

- Prepend \`ALTER TABLE clinical_flags ADD COLUMN IF NOT EXISTS rule_id text;\` to \`0024_flag_dedup_constraint.sql\` so \`pnpm db:migrate\` succeeds against a fresh Postgres (was erroring with \"column rule_id does not exist\" when the partial unique index tried to reference it).
- Add a static-analysis guard test that ensures the ALTER TABLE appears before the partial index that references it.
- Wire the missing \`test\` script in \`packages/db-schema/package.json\` so the new guard (and the pre-existing encryption / hmac / mfa tests) actually run in CI via \`turbo test\`.

## Why this is safe for existing databases

The \`ADD COLUMN IF NOT EXISTS\` is idempotent — existing databases (where \`rule_id\` was added via a prior \`drizzle-kit push\`, schema sync, or manual DDL) are unaffected. Fresh databases now get the column before the partial index references it.

## Test plan

- [x] \`pnpm --filter @carebridge/db-schema test\` — 34 tests pass (9 suites)
- [x] \`pnpm typecheck\` clean
- [x] \`pnpm lint\` clean
- [ ] \`pnpm db:migrate\` end-to-end against a fresh Postgres — blocked locally (docker not running); will verify in CI once \`TEST_DATABASE_URL\` is wired in a follow-up PR (tracked in #818 / #917 chain)

Closes #925

### DIFF
--- a/packages/db-schema/drizzle/0024_flag_dedup_constraint.sql
+++ b/packages/db-schema/drizzle/0024_flag_dedup_constraint.sql
@@ -2,6 +2,12 @@
 -- Closes the TOCTOU race window in flag-service.ts where concurrent
 -- workers could both SELECT (no dup) then INSERT (two identical flags).
 
+-- Ensure rule_id column exists. It is defined in the Drizzle schema
+-- (packages/db-schema/src/schema/ai-flags.ts) but no prior migration
+-- adds it, so `pnpm db:migrate` against a fresh DB would error on
+-- the partial index below without this ALTER.
+ALTER TABLE clinical_flags ADD COLUMN IF NOT EXISTS rule_id text;
+
 -- Rule-based flags: only one open flag per (patient, rule) at a time.
 CREATE UNIQUE INDEX IF NOT EXISTS idx_flags_open_rule_dedup
   ON clinical_flags (patient_id, rule_id)
--- a/packages/db-schema/package.json
+++ b/packages/db-schema/package.json
@@ -14,6 +14,7 @@
   "scripts": {
     "build": "tsc",
     "typecheck": "tsc --noEmit",
+    "test": "node --test --import tsx src/__tests__/*.test.ts",
     "generate": "drizzle-kit generate",
     "migrate": "tsx src/migrate.ts",
     "clean": "rm -rf dist"
--- a/packages/db-schema/src/__tests__/migration-0024.test.ts
+++ b/packages/db-schema/src/__tests__/migration-0024.test.ts
@@ -1,0 +1,35 @@
+import { describe, it } from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+import { fileURLToPath } from "node:url";
+
+const __dirname = fileURLToPath(new URL(".", import.meta.url));
+const MIGRATION_PATH = join(
+  __dirname,
+  "..",
+  "..",
+  "drizzle",
+  "0024_flag_dedup_constraint.sql",
+);
+
+describe("migration 0024_flag_dedup_constraint", () => {
+  const sql = readFileSync(MIGRATION_PATH, "utf8");
+
+  it("adds clinical_flags.rule_id column before referencing it", () => {
+    const addColIdx = sql.search(
+      /ALTER\s+TABLE\s+clinical_flags\s+ADD\s+COLUMN\s+IF\s+NOT\s+EXISTS\s+rule_id/i,
+    );
+    const refIdx = sql.search(/idx_flags_open_rule_dedup/);
+
+    assert.ok(
+      addColIdx >= 0,
+      "migration must contain ADD COLUMN IF NOT EXISTS rule_id",
+    );
+    assert.ok(refIdx >= 0, "migration must contain partial index using rule_id");
+    assert.ok(
+      addColIdx < refIdx,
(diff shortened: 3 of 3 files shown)


blamechris:
## Code Review: PR #953

### Summary
This PR fixes a real, reproducible migration bug documented in issue #925: migration `0024_flag_dedup_constraint.sql` creates a partial unique index on `clinical_flags.rule_id`, but no prior migration adds that column. The initial schema migration `0000_initial_schema.sql` (lines 269-292) defines `clinical_flags` without `rule_id`, and no intervening migration adds it — confirmed by `grep -rn "rule_id" packages/db-schema/drizzle/*.sql` (only hits are inside 0024 itself). The Drizzle schema in `src/schema/ai-flags.ts:10` does declare `rule_id`, so existing dev/staging DBs presumably have the column from a stray `drizzle-kit push` or manual DDL, which is why the bug hadn't surfaced.

The fix — prepending `ALTER TABLE clinical_flags ADD COLUMN IF NOT EXISTS rule_id text;` — is idempotent and safe for both fresh and existing databases.

### Strengths
- **`IF NOT EXISTS` is the correct idempotency guard.** Existing DBs with the column already present are a no-op; fresh DBs get the column before the partial index references it.
- **Guard test locks in the ordering invariant.** `migration-0024.test.ts` statically asserts `ADD COLUMN` appears before `idx_flags_open_rule_dedup`, preventing regression if someone edits the SQL later.
- **Wiring the `test` script is a real win.** The three pre-existing test files (`encryption.test.ts`, `hmac.test.ts`, `mfa-secret-encryption.test.ts`) were silently not running in CI. After this PR, `pnpm test` -> `turbo test` picks them up (28 additional tests now exercised). Verified locally: 34/34 pass.
- **Test framework consistent** with the existing `node:test` + `node:assert/strict` convention used by the other files in this directory.
- **Column nullability matches semantics.** `rule_id text` (nullable, no default) aligns with `ai-flags.ts` schema and the dedup index's `WHERE ... AND rule_id IS NULL` branch for LLM-generated flags.

### Issues Found

#### Critical
None.

#### Suggestions
| `drizzle/0024_flag_dedup_constraint.sql` | 9 | Consider adding a brief note in the comment about why `IF NOT EXISTS` is safe (matches schema default, nullable, no backfill needed). | The existing comment explains why the ALTER is needed, but not why it's safe to run against DBs that already added the column via `drizzle-kit push` — future maintainers reviewing prod migration history will wonder if the column definitions could differ. |
| `src/__tests__/migration-0024.test.ts` | 17 | Add a second assertion verifying the column is declared as `text` (not `text NOT NULL`). | The dedup index's `WHERE ... AND rule_id IS NULL` branch depends on nullability; a typo adding `NOT NULL` would silently break LLM flag dedup without failing the current test. |

#### Nitpicks
- The filename prefix collision between `0024_flag_dedup_constraint.sql` and `0024_notification_urgency.sql` is pre-existing (not introduced by this PR) but worth noting. Drizzle's `_journal.json` disambiguates via `idx` (23 vs 24), so migrations apply correctly — just confusing to humans grepping by prefix.

### Deferred Items
None worth opening standalone issues — the two suggestions above are optional polish that can either be addressed in this PR or safely dropped.

### Verdict
**Approve.**

Surgical, well-scoped fix for a genuine bug blocking fresh-DB bootstrap. Idempotency is correct, the regression test is meaningful, and the side benefit of activating 28 previously-dormant encryption tests in CI is a clear win. Ready to merge once CI is green (currently Build / Test / Typecheck all passing).

Issue #946, opened 2026-04-19
fhir: MedicationRequest doseQuantity UCUM code validity (reject non-UCUM units like 'mg/kg')


blamechris:
## Context
Discovered during review of PR #923.

\`services/fhir-gateway/src/generators/medication-request.ts\` emits:

\`\`\`ts
doseQuantity: {
  value: medication.dose_amount,
  unit: medication.dose_unit,
  system: \"http://unitsofmeasure.org\",
  code: medication.dose_unit,
}
\`\`\`

It stamps UCUM as the system AND reuses \`dose_unit\` as the \`code\`. That works when \`dose_unit\` is a valid UCUM token (\`mg\`, \`mL\`, \`g\`, \`units\` — latter is debatable), but is wrong when the internal DB stores:

- \`mg/kg\` — not a UCUM atomic code (\`mg/kg\` happens to parse as a UCUM expression \`mg/kg\`, but downstream validators may require \`mg/kg\` with that exact syntax; safer to validate).
- \`tablet\`, \`puff\`, \`drop\` — NOT UCUM; the UCUM equivalents are \`{tbl}\`, \`{puff}\`, \`{drop}\` (curly-braced annotations).
- \`IU\` — UCUM uses \`[iU]\`.

Emitting a non-UCUM string under \`system: \"http://unitsofmeasure.org\"\` is a FHIR conformance violation — either the \`system\` or the \`code\` is wrong.

## Proposal
1. Maintain a UCUM-code allowlist in \`medication-request.ts\` (and \`medication-statement.ts\`).
2. When \`dose_unit\` matches, emit \`system\`+\`code\` as UCUM.
3. When it doesn't match, emit \`unit\` only (human-readable), omit \`system\` and \`code\`, OR map to the curly-brace UCUM annotation form.
4. Mirror the same logic in \`medication-statement.ts\`.

## Scope
- \`services/fhir-gateway/src/generators/medication-request.ts\`
- \`services/fhir-gateway/src/generators/medication-statement.ts\` (same bug likely present)
- Tests for each known unit variant.

PR #991, opened 2026-05-22
fix(fhir): read vital.unit, emit UCUM-correct atoms (#985)

### LINKED ISSUE #985: fix(fhir): observation generator hardcodes [degF]/[lb_av], ignores vital.unit — 37°C row exports as 37°F
## Evidence

`services/fhir-gateway/src/generators/observation.ts:59-68` unconditionally hardcodes the FHIR Quantity unit:

- `temperature` → `[degF]`
- `weight` → `[lb_av]`

The actual `vital.unit` value persisted in the DB is **ignored**. A row stored as `37, degC` exports as a FHIR Observation with `valueQuantity.value = 37, unit = [degF]`. A 70 kg weight exports as 70 lb.

## Why it matters

Clinical-safety P1. This is a sentinel-event-class bug. Any clinical system consuming our FHIR export (Epic, downstream registries, patient-share workflows) will read these values as Fahrenheit/pounds. A patient at 37°C (normal) reported as 37°F (severe hypothermia, lethal) is the most dangerous misalignment in clinical data.

`packages/medical-logic/src/lab-units.ts` exists for exactly this kind of unit-safety — but `services/fhir-gateway` doesn't use it on the export path.

## Suggested fix

1. Read `vital.unit` from the DB and pass it through to the FHIR Quantity unit field — use UCUM-compliant mappings (`Cel` for degC, `kg` for kg).
2. Add a conversion layer that detects when the stored unit and the FHIR-export-conventional unit differ and either converts (preserving the original system reading) or refuses to export.
3. Snapshot test with fixtures: `{value: 37, unit: "degC"}` → FHIR `{value: 37, unit: "Cel"}`; `{value: 70, unit: "kg"}` → `{value: 70, unit: "kg"}`.
4. Audit other generators in `services/fhir-gateway/src/generators/` for the same pattern.

## Source

Swarm audit 2026-05-21 (Medical Data agent). See `audit-results/2026-05-21-issue-hunt/05-medical-data.md`.

blamechris:
## Summary
- `generators/observation.ts` hardcoded the FHIR Quantity unit per VitalType (`temperature` → `[degF]`, `weight` → `[lb_av]`) and ignored the stored `vital.unit`. A row persisted as `{value: 37, unit: "degC"}` exported as `valueQuantity.code = [degF]` — a 37 °C reading downstream read as 37 °F (severe hypothermia)
- Replaced the `UCUM_UNITS` lookup with a `vitalQuantity(value, vital.unit)` helper that consults a stored-unit → UCUM atom map (case-insensitive). Recognised units emit a fully-coded Quantity; unrecognised units fall back to `value + unit` only (no false UCUM system claim)
- Added 9 test cases in `observation-vital-unit.test.ts`: Celsius/Fahrenheit isolation, kg/lbs isolation, BP/HR/SpO2/respiratory rate UCUM atoms, unknown unit fallback

## Audit of other generators
Surveyed `allergy-intolerance`, `condition`, `encounter`, `medication-request`, `medication-statement`, `patient`, `practitioner`, `procedure`. Only `observation.ts` had this pattern. `medication-*` already route through `ucum.ts::toDoseQuantity` (#946), which has the same "emit code only when valid UCUM" property.

## Test plan
- [x] `pnpm --filter @carebridge/fhir-gateway test observation-vital-unit` — 9/9 green
- [x] `pnpm --filter @carebridge/fhir-gateway test` — 195/195 green
- [x] `pnpm typecheck` — 40/40
- [x] `pnpm lint` — clean

Closes #985

### DIFF
--- a/services/fhir-gateway/src/__tests__/observation-vital-unit.test.ts
+++ b/services/fhir-gateway/src/__tests__/observation-vital-unit.test.ts
@@ -1,0 +1,132 @@
+/**
+ * Tests for vital unit propagation in FHIR Observation export (#985).
+ *
+ * Background
+ * ----------
+ * The original observation generator hardcoded the FHIR Quantity unit per
+ * vital type — temperature → [degF], weight → [lb_av] — and IGNORED the
+ * stored vital.unit. A row persisted as `{value: 37, unit: "degC"}`
+ * exported as `valueQuantity.value = 37, code = [degF]`. A patient at 37°C
+ * (normal) consumed by a downstream system (Epic, registries, patient
+ * share workflows) reads as 37°F — severe hypothermia, lethal. This is the
+ * single most dangerous misalignment in clinical data.
+ *
+ * These tests pin the contract: read vital.unit, emit UCUM-correct codes,
+ * never silently re-label a Celsius reading as Fahrenheit.
+ */
+
+import { describe, it, expect } from "vitest";
+import { toFhirVitalObservation } from "../generators/observation.js";
+import type { Vital } from "@carebridge/shared-types";
+
+function vital(overrides: Partial<Vital> & { type: Vital["type"]; value_primary: number; unit: string }): Vital {
+  return {
+    id: "vit-1",
+    patient_id: "pat-1",
+    recorded_at: "2026-05-21T12:00:00.000Z",
+    created_at: "2026-05-21T12:00:00.000Z",
+    updated_at: "2026-05-21T12:00:00.000Z",
+    ...overrides,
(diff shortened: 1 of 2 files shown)


blamechris:
Re: `vital.value_secondary ?? 0` — flagging this is right but the `?? 0` fallback is pre-existing on main (PR was unit-extraction only, didn't touch the secondary-value default). Tracking separately rather than expanding this PR's scope. The unit-mapping fix is the immediate clinical-safety win; the diastolic-zero default is the next layer to address.

Issue #932, opened 2026-04-19
Allergen synonyms: add vancomycin for Red Man syndrome coverage


blamechris:
## Context

Identified during review of PR #920 (issue #232 — allergen normalization).

## Description

The ALLERGEN_SYNONYMS table in `packages/medical-logic/src/allergen-synonyms.ts` covers penicillin, cephalosporin, sulfonamide, macrolide, fluoroquinolone, tetracycline, NSAID, APAP, opioid, heparin, warfarin, ACE-I, statin, benzodiazepine, iodinated contrast, and latex.

Vancomycin allergies (frequently charted alongside "Red Man syndrome" — which is technically a pseudoallergic histamine-release reaction, not true IgE, but charts do record it) are absent. Many inpatient charts list Vancomycin as an allergy, and other glycopeptides (teicoplanin — less common in US but present) cross-react. The current pass-through returns only "vancomycin" as its own alias list so the direct-match strategy still fires for a Vancomycin prescription, but brand aliases (Vancocin) and the class relationship don't exist.

## Acceptance criteria

- [ ] Add `vancomycin` canonical with aliases: `vancomycin`, `vanco`, `vancocin`, and glycopeptide class synonyms
- [ ] Consider whether Red Man syndrome should be tracked separately from true IgE vancomycin allergy (probably not worth distinguishing at the synonym layer — clinical judgement lives in the flag consumer)
- [ ] Test: charted "Vanco" allergy flags a Vancocin prescription via direct match

## References

- PR #920

Issue #232, opened 2026-04-07
Free-text allergen with no normalization or synonym matching


blamechris:
**File:** packages/db-schema/src/schema/patients.ts; services/fhir-gateway/src/generators/allergy-intolerance.ts:94-119

Allergen stored as encryptedText() free-form. 'PCN' vs 'penicillin', 'Lovenox' vs 'enoxaparin', 'Sulfa' vs 'sulfamethoxazole' won't match.

**Fix:** Allergen synonym tables; require/encourage SNOMED/RxNorm coding; brand→generic mapping.

PR #992, opened 2026-05-22
fix(clinical-data): single canonical CROSS_REACTIVITY_MAP in medical-logic (#986)

### LINKED ISSUE #986: fix(clinical-data): CROSS_REACTIVITY_MAP duplicated; writer copy missing iodine/contrast, latex, vancomycin
## Evidence

Two copies of `CROSS_REACTIVITY_MAP` exist with drift:

- **Writer** (`services/clinical-data/src/repositories/medication-repo.ts:11-78`) — silently missing entries for iodine/contrast, latex, vancomycin
- **AI rule** (`services/ai-oversight/src/rules/allergy-medication.ts:25-129`) — has the full set

## Why it matters

Clinical-safety P1. The drift means:

- At order time, the writer's allergy check passes ("no conflict") for a patient with a contrast allergy receiving an iodinated contrast order, a latex-allergic patient receiving a latex-containing device order, or a vancomycin-allergic patient receiving a glycopeptide.
- The AI oversight rule fires later, but only after the order is written.
- The cross-reactivities most missing from the writer are **the ones that cause intra-procedural anaphylaxis** — sentinel-event class.

Related to existing #977 (drug-class anchors dedup) and #973 (glycopeptide synonyms), but this is the writer-vs-rule **drift** specifically, not the rule expansion.

## Suggested fix

1. Extract the canonical map into a shared package (likely `@carebridge/medical-logic`) — a single source of truth both writer and AI rule consume.
2. Delete the writer-side and AI-rule-side copies; both import from the shared module.
3. Add a coverage test asserting iodine/contrast, latex, vancomycin paths block at the writer.
4. Cross-link #977 and #973 — those should consume the same canonical structure.

## Source

Swarm audit 2026-05-21 (Medical Data agent). Related but distinct from #977 and #973.
See `audit-results/2026-05-21-issue-hunt/05-medical-data.md`.

blamechris:
## Summary
- Two copies of `CROSS_REACTIVITY_MAP` had drifted — writer (12 entries) was silently missing iodine/contrast and latex that the AI rule (15 entries) had. Contrast and latex orders cleared the order-time block and only got flagged after the row was written
- Extracted the canonical 15-entry map to `@carebridge/medical-logic/cross-reactivity-map.ts`. Both writer and AI rule now import the single source of truth
- AI rule re-exports `CROSS_REACTIVITY_MAP` so the existing `cross-reactivity-sync.test.ts` resolves its prior import path without a churn change
- Writer's direct-match path now uses `expandAllergenAliases` (matching the AI rule's behaviour from #232) so a "Red Man Syndrome" allergy + vancomycin prescription blocks at order time

## Vancomycin nuance
The issue called out vancomycin as missing from the writer. The actual gap was the writer not using the synonym table — `allergen-synonyms.ts` already maps the "vancomycin" canonical class to `red man` / `glycopeptide` / `teicoplanin` (#932). Switching the writer's direct-match to operate against alias expansion closes the gap end-to-end without adding a new map entry (which would have required updating the LLM symmetry side too).

## Behaviour change captured in tests
One existing test was updated: amoxicillin against a penicillin allergy is now reported as "direct" (amoxicillin IS a penicillin) rather than "cross-reactivity via penicillin class". Same rejection, more accurate framing.

## Test plan
- [x] `pnpm --filter @carebridge/clinical-data test medication-repo` — 16/16 (4 new assertions on iodine/contrast, iodine→contrast advisory, latex, Red Man → vancomycin)
- [x] `pnpm --filter @carebridge/ai-oversight test cross-reactivity-sync` — 4/4 (LLM symmetry preserved via re-export)
- [x] `pnpm --filter @carebridge/ai-oversight test allergy-medication` — 13/13 (rule behaviour unchanged)
- [x] `pnpm --filter @carebridge/ai-oversight test` — 857/857
- [x] `pnpm --filter @carebridge/clinical-data test` — 45/45
- [x] `pnpm typecheck` — 40/40
- [x] `pnpm lint` — clean

Closes #986

### DIFF
--- a/packages/medical-logic/src/cross-reactivity-map.ts
+++ b/packages/medical-logic/src/cross-reactivity-map.ts
@@ -1,0 +1,126 @@
+/**
+ * Canonical allergen-medication cross-reactivity table.
+ *
+ * Issue #986: prior to this module the table was duplicated between the
+ * order-time allergy check (services/clinical-data) and the post-write
+ * AI oversight rule (services/ai-oversight). The two copies drifted —
+ * the writer was silently missing entries (iodine/contrast, latex) that
+ * the AI rule had, so contrast/latex orders cleared the writer's check
+ * and only got flagged AFTER the row was written. Sentinel-event class
+ * misses (intra-procedural anaphylaxis).
+ *
+ * Both writer and rule now import this single source of truth.
+ *
+ * Naming
+ * ------
+ * The `class` field is the canonical drug-class slug used by the
+ * cross-reactivity-sync test (services/ai-oversight/src/__tests__/
+ * cross-reactivity-sync.test.ts) to verify symmetry with the LLM prompt
+ * anchors in @carebridge/ai-prompts. Adding a new entry here requires a
+ * matching anchor on the LLM side (or a CANONICAL_NAME mapping in the
+ * sync test) — see feedback_cross_reactivity_sync_canonical.md.
+ */
+
+import type { FlagSeverity } from "@carebridge/shared-types";
+
+export interface CrossReactivityEntry {
+  allergenPattern: RegExp;
+  medicationPattern: RegExp;
+  /** Canonical class slug; matched against LLM anchors by the sync test. */
+  class: string;
+  /**
+   * Optional severity override. When present, AI rule consumers emit the
+   * cross-reactivity flag at this severity regardless of the charted
+   * allergy severity. Writer consumers ignore this field (they raise a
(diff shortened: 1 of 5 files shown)


blamechris:
Re: short-alias substring false-positives — confirmed real, but pre-existing in the AI rule since #232. The current PR moves the canonical map to medical-logic without changing the matching algorithm. Filed #994 to fix the substring-vs-boundary asymmetry across both consumers.

Issue #1089, opened 2026-05-23
Epic auth hardening follow-ups (deferred from #1084 review)


blamechris:
## Context
Two items from the Copilot review of PR #1084 were resolved with rationale but not implemented in-band — addressing them stand-alone keeps the Epic stack mergeable.

### 1. EpicTokenClient — concurrent in-flight token coalescing
**Source:** [thread on token-client.ts:85](https://github.com/blamechris/carebridge/pull/1084#discussion_r-PRRT_kwDOR6YwEc6EQswH)

Today, if `getAccessToken()` is called concurrently before the first fetch resolves, multiple token requests fly to Epic. In production the Backend Services worker is single-instance per deployment so concurrent calls within a tick aren't on the hot path, but horizontal scaling will surface this.

**Acceptance criteria:**
- [ ] Track an in-flight `Promise<TokenResponse>` on the client
- [ ] Concurrent callers await the same in-flight promise
- [ ] `invalidate()` clears the in-flight reference too
- [ ] Add a unit test that verifies a single fetch for N concurrent callers

### 2. PEM validation hardening at load time
**Source:** [thread on config.ts:45](https://github.com/blamechris/carebridge/pull/1084#discussion_r-PRRT_kwDOR6YwEc6EQswd)

`loadEpicConfig` currently checks for PEM-header substrings. A non-RSA PKCS#8 or encrypted key sneaks through and fails later during signing with a confusing error.

**Acceptance criteria:**
- [ ] Parse the PEM via `crypto.createPrivateKey({ key, format: "pem" })` at config load
- [ ] Assert `asymmetricKeyType === "rsa"` and surface a clear error message otherwise
- [ ] Add a unit test for: valid RSA PKCS#8, valid RSA PKCS#1, EC key (rejected), encrypted PEM (rejected)

PR #1152, opened 2026-05-24
fix(epic): token-client generation counter (#1143)

### LINKED ISSUE #1143: epic-connector: token-client stale-write race after invalidate() during in-flight refresh
## Context

Identified during review of PR #1136 (Copilot reviewer, services/epic-connector/src/token-client.ts).

## Description

`EpicTokenClient.invalidate()` clears `this.inFlight`, but the dropped promise can still resolve after the new fetch and overwrite `this.cached` with the stale-credential token. The 401 retry path in `EpicFhirClient` (fhir-client.ts:358-367) goes:

1. Call A reaches `getAccessToken()`, sets `inFlight = P1`, awaits fetch.
2. Server replies 401; fhir-client calls `tokens.invalidate()` (clears `inFlight` and `cached`).
3. fhir-client retries — Call B starts a new fetch (`P2`), populates `cached` with the fresh token.
4. P1 resolves *after* P2 — the `fetchToken` body unconditionally writes `this.cached = { response: parsed, ... }`, reverting cache to the stale token.

The next caller now reads the stale token from cache and 401s again.

### Suggested fix

Track a monotonic `generation` (or per-fetch token) that `invalidate()` increments. Capture it at fetch start and only write `this.cached` when it still matches at resolve time.

```ts
private generation = 0;
invalidate() { this.cached = null; this.inFlight = null; this.generation++; }
private async fetchToken(...) {
  const myGen = this.generation;
  // ...
  if (myGen === this.generation) {
    this.cached = { response: parsed, expiresAtMs: ... };
  }
  return parsed;
}
```

Also tighten the test in `token-client.test.ts` (~line 221): currently resolves the invalidated first fetch *before* the second, so the bug stays hidden. Flip the resolve order and assert `getAccessToken()` afterwards returns the second token.

## Original Review Comment

> \`invalidate()\` clears \`inFlight\`, but the invalidated in-flight request can still resolve later and overwrite \`this.cached\` here. If a 401 triggers \`invalidate()\` + a new fetch, the older promise resolving after the newer one would revert the cache to a stale token, defeating the retry hardening. Consider tracking a monotonic \"generation\"/requestId that \`invalidate()\` increments and only writing \`this.cached\` when the generation captured at fetch start still matches.

## Acceptance Criteria

- [ ] `EpicTokenClient` carries a generation counter incremented by `invalidate()`.
- [ ] `fetchToken` writes to `this.cached` only when its captured generation still matches.
- [ ] Test resolves second (newer) fetch before first (invalidated) fetch and asserts cache is not reverted to the stale token.
- [ ] Existing 225 epic-connector tests still pass.

blamechris:
Closes #1143

## The race

`EpicTokenClient.invalidate()` clears `this.inFlight` and `this.cached`, but the in-flight fetch promise it dropped is still scheduled to resolve. Its body unconditionally writes `this.cached = { response, expiresAtMs }` once the network resolves.

Sequence:
1. Call A starts -> `inFlight = P1`, awaits fetch.
2. Server 401 elsewhere -> fhir-client invokes `tokens.invalidate()` -> `inFlight = null`, `cached = null`.
3. Retry: Call B starts -> `inFlight = P2`. P2 resolves first, sets `cached = {fresh}`.
4. P1 finally resolves and overwrites `cached = {stale}`.
5. Next caller reads stale -> 401 loop.

## The fix

Monotonic generation counter on the client:
- `invalidate()` bumps `this.generation`.
- `getAccessToken()` captures `myGen = this.generation` before assigning `this.inFlight = pending`. The `finally` clearer only nulls the slot when `this.generation === myGen && this.inFlight === pending` so a newer fetch isn't stomped.
- `fetchToken()` receives the generation it started under and only writes `this.cached = ...` when `this.generation === startedAtGen`. Stale resolves are dropped on the floor.

The fetch is still awaited so its promise can settle / be observed by the original caller; only the side-effect write to shared cache state is gated.

## Tests

- New `#1143` test simulates the exact race: P1 in-flight -> `invalidate()` -> P2 starts and resolves with fresh token first -> P1 resolves with stale token last. Asserts the next `getAccessToken()` reads the fresh token from cache (and that no third fetch is issued).
- The existing `#1089` "invalidate clears in-flight" test was tightened — it previously resolved the invalidated first fetch BEFORE the second, which let stale writes go undetected. The fresh fetch now resolves first; the invalidated one resolves last; a follow-on call must still see the fresh token cached.

Full epic-connector suite: 240/240 pass. `pnpm typecheck` and `pnpm lint` both green from the worktree root.

### DIFF
--- a/services/epic-connector/src/__tests__/token-client.test.ts
+++ b/services/epic-connector/src/__tests__/token-client.test.ts
@@ -188,6 +188,13 @@ describe("EpicTokenClient (#389)", () => {
     // Set up a pending fetch that we will invalidate before resolving.
     // After invalidate(), the next call should kick off its own fetch
     // rather than reusing the dropped in-flight promise.
+    //
+    // CRITICAL ordering (#1143): we resolve the SECOND fetch FIRST and
+    // the FIRST (invalidated) fetch LAST. This is the real-world race —
+    // the dropped promise often settles after the retry has already
+    // populated the cache with a fresh token. Without a generation
+    // guard, the late-resolving first fetch overwrites the fresh cache
+    // with stale data and the next caller reads a stale token.
     let resolveFirst: (response: Response) => void;
     let resolveSecond: (response: Response) => void;
     const firstPending = new Promise<Response>((resolve) => {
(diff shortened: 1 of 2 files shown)


blamechris (commented):
**Verdict: APPROVE — Critical: 0 | Suggestions: 1 | Nits: 1**

Reviewed as CareBridge Inspector against the clinical-safety bar. This PR correctly closes the stale-write race documented in #1143. The fix is minimal, tightly scoped to `token-client.ts`, and the test coverage actually exercises the bug.

### Race-correctness trace

Walked through the documented sequence step by step:

1. Call A enters `getAccessToken`, sees no cache, captures `myGen=0`, sets `inFlight=P1`, awaits.
2. Server 401 elsewhere → `tokens.invalidate()` fires → `cached=null`, `inFlight=null`, `generation=1`.
3. Call B enters, no cache/inflight, captures `myGen=1`, sets `inFlight=P2`, awaits.
4. P2 resolves fresh first → in `fetchToken`, `this.generation===1===startedAtGen` → writes `cached={fresh}`. Back in `getAccessToken` finally: `this.generation===1===myGen && this.inFlight===P2===pending` → clears `inFlight=null`. Correct.
5. P1 resolves stale last → in `fetchToken`, `this.generation===1 !== startedAtGen===0` → cache write **skipped**. Back in `getAccessToken` finally: `this.generation===1 !== myGen===0` → identity check fails (or generation check fails first), `inFlight` left alone (already null). Correct.
6. Next caller: reads fresh from cache. Race closed.

Reverse-order (P1 resolves first) and the case where P1's `getAccessToken` finally would have run before P2 was even fired both also trace correctly — generation guard catches the stale write, identity guard on `inFlight` prevents a stale fetch from nulling a new in-flight slot.

JS single-threadedness means no interleaving between the `const pending = this.fetchToken(...)` and `this.inFlight = pending` lines, and no interleaving inside the synchronous finally block. Good.

### Cache-write protection

Grepped `this.cached` writes: only two — line 202 (gated by generation) and line 219 (intentional null in `invalidate()`). All write paths covered.

### Test rigor

- The new `#1143` test resolves P2 first, then P1, and crucially asserts the **subsequent** `getAccessToken` reads `fresh-token` from cache. Mentally ran it against un-patched code: P1's late resolve would write `cached={stale-token}` and the next caller would read stale → test fails. The test does discriminate.
- Tightened `#1089` test now resolves second-first-first-last and also adds the post-settle subsequent read assertion. Good defensive ratchet.

### Identity equality on `inFlight` clear

Belt-and-suspenders: both `this.generation === myGen` AND `this.inFlight === pending`. The generation check alone would suffice for the documented race, but the identity check protects against a future variant where `invalidate()` isn't called but two `getAccessToken()` calls somehow race for the slot — defensive and cheap. No concern.

### API surface & conventions

- No public-API change. `EpicTokenClient` constructor and `getAccessToken` / `invalidate` signatures unchanged.
- `private generation = 0` and `startedAtGen: number` parameter are internal.
- ESM `.js` import suffix preserved.
- Strict TS compiles clean (token-client.ts alone is error-free; remaining workspace errors are unrelated pre-build state).
- Zero Attribution Policy: no Claude/Anthropic references in code or commit message.

### Joined-caller observation (NOT a bug, worth understanding)

A concurrent coalesced caller that joined `P1` via the `if (this.inFlight) { ... await this.inFlight ... }` branch BEFORE `invalidate()` fired will still receive the stale token from P1's eventual resolve, since the joiner doesn't go through `fetchToken`. That caller will then make an Epic call with a stale token, get 401, and re-trigger the invalidate/retry cycle.

The PR body acknowledges this: "The fetch is still awaited so its promise can settle / be observed by the original caller; only the side-effect write to shared cache state is gated." Acceptable — no PHI corruption, just one extra 401 retry. Worth documenting in a follow-up if seen in production logs, but not a blocker.

### Suggestion (non-blocking): observability for dropped writes

When `fetchToken` skips a cache write because the generation advanced, the dropped write is silent. In a real outage (Epic returning intermittent 401s, lots of invalidate churn), it would be useful to log a debug/warn line so operators can see how often stale writes are being suppressed. Something like:

```ts
if (this.generation === startedAtGen) {
  this.cached = { response: parsed, expiresAtMs: issuedAtMs + parsed.expires_in * 1000 };
} else {
  log.debug("Dropping stale token-fetch result; cache invalidated mid-flight", {
    startedAtGen,
    currentGen: this.generation,
  });
}
```

Not blocking — happy to file a from-review issue if you prefer to track separately.

### Nit (non-blocking)

Line 122 comment says "fetchToken() checks that and refuses to write back stale results to either the cache or the in-flight slot." Strictly speaking, `fetchToken` only guards the **cache** write; the in-flight slot is guarded in `getAccessToken`'s finally. The comment slightly overpromises what `fetchToken` does. Tiny.

### Verdict

LGTM. Race correctness verified, tests are real regressions, no API regression, conventions clean. Approving — both suggestion and nit are optional and can land in a follow-up.

PR #1158, opened 2026-05-24
feat(infra): patient-education URL rot detection (#1151)

### LINKED ISSUE #1151: chore(portal): periodic check for patient-education URL rot
## Context

Identified during review of PR #1139 (curated MedlinePlus/CDC/AHA links on patient-education cards).

## Description

The 31 curated URLs in `packages/medical-logic/src/patient-education.ts` are well-formed today, but authoritative health domains do reorganize (CDC has historically reshuffled `/index.html` vs `/default.htm` paths; AHA's `/en/health-topics/...` tree has churned). The invariant test in `packages/medical-logic/src/__tests__/patient-education.test.ts` only validates URL shape and domain, not reachability — so a 404 would ship silently.

## Original Review Comment

> AHA CAD URL is unusually deep — higher than average link-rot risk on AHA reorgs.
> Recommend periodic check that authoritative URLs do drift (especially CDC reorgs).

## Acceptance Criteria

- [ ] Decide automation strategy: GH Actions cron (e.g. monthly) doing HEAD requests, or quarterly manual sweep tracked as a recurring issue.
- [ ] If automated: add a workflow that HEAD-fetches every URL in `DIAGNOSIS_EDUCATION_TABLE[*].links[*].url` and opens an issue on any non-2xx.
- [ ] Document the chosen process in `packages/medical-logic/README.md` or a comment in `patient-education.ts`.

blamechris:
Closes #1151

## Summary
- New monthly GitHub Actions workflow (`.github/workflows/patient-education-url-check.yml`) HEAD-checks every curated `links[*].url` in `DIAGNOSIS_EDUCATION_TABLE` + `MEDICATION_EDUCATION_TABLE` and opens (or comments on) a tracking issue when any URL returns non-2xx.
- Script lives at `packages/medical-logic/scripts/check-education-urls.ts`, runnable via `pnpm --filter @carebridge/medical-logic run check:education-urls`. Sequential, 10s timeout per URL, follows redirects, falls back from HEAD to GET on 405 (CDN-fronted health sites like AHA reject HEAD).
- Idempotent issue handling: workflow searches for an open issue whose title starts with `fhir: patient-education URL rot detected` and comments on the existing one instead of creating duplicates. The `url-rot` label is created at workflow runtime if missing; `from-review` already exists.
- 14 new unit tests with mocked fetch cover `classifyStatus`, `checkUrl` (HEAD pass, HEAD-405-then-GET success, HEAD-405-then-GET fail, non-405 HEAD fail, network error, timeout), `formatReport` (clean run + failure table + pipe-escape), and `collectCuratedUrls` (dedupe + no-links case). No live network in CI — the cron is intentionally separate.
- Header comment on `patient-education.ts` documents the monthly check and the authoritative-domain convention for future curated links.

## Test plan
- [x] `pnpm typecheck` — 38/38 passing
- [x] `pnpm --filter @carebridge/medical-logic test` — 502/502 passing (14 new)
- [x] `pnpm --filter @carebridge/medical-logic build` — clean dist, no scripts/ bleed
- [x] `pnpm lint` — clean (pre-existing portal warnings unchanged)
- [x] Local smoke run hit MedlinePlus URLs and returned 200 via HEAD
- [ ] First scheduled run on the 1st of next month (or trigger manually via Actions tab)

### DIFF
--- a/.github/workflows/patient-education-url-check.yml
+++ b/.github/workflows/patient-education-url-check.yml
@@ -1,0 +1,97 @@
+name: Patient Education URL Check
+
+on:
+  schedule:
+    # 09:00 UTC on the 1st of every month.
+    - cron: '0 9 1 * *'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  issues: write
+
+concurrency:
+  group: patient-education-url-check-${{ github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  check-urls:
+    name: Check curated education URLs
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v4
+      - uses: pnpm/action-setup@v4
+        with:
+          version: 9.15.4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          cache: pnpm
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Run URL rot check
+        id: check
+        run: |
+          set +e
+          pnpm --filter @carebridge/medical-logic run check:education-urls \
+            > url-check.log 2>&1
+          status=$?
+          cat url-check.log
+          echo "exit_status=$status" >> "$GITHUB_OUTPUT"
+          # The script already appends to $GITHUB_STEP_SUMMARY itself,
+          # so no extra summary step needed here.
+          exit 0
+
+      - name: Open or update tracking issue on failure
+        if: steps.check.outputs.exit_status != '0'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -euo pipefail
+
+          title="fhir: patient-education URL rot detected (auto)"
+          body_file=$(mktemp)
+          {
+            echo "Automated URL rot check found one or more curated"
+            echo "patient-education URLs that did not respond with a 2xx."
+            echo ""
+            echo "Workflow run: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+            echo ""
+            echo '```'
+            cat url-check.log
+            echo '```'
+          } > "$body_file"
+
+          # Make sure the url-rot label exists (no-op if already there).
+          # `from-review` is created by other workflows already.
+          gh label create url-rot \
+            --color "B60205" \
+            --description "Detected URL rot in curated content" \
+            2>/dev/null || true
+
+          # Idempotency: if an open issue with the canonical title prefix
+          # already exists, comment on it instead of opening a duplicate.
+          existing=$(gh issue list \
+            --state open \
+            --search "in:title \"fhir: patient-education URL rot detected\"" \
+            --json number,title \
+            --jq '.[] | select(.title | startswith("fhir: patient-education URL rot detected")) | .number' \
+            | head -n1)
+
+          if [ -n "$existing" ]; then
+            echo "Updating existing issue #$existing"
+            gh issue comment "$existing" --body-file "$body_file"
+          else
+            echo "Creating new tracking issue"
+            gh issue create \
+              --title "$title" \
+              --label "url-rot,from-review" \
+              --body-file "$body_file"
+          fi
+
+      - name: Fail job if URL check failed
+        if: steps.check.outputs.exit_status != '0'
+        run: exit 1
--- a/packages/medical-logic/package.json
+++ b/packages/medical-logic/package.json
@@ -13,14 +13,17 @@
   },
   "scripts": {
     "build": "tsc",
-    "typecheck": "tsc --noEmit",
+    "typecheck": "tsc --noEmit -p tsconfig.check.json",
     "test": "vitest run",
-    "clean": "rm -rf dist"
+    "clean": "rm -rf dist",
+    "check:education-urls": "tsx scripts/check-education-urls.ts"
   },
   "dependencies": {
     "@carebridge/shared-types": "workspace:*"
   },
   "devDependencies": {
+    "@types/node": "^22.0.0",
+    "tsx": "^4.19.0",
     "typescript": "^5.7.0",
     "vitest": "^3.0.0"
   }
--- a/packages/medical-logic/scripts/check-education-urls.ts
+++ b/packages/medical-logic/scripts/check-education-urls.ts
@@ -1,0 +1,230 @@
+/**
+ * Patient-education URL rot detector (issue #1151).
+ *
+ * Walks every `links[*].url` in DIAGNOSIS_EDUCATION_TABLE +
+ * MEDICATION_EDUCATION_TABLE and HEADs each one. Health-domain CDNs
+ * (notably the Cloudflare front in front of `heart.org`) sometimes reject
+ * HEAD with 405, so we fall back to a GET on 405. Followed redirects up
+ * to 5 hops count as PASS.
+ *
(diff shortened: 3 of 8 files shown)


blamechris (commented):
## CareBridge Inspector — review of PR #1158

Closes #1151. Infra/automation review focused on idempotency, resilience, and security. PR is **APPROVE** — sound design, comprehensive tests, no security or correctness blockers. A handful of small nits below; happy to land as-is and have me file from-review follow-ups for the optional polish.

### Strengths

**Workflow YAML (`patient-education-url-check.yml`)**
- Cron `0 9 1 * *` parses to 09:00 UTC on the 1st of every month — correct.
- `permissions:` scoped minimally to `contents: read, issues: write`. No write access to anything else.
- Action pins (`actions/checkout@v4`, `pnpm/action-setup@v4` v9.15.4, `actions/setup-node@v4`, Node 22) line up exactly with `.github/workflows/ci.yml`.
- `concurrency:` group with `cancel-in-progress: false` is the right call for a monthly cron — overlapping triggers from `workflow_dispatch` shouldn't race-cancel each other.
- `timeout-minutes: 15` is comfortable: 31 URLs * 10s worst-case = ~5min, double that for 405-retries on every URL still fits.
- The step layout (`set +e` capture → conditional issue creation → final fail step) cleanly separates the run-no-matter-what reporting from the eventual job-fail signal.
- Idempotency is **double-defensive**: the `gh issue list --search 'in:title "fhir: patient-education URL rot detected"'` query is a substring match (so safe-by-default), then `--jq '... | select(.title | startswith(...))'` enforces a literal prefix match. A future issue titled "fhir: patient-education URL rot detected (auto, q2-2026)" would still be matched and commented on rather than duplicated. 
- `gh label create url-rot ... || true` makes label creation best-effort; if the label already exists or the API hiccups, the issue still gets created (the subsequent `gh issue create --label "url-rot,from-review"` will fail if `url-rot` truly doesn't exist, but in practice the create-or-noop sequencing eliminates that). Solid.
- `GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}` paired with the explicit `issues: write` permission — gh CLI authenticates correctly without any extra secret.
- Zero attribution in the YAML — confirmed.

**Script (`check-education-urls.ts`)**
- HEAD-then-GET-on-405 semantics are clean: `attempt("HEAD")` returns a sentinel result on 405, and the outer driver re-invokes with GET. Easier to test than a recursive call (the suite exploits exactly this).
- AbortController + `setTimeout` + `clearTimeout` in `finally` — correct timeout pattern, no leaked timers.
- `redirect: "follow"` so redirect chains resolve to a terminal response; sequential per-URL keeps us off any host's "too many concurrent requests" radar — comment explains this nicely.
- Polite user-agent identifying the bot — good citizenship for the curated health-domain hosts.
- Markdown report is plaintext-safe (renders cleanly in step summary AND email notifications). Pipe-escape in failure rows is a nice touch.
- `import.meta.url === pathToFileURL(process.argv[1]).href` guard prevents `main()` from running when the file is imported by the vitest suite — works for both `tsx` and a future `node dist/...` invocation.

**Tests (`check-education-urls.test.ts`)**
- 14 tests covering: `classifyStatus` (2xx/405/4xx/3xx), `checkUrl` (HEAD pass, HEAD-405→GET pass, HEAD-405→GET fail, non-405 HEAD fail, network error, AbortController timeout), `formatReport` (clean run, failure table, pipe-escape), `collectCuratedUrls` (dedupe + no-links).
- All use mocked `fetch` via `as unknown as typeof fetch` — **no live network in CI**, exactly as advertised.
- The timeout test honours the abort signal (rejects the in-flight fetch when aborted) so it doesn't hang the suite. Nice.

**Other**
- `tsconfig.check.json` follows the existing `packages/ai-prompts/tsconfig.check.json` pattern. Building shared-types first → `pnpm --filter @carebridge/medical-logic typecheck` clean.
- `@types/node ^22` and `tsx ^4.19` added as devDependencies (correct scope — script is a dev/CI tool, not runtime).
- `patient-education.ts` doc block is informative, not noisy: documents the cron job + authoritative-domain allowlist so future link adds know the invariant.
- Local verification in this worktree: `pnpm --filter @carebridge/medical-logic test` → 502/502 passing (14 new); `pnpm --filter @carebridge/medical-logic typecheck` clean; `pnpm lint` clean (pre-existing portal warnings unchanged).

### Nits (non-blocking, happy to file as from-review follow-ups)

**N1 — dead `MAX_REDIRECTS` constant and stale doc comment.**
Line 26: `const MAX_REDIRECTS = 5;` is declared and exported in spirit but never used. The doc comment on the file header and on `classifyStatus` both claim "Followed redirects up to 5 hops count as PASS" / "fetch follows them by default, but we cap at MAX_REDIRECTS" — neither is true. Node's `fetch` follows redirects up to its own internal default (20), and there's no cap wired in here. Two clean options: (a) drop the constant and reword the comments to match reality ("we delegate redirect-following to Node's fetch default"), or (b) actually wire a manual redirect cap using `redirect: "manual"` and a loop. Option (a) is the minimal change.

**N2 — 3xx classification comment doesn't match Node fetch semantics.**
`classifyStatus` comment claims "3xx only reaches this function when redirects are exhausted, so a leftover 3xx is treated as a redirect-loop failure." With `redirect: "follow"`, Node fetch resolves the chain and returns the terminal response — a leftover 3xx hitting `classifyStatus` would mean the server returned a 3xx that isn't a redirect (rare) or a redirect-loop exceeded the internal cap (which throws rather than returning). Treating 3xx as fail is still correct; the comment just over-states the mechanism. Tighten or drop the explanation.

**N3 — `tsconfig.check.json` diverges from the `ai-prompts` precedent on test exclusion.**
The new `packages/medical-logic/tsconfig.check.json` overrides `include: ["src", "scripts"]` without excluding `src/__tests__`, which pulls tests back into the typecheck (`tsconfig.json` excluded them). `packages/ai-prompts/tsconfig.check.json` explicitly re-excludes `src/__tests__`. Currently this works because vitest types are available via the package, but for consistency consider adding `"src/__tests__"` to the `exclude` list (or leaving as-is if the intent is to typecheck tests too — in which case worth a one-line comment). 

**N4 — gh search could be label-scoped for marginal efficiency.**
`gh issue list --search 'in:title "..."'` scans all repo issues; adding `label:url-rot` to the search would scope it. With the `--jq startswith` post-filter the current code is already safe; this is purely a perf nit on a monthly cron, so probably not worth a follow-up.

**N5 — failure-detection contract: issue body includes the full log.**
The shell `cat url-check.log` is dropped inside a fenced code block in the issue body. For 31 URLs this is fine, but if a future curated set expands and a single run produces a huge log, the issue body could approach GitHub's 65,536-character limit. Not a now-problem; just worth a mental bookmark.

### Security

- No secrets exfiltrated; the only secret in use is the built-in `GITHUB_TOKEN`.
- Script doesn't execute user-controlled code — all URLs come from a compile-time constant in `patient-education.ts`.
- gh CLI calls are restricted to the workflow's `issues: write` permission.
- No untrusted-input → shell injection: the title is a hardcoded literal, the body is built from a fixed prefix + a captured-script log + a GitHub-supplied run URL. The log content itself is HEAD responses from curated authoritative-health domains, written into a fenced block, so even if a server returned weird bytes they couldn't escape the bash heredoc / body file.
- Zero attribution: confirmed across YAML, script, and tests.

### Verdict

LGTM. Ship as-is. I'll file two small from-review follow-ups (N1 dead constant + comment, N3 test-exclude parity) and leave N2/N4/N5 as inline polish or future work.

PR #1159, opened 2026-05-24
fix(epic): token-client scope-keyed coalescing (#1144)

### LINKED ISSUE #1144: epic-connector: token-client scope-mismatch coalescing — inFlight not keyed by scopes
## Context

Identified during review of PR #1136 (Copilot reviewer, services/epic-connector/src/token-client.ts).

## Description

`EpicTokenClient.getAccessToken(scopes)` accepts a per-call scope override (defaults to `DEFAULT_SYSTEM_SCOPES`). The new `inFlight` slot coalesces concurrent callers regardless of the `scopes` argument, so:

- Caller A: `getAccessToken(["system/Patient.read"])` — starts fetch with narrow scope, sets `inFlight`.
- Caller B (same tick): `getAccessToken(["system/Patient.read", "system/Observation.read"])` — observes `inFlight`, awaits A's promise, receives a token *missing required scopes*.

The cache itself has the same shape — `this.cached` is unkeyed and any subsequent caller with different scopes reuses whatever was last fetched.

Currently latent: the only production caller (`fhir-client.ts:345`) uses defaults and never passes a scope override. The bug would only fire if a future caller exercises the documented `scopes` parameter.

### Suggested fix

Either:

1. **Remove the documented `scopes` override** if multi-scope per-tenant isn't a real requirement (simpler, matches actual usage).
2. **Key `inFlight` and `cached` by a normalized (sorted, joined) scopes string**:

```ts
private inFlight: Map<string, Promise<TokenResponse>> = new Map();
private cached: Map<string, CachedToken> = new Map();
const key = [...scopes].sort().join(" ");
```

Option 1 is simpler given current usage; option 2 preserves the public API.

## Original Review Comment

> The single \`inFlight\` slot coalesces all concurrent callers regardless of the \`scopes\` argument. If two callers request different scopes concurrently (or one uses a custom override), the later caller will await the first request and may receive a token missing required scopes. If \`scopes\` is intended to be supported, consider keying \`inFlight\` (and ideally the cache) by a normalized scopes string, or only coalescing when requested scopes match.

## Acceptance Criteria

- [ ] Decision recorded: keep scope override (and key inFlight+cache by scope) OR remove the parameter.
- [ ] If kept: test covers two concurrent callers with disjoint scope sets receiving independent tokens.
- [ ] If removed: `getAccessToken()` signature dropped, `DEFAULT_SYSTEM_SCOPES` becomes the only path, JSDoc updated.

blamechris:
## Summary

Keys `inFlight` and `cached` in `EpicTokenClient` by normalized (sorted+joined) scopes string so concurrent callers with different scope sets do not share a token and unkeyed cache cannot serve a wrong-scope token to a subsequent caller.

Builds on the generation counter from #1143 (stale-write guard now operates per-key).

## Test plan

- [ ] Two concurrent fetches with different scopes -> two `fetch` calls, each correct
- [ ] Two concurrent fetches with same scopes (any order) -> coalesces to one
- [ ] `invalidate()` clears all keys
- [ ] Generation guard from #1143 still works per-key

Closes #1144
Depends on #1152

### DIFF
--- a/services/epic-connector/src/__tests__/token-client.test.ts
+++ b/services/epic-connector/src/__tests__/token-client.test.ts
@@ -281,6 +281,195 @@ describe("EpicTokenClient (#389)", () => {
     expect(fetchMock).toHaveBeenCalledTimes(2);
   });
 
+  it("does not coalesce concurrent callers requesting different scope sets (#1144)", async () => {
+    // Two concurrent callers with different scope sets must each receive
+    // a token fetched with their own scopes — not be served whichever
+    // narrow-scope token happened to be in flight first. Otherwise a
+    // caller that requested ["system/Patient.read", "system/Observation.read"]
+    // could receive a token issued only for ["system/Patient.read"] and
+    // then fail mid-call when it tried to read Observation resources.
+    let resolveNarrow: (response: Response) => void;
+    let resolveWide: (response: Response) => void;
+    const narrowPending = new Promise<Response>((resolve) => {
+      resolveNarrow = resolve;
+    });
+    const widePending = new Promise<Response>((resolve) => {
+      resolveWide = resolve;
+    });
+    const fetchMock = vi
+      .fn()
+      .mockImplementationOnce(() => narrowPending)
+      .mockImplementationOnce(() => widePending);
+    const client = new EpicTokenClient(makeConfig(privatePem), fetchMock);
+
+    const narrowCall = client.getAccessToken(["system/Patient.read"]);
+    const wideCall = client.getAccessToken([
+      "system/Patient.read",
+      "system/Observation.read",
+    ]);
+
+    // Each scope set must drive its OWN fetch — coalescing would
+    // collapse these to one.
+    expect(fetchMock).toHaveBeenCalledTimes(2);
+
+    resolveNarrow!(tokenResponse({ access_token: "narrow-token" }));
+    resolveWide!(tokenResponse({ access_token: "wide-token" }));
+
+    expect(await narrowCall).toBe("narrow-token");
+    expect(await wideCall).toBe("wide-token");
+
+    // Verify each fetch carried the correct scope string.
+    const narrowBody = new URLSearchParams(
+      fetchMock.mock.calls[0]![1].body as string,
+    );
+    const wideBody = new URLSearchParams(
+      fetchMock.mock.calls[1]![1].body as string,
+    );
+    expect(narrowBody.get("scope")).toBe("system/Patient.read");
+    expect(wideBody.get("scope")?.split(" ").sort()).toEqual(
+      ["system/Observation.read", "system/Patient.read"],
+    );
+  });
+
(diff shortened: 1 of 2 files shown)


blamechris (commented):
## CareBridge Inspector Review — PR #1159 (delta-only)

**Delta-only review against PR #1152 base — full chain reviewed in #1152.**

Scope of this review is the additional changes made on top of `feat/issue-1143-token-client-gen-counter` (PR #1152): converting `cached` and `inFlight` from single slots to `Map<string, ...>` keyed by `[...scopes].sort().join(" ")`.

### Verdict: APPROVE

### What I checked

- **Scope normalization** — `[...scopes].sort().join(" ")` is stable: sorting on the spread copy avoids mutating caller-owned arrays; sort is lexicographic and deterministic; the empty-array case maps to `""` (a valid distinct key, consistent with pre-existing behaviour of passing through to the form-body `scope` field).
- **Map mutation safety** — All `set` / `get` / `delete` paths look correct. The finally block in `getAccessToken` (`if (this.generation === myGen && this.inFlight.get(key) === pending) this.inFlight.delete(key);`) preserves the identity check from #1152, so an `invalidate()`+restart sequence will not have the older fetch erase the newer one's slot.
- **Generation × Map composition** — `invalidate()` clears both maps and bumps the global generation counter. A stale fetch resolving later sees `this.generation !== startedAtGen` and refuses to write back to ANY key (no cross-key pollution). The new test `generation guard works per scope key` exercises exactly that race.
- **Memory** — Maps are bounded by the number of distinct scope sets per `EpicTokenClient` instance, which is finite per service (one production caller currently uses defaults only). Comments call this out; no action needed.
- **Test coverage** — Five new tests are meaningful and disjoint:
  - different-scope concurrent → 2 fetches with correct scope bodies (asserts the form-body `scope` field, not just call count)
  - same-scope shuffled → 1 fetch, both callers share a result
  - sequential per-scope isolation → cache hit only re-serves matching scope
  - `invalidate()` clears every key
  - generation guard remains per-key safe after invalidate
- **Public API** — `getAccessToken(scopes: string[] = DEFAULT_SYSTEM_SCOPES): Promise<string>` signature unchanged.
- **Stack discipline** — Base is correctly `feat/issue-1143-token-client-gen-counter`, not `main`. Body documents `Depends on #1152`.
- **Zero attribution** — Clean (commit + diff scanned).

### Nit (non-blocking)

A truly duplicated input like `["system/Patient.read", "system/Patient.read"]` normalizes to the key `"system/Patient.read system/Patient.read"`, which is a distinct cache entry from `["system/Patient.read"]` even though both request the same scope set. Not a correctness issue (Epic returns a valid token either way and the production caller doesn't pass duplicates), but a `new Set()` pass before sort would coalesce these. Filing as a follow-up rather than blocking; not worth churn on a stacked PR.

### Result

LGTM. Cannot merge until parent #1152 lands.

PR #1169, opened 2026-05-24
feat(epic): observability for dropped stale cache writes (#1161)

### LINKED ISSUE #1161: epic-connector: log dropped cache writes when invalidate() fires mid-flight (token-client observability)
## Context

Identified during review of PR #1152 (token-client generation counter for stale-write race).

## Description

When `EpicTokenClient.fetchToken` skips a cache write because the generation has advanced (i.e. `invalidate()` fired while the fetch was in flight), the dropped write is silent. In a real outage scenario — Epic returning intermittent 401s, lots of invalidate churn — operators would have no signal that stale writes are being suppressed, only that token requests are spiking.

A debug or warn log line would let operations see invalidation churn at a glance.

### Suggested fix

```ts
// services/epic-connector/src/token-client.ts, in fetchToken()
if (this.generation === startedAtGen) {
  this.cached = {
    response: parsed,
    expiresAtMs: issuedAtMs + parsed.expires_in * 1000,
  };
} else {
  log.debug("Dropping stale token-fetch result; cache invalidated mid-flight", {
    startedAtGen,
    currentGen: this.generation,
  });
}
```

Symmetric log on the in-flight slot finally would also help, though that path is already implicitly observable via `inFlight === pending` mismatch which is structurally a no-op.

## Acceptance Criteria

- [ ] `fetchToken` emits a `debug` (or `warn` if loud-by-default is preferred) log line when it suppresses a cache write due to generation mismatch.
- [ ] Log includes `startedAtGen` and `currentGen` (or `currentGeneration`).
- [ ] Existing token-client tests still pass.
- [ ] Add a test asserting the log fires on the documented race (optional but nice).

blamechris:
## Summary

Adds a debug log line when `fetchToken` skips a cache write because `invalidate()` advanced the generation mid-fetch. Includes `startedAtGen` + `currentGen` in the structured payload so operators can see invalidate churn during 401-driven outages.

## Test plan

- [ ] New test asserts the debug log fires on the stale-write path
- [ ] Existing #1143 race tests still pass

Closes #1161

### DIFF
--- a/services/epic-connector/src/__tests__/token-client.test.ts
+++ b/services/epic-connector/src/__tests__/token-client.test.ts
@@ -488,4 +488,66 @@ describe("EpicTokenClient (#389)", () => {
     expect(fetchMock).toHaveBeenCalledTimes(2);
     expect(recovered).toBe("after-retry");
   });
+
+  it("emits a debug log when a stale-generation fetch is dropped (#1161)", async () => {
+    // After #1143 introduced the generation guard, a stale fetch that
+    // resolves AFTER invalidate() silently refuses to write back to the
+    // cache. Operators need observability into how often that happens —
+    // during a 401-driven invalidate storm the silent skip looked like a
+    // black hole. Assert that the skip path emits a structured debug log
+    // carrying both the captured and current generation counters so the
+    // churn is visible in log aggregators.
+    const logSpy = vi.spyOn(console, "log").mockImplementation(() => {});
+
+    let resolveFirst: (response: Response) => void;
+    let resolveSecond: (response: Response) => void;
+    const firstPending = new Promise<Response>((resolve) => {
+      resolveFirst = resolve;
+    });
+    const secondPending = new Promise<Response>((resolve) => {
+      resolveSecond = resolve;
+    });
+    const fetchMock = vi
+      .fn()
+      .mockImplementationOnce(() => firstPending)
+      .mockImplementationOnce(() => secondPending);
+    const client = new EpicTokenClient(makeConfig(privatePem), fetchMock);
+
+    // P1 in-flight -> invalidate() bumps generation -> P2 in-flight ->
+    // resolve P2 first (fresh) then P1 last (stale, must be dropped +
+    // logged).
+    const firstCall = client.getAccessToken();
+    client.invalidate();
+    const secondCall = client.getAccessToken();
+
+    resolveSecond!(tokenResponse({ access_token: "fresh-token" }));
+    await secondCall;
+    resolveFirst!(tokenResponse({ access_token: "stale-token" }));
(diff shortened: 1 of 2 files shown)


blamechris (commented):
## CareBridge Inspector review — PR #1169

**Verdict: APPROVE**

Clean, minimal observability patch. Hits every point from issue #1161's acceptance criteria.

### What was checked

- **Placement** — `log.debug` lives in the `else` after `if (this.generation === startedAtGen)` in `fetchToken`. Correct branch — only fires on the documented stale-write skip path (#1143). No spurious emissions on the happy path.
- **Payload** — Carries `startedAtGen` + `currentGen` exactly as the issue requested. Two integers; no token contents, no scopes, no JWT, no PHI-adjacent material.
- **Level** — `debug` (not `warn`). Matches the issue's stated preference and avoids drowning aggregators during steady state. Operators can still alert on rate-of-debug-occurrence if they want.
- **In-flight symmetry** — Issue itself notes the in-flight slot's stale-skip is "structurally a no-op" (the `if (this.generation === myGen && this.inFlight.get(key) === pending)` finally block in `getAccessToken`). Asymmetric coverage is intentional and the issue explicitly accepts it.
- **Test** — Race ordering is correct: P1 starts at gen=0 → `invalidate()` bumps to gen=1 → P2 starts at gen=1 → resolveSecond (fresh, gen-match writes cache) → resolveFirst (stale, gen-mismatch triggers log). Spies `console.log` which is the correct channel for `debug` (logger sends `debug`/`info` to stdout, `warn`/`error` to stderr — verified in `packages/logger/src/index.ts`). Filters by `level==="debug"` and msg substring "stale", asserts both generation counters, then restores the spy.
- **#1143/#1144 race tests** — Diff only appends one test; no existing tests modified. All 253 epic-connector tests pass locally.
- **API stability for stacked #1172** — `fetchToken` signature unchanged, `normalizeScopes` untouched. #1172 (normalizeScopes dedupe) can rebase cleanly.
- **Zero Attribution Policy** — Commit message, diff content, and PR body are all clean. No Co-Authored-By, no "Generated with", no Claude/Anthropic mention.

### Local verification

- `pnpm typecheck` — FULL TURBO, 38/38 cached pass
- `pnpm lint` — FULL TURBO, 21/21 cached pass
- `pnpm --filter @carebridge/epic-connector test` — 253/253 pass, including the new #1161 test

### Findings

None blocking. One nit-level observation that's not worth a from-review issue: the test asserts the literal generation values (`0` and `1`) tied to the initial generation seed. If a future change ever seeds the counter differently, the assertion will need updating — but the values currently correctly express "exactly one invalidate fired mid-fetch" which is the semantic under test.

Ready for merge once child #1172 is rebased.

PR #1170, opened 2026-05-24
docs(ops): migration patterns for populated-column CHECK additions (#1166)

### LINKED ISSUE #1166: docs/ops: NOT VALID + VALIDATE CONSTRAINT pattern for post-backfill CHECK additions
## Context

Identified during review of PR #1157 ([Copilot comment](https://github.com/blamechris/carebridge/pull/1157#discussion_r3294033979)).

PR #1157 added DB CHECK constraints on `users.npi` and `users.nucc_code`. That migration is safe because the columns were just introduced in #1141 with no production writes — no rows to fail validation. Future schema tightening on a column that already has live data is a different story: a plain `ADD CONSTRAINT` will validate every existing row and block deploy on any violation.

## What to add

A short ops playbook section (likely under `docs/database-operations.md` or a new `docs/migration-patterns.md`) covering the two safe paths for adding a CHECK to a populated column:

1. **Pre-clean then add** — sweep / nullify offending rows in the same migration, then ADD CONSTRAINT.
2. **NOT VALID + VALIDATE** — add the constraint as `NOT VALID` (skips existing-row check, but enforces on future writes) → run a separate cleanup → `VALIDATE CONSTRAINT` once clean.

Should include a worked example and a checklist item for the PR template (`- [ ] If this CHECK targets a populated column, used NOT VALID or pre-cleanup`).

## Acceptance criteria

- [ ] Doc page added or section appended.
- [ ] Linked from `packages/db-schema/README.md` (if it exists) and the PR template checklist.
- [ ] Reference both 0053 (safe ADD CONSTRAINT — no live data) and 0026 / 0038 (DROP IF EXISTS idempotency).

blamechris:
## Summary

Adds `docs/migration-patterns.md` covering the two safe patterns for adding CHECK / FK / NOT NULL constraints to populated columns: pre-clean + add (Pattern A) and NOT VALID + VALIDATE (Pattern B). Worked example references PR #1157's safe path (freshly-introduced columns, no production data). Updates the PR template with a checklist item.

Closes #1166

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,26 @@
+<!--
+PR template. Delete sections that don't apply rather than leaving them empty.
+-->
+
+## Summary
+
+<!-- 1-3 sentences: what changed and why. -->
+
+## Linked issues
+
+<!-- e.g. "Closes #1234" — use "Closes" (not "Refs") so GitHub auto-closes the issue on merge. -->
+
+## Test plan
+
+- [ ]
+- [ ]
+
+## Database migrations
+
+- [ ] N/A — this PR adds no migrations
+- [ ] If this migration adds a CHECK / FK / NOT NULL on a populated column, used Pattern B (NOT VALID + VALIDATE) or Pattern A (pre-clean + add). See docs/migration-patterns.md.
+
+## HIPAA / PHI
+
+- [ ] No PHI in logs, error messages, or test fixtures
+- [ ] Audit log writes added for any new PHI read/write paths
--- a/docs/migration-patterns.md
+++ b/docs/migration-patterns.md
@@ -1,0 +1,224 @@
+# Migration Patterns: Adding CHECK / FK / NOT NULL to Populated Columns
+
+Drizzle emits the SQL exactly as written. Postgres validates new
+constraints against existing rows by default. That combination is what
+makes a "small" migration block a deploy on a populated table.
+
+This doc covers the two safe patterns for adding a CHECK constraint
+(or a FOREIGN KEY, or a NOT NULL) on a column that already has live
+production rows. The choice between them is driven by how confident
+you are that the existing data passes the constraint.
+
+## Why this matters
+
+A plain `ALTER TABLE ... ADD CONSTRAINT ... CHECK (...)` does two
+things in one statement:
+
+1. Acquires an `ACCESS EXCLUSIVE` lock on the table.
+2. Scans every existing row and evaluates the predicate. If any row
+   fails, the statement aborts and the migration rolls back.
+
+Concretely:
+
+- **On any table with violating rows**, deploy fails. The migration
+  runner exits non-zero, the release pipeline halts, and you find out
+  about the bad data the hard way — at deploy time, with the table
+  half-locked behind a transaction the runner is rolling back.
+- **On a large table** (think `audit_log` after a year of writes, or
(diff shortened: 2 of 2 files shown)


blamechris (commented):
## CareBridge Inspector — agent-review

### Verdict: APPROVE

Doc-only PR (250 additions, 2 files: `docs/migration-patterns.md` + `.github/pull_request_template.md`). Zero code touched, CI green across Build/Test/Typecheck.

### Postgres semantics — verified against official PG 16 docs

- `ADD CONSTRAINT CHECK ... NOT VALID` → brief `ACCESS EXCLUSIVE` (catalog update only), no validation scan. **Confirmed.**
- `VALIDATE CONSTRAINT` → `SHARE UPDATE EXCLUSIVE`, does not block concurrent reads/writes. **Confirmed.** ("The validation step does not need to lock out concurrent updates" — PG docs)
- NOT VALID applies to CHECK and FK; PG 16 does **not** support `NOT VALID` on a column-level `SET NOT NULL` (that landed in PG 18). The doc's workaround — `CHECK (x IS NOT NULL) NOT VALID` → `VALIDATE` → `SET NOT NULL` — is exactly right for PG 16. **Confirmed.**
- `SET NOT NULL` skips its scan when a validated `IS NOT NULL` CHECK already proves no-NULLs. **Confirmed** in PG docs verbatim.

CareBridge runs `postgres:16-alpine` (docker-compose.yml), so the version constraint matters and the doc gets it right.

### Worked example accuracy

- PR #1157 → migration `0053_npi_nucc_check.sql` exists and matches the doc's quote.
- The "safe only because" framing is correct: PR #1141 introduced the columns nullable, no production write path stamps a non-NULL value, so the validating scan had no work.
- DROP CONSTRAINT IF EXISTS idempotency pattern is actually present in migrations 0026 and 0038 as cited.

### audit_log immutability wrinkle

The doc claims Pattern A's pre-clean UPDATE can't run on `audit_log` because the immutability triggers reject UPDATE. Verified against `0012_audit_log_immutability.sql`: the `audit_log_no_update` row trigger raises an exception on any UPDATE. The framing in the doc ("the only viable path is Pattern B with violators left as a known historical artefact… cannot be VALIDATEd until every violating row ages out under the 7-year retention window") is accurate and threads the needle correctly — Pattern B's `NOT VALID` still enforces correctness on new writes without touching history.

### Style / hygiene

- Markdown structure matches `docs/hipaa-retention.md` (same heading depth, fenced code blocks, table conventions).
- Zero Attribution Policy upheld in doc, template, and commit.
- PR body uses `Closes #1166` per the template's own guidance — good dogfooding.

### Nits (deferring to from-review issues, not blocking)

1. Pattern B's lock claim is stated only for CHECK. **FOREIGN KEY NOT VALID is even better** — it takes `SHARE ROW EXCLUSIVE` instead of `ACCESS EXCLUSIVE`, so concurrent reads/writes continue during the add. Worth a one-line callout the next time this doc is touched.
2. The `audit_log` paragraph implies historical violators "age out" under retention, but the in-flight archival job described in `hipaa-retention.md` is "future work, not yet implemented." Either fine to leave (the doc is written for the eventual state) or worth a forward-pointer.
3. Could cross-link from a future `packages/db-schema/README.md` per the issue's acceptance criteria — the README doesn't exist today, so this is correctly deferred.

None of these block merge.

PR #1177, opened 2026-05-24
docs(ops): FK NOT VALID lock semantics (#1173)

### LINKED ISSUE #1173: docs/ops: document FK NOT VALID lock difference (SHARE ROW EXCLUSIVE) in migration-patterns.md
## Context

Identified during review of PR #1170 ([migration-patterns.md](https://github.com/blamechris/carebridge/blob/main/docs/migration-patterns.md)).

`docs/migration-patterns.md` covers `NOT VALID` for CHECK constraints but only mentions FK in passing in the section header. FK constraints behave **better** than CHECK in the `NOT VALID` flavour:

- `ALTER TABLE ... ADD CONSTRAINT ... FOREIGN KEY (...) REFERENCES ... NOT VALID` takes only `SHARE ROW EXCLUSIVE` instead of `ACCESS EXCLUSIVE` (per PG docs).
- That means concurrent SELECTs and even row-level INSERT/UPDATE continue during the add, where CHECK still blocks them.
- `VALIDATE CONSTRAINT` semantics are identical to CHECK (`SHARE UPDATE EXCLUSIVE`).

## What to add

A short subsection under Pattern B (or a sibling "FK-specific" note) calling out:
1. The reduced lock level for FK `NOT VALID`.
2. The two-step rollout still applies (NOT VALID → backfill orphans → VALIDATE).
3. Worked example would need a migration in the repo that adds an FK to a populated column — none today, so could be left as conceptual until a real example exists.

## Acceptance criteria

- [ ] FK-specific lock claim (`SHARE ROW EXCLUSIVE`) added to `docs/migration-patterns.md`.
- [ ] PR template item updated if the wording becomes incorrect.

blamechris:
## Summary

Adds a subsection to `docs/migration-patterns.md` documenting that FK `NOT VALID` takes only `SHARE ROW EXCLUSIVE` instead of CHECK's `ACCESS EXCLUSIVE`, so concurrent SELECT + row-level INSERT/UPDATE continue during the add. Two-step rollout (NOT VALID → backfill orphans → VALIDATE) still applies.

Closes #1173

### DIFF
--- a/docs/migration-patterns.md
+++ b/docs/migration-patterns.md
@@ -165,20 +165,33 @@ indistinguishable from one added the plain way.
 > lock on the referenced table, in addition to the lock on the table
 > on which the constraint is declared.
 
-In practice this means:
+In practice this means (per the PG 16 [lock conflict
+matrix](https://www.postgresql.org/docs/16/explicit-locking.html),
+Table 13.2):
 
 - `ALTER TABLE child ADD CONSTRAINT ... FOREIGN KEY (parent_id) REFERENCES parent (id) NOT VALID`
-  blocks other writers (`SHARE ROW EXCLUSIVE` conflicts with itself
-  and with `ROW EXCLUSIVE`-holding statements like concurrent DDL
-  and `VACUUM FULL`) **but does not block plain reads or row-level
-  INSERT/UPDATE/DELETE**, which use weaker `ROW SHARE` /
-  `ROW EXCLUSIVE` locks. A CHECK constraint added the same way
-  would freeze the table.
+  takes `SHARE ROW EXCLUSIVE` on both the child and the referenced
+  parent. `SHARE ROW EXCLUSIVE` conflicts with `ROW EXCLUSIVE`,
+  `SHARE UPDATE EXCLUSIVE`, `SHARE`, `SHARE ROW EXCLUSIVE`,
+  `EXCLUSIVE`, and `ACCESS EXCLUSIVE`. In production terms: plain
+  `SELECT` (`ACCESS SHARE`) and `SELECT FOR UPDATE/SHARE`
+  (`ROW SHARE`) continue, but plain `INSERT`/`UPDATE`/`DELETE`
+  (which take `ROW EXCLUSIVE`) are blocked on both tables for the
+  duration of the catalog write. That window is normally
+  milliseconds because `NOT VALID` skips the row scan; the win
+  versus CHECK is that reads keep flowing and the writer pause is
+  near-instant instead of "however long the full-table scan takes".
+- Practical risk: the `SHARE ROW EXCLUSIVE` on the **referenced
+  parent** matters more than the lock on the child. Adding an FK to
+  a small child table can still briefly pause writes on a
+  high-traffic parent (e.g. `users`, `patients`). Plan the add for
+  a low-write window on the parent, not just on the child.
 - The two-step rollout is still required when the column is
   populated: add `NOT VALID`, backfill any orphan rows that don't
   resolve to a parent, then `VALIDATE CONSTRAINT`. `VALIDATE` takes
-  the same `SHARE UPDATE EXCLUSIVE` as for CHECK (plus a `ROW SHARE`
-  on the referenced table), so it does not block reads or writes.
+  `SHARE UPDATE EXCLUSIVE` on the child plus `ROW SHARE` on the
+  referenced parent, so it does not block reads or row-level
+  writes.
 - New INSERTs and UPDATEs are FK-checked from the moment the
   `NOT VALID` constraint lands — only pre-existing rows are
   unvalidated.
--- a/docs/migration-patterns.md
+++ b/docs/migration-patterns.md
@@ -153,6 +153,41 @@ indistinguishable from one added the plain way.
   the `VALIDATE` migration as a follow-up issue at the time you
   merge the `NOT VALID` one.
 
+### FK-specific note: reduced lock level
+
+`ADD CONSTRAINT ... FOREIGN KEY` is the one exception to the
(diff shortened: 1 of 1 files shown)


blamechris:
## CareBridge Inspector — PR #1177 review

### Verdict: REQUEST CHANGES (1 Critical correctness bug in the docs)

The PR scope is documentation only (one new subsection in `docs/migration-patterns.md`). I verified the lock-mode claims against the PG 16 docs (`sql-altertable.html` + `explicit-locking.html`) and confirmed against Table 13.2 in the conflict matrix.

### Critical (1) — incorrect lock-conflict claim

> blocks other writers (`SHARE ROW EXCLUSIVE` conflicts with itself and with `ROW EXCLUSIVE`-holding statements like concurrent DDL and `VACUUM FULL`) **but does not block plain reads or row-level INSERT/UPDATE/DELETE**, which use weaker `ROW SHARE` / `ROW EXCLUSIVE` locks.

This sentence is self-contradictory and the final clause is wrong per PG 16:

- Plain `INSERT`/`UPDATE`/`DELETE` take `ROW EXCLUSIVE` (not `ROW SHARE`). `SELECT FOR UPDATE`/`FOR SHARE` is what takes `ROW SHARE`. Plain `SELECT` takes `ACCESS SHARE`.
- The PG 16 conflict matrix (Table 13.2) and the prose description say:
  > `SHARE ROW EXCLUSIVE` ... Conflicts with the `ROW EXCLUSIVE`, `SHARE UPDATE EXCLUSIVE`, `SHARE`, `SHARE ROW EXCLUSIVE`, `EXCLUSIVE`, and `ACCESS EXCLUSIVE` lock modes.
- Therefore `ADD FOREIGN KEY ... NOT VALID` **does** block plain `INSERT`/`UPDATE`/`DELETE` on both child and referenced parent for the duration of the catalog write. It allows plain `SELECT` (and `SELECT FOR UPDATE`/`FOR SHARE`) to continue.

What's true: the catalog-only `NOT VALID` add is fast (no row scan), so the write-blocking window is normally milliseconds — which is what makes it operationally safe, not the absence of conflict with row writes.

Also incorrect on a related claim: "`ROW EXCLUSIVE`-holding statements like concurrent DDL and `VACUUM FULL`". `VACUUM FULL` takes `ACCESS EXCLUSIVE`, not `ROW EXCLUSIVE`. The canonical `ROW EXCLUSIVE` holders are plain `INSERT`/`UPDATE`/`DELETE`/`MERGE`.

#### Suggested rewrite

```markdown
- `ALTER TABLE child ADD CONSTRAINT ... FOREIGN KEY (parent_id) REFERENCES parent (id) NOT VALID`
  takes `SHARE ROW EXCLUSIVE` on both the child and the referenced
  parent table. Per the PG 16 lock conflict matrix this allows
  concurrent `SELECT` (`ACCESS SHARE`) and `SELECT FOR UPDATE/SHARE`
  (`ROW SHARE`) to proceed, but **blocks plain
  `INSERT`/`UPDATE`/`DELETE`** (which take `ROW EXCLUSIVE`) and
  conflicts with itself, `SHARE`, `SHARE UPDATE EXCLUSIVE`,
  `EXCLUSIVE`, and `ACCESS EXCLUSIVE`. A CHECK constraint added the
  same way would also block reads (`ACCESS EXCLUSIVE`). The win for
  FK is "reads continue" plus a much smaller blast radius for the
  catalog-only `NOT VALID` add, which returns in milliseconds.
```

### Suggestions (1)

- Consider noting that the `SHARE ROW EXCLUSIVE` on the *referenced parent* is the load-bearing operational risk on a high-traffic parent table (e.g. `users`, `patients`): adding an FK to a tiny child table can still briefly pause writes on a hot parent. Worth a sentence so the next reader doesn't think child-table-only is the locking story.

### Nits (1)

- Markdown style: the rest of the doc uses fenced code blocks for SQL examples. The inline `ALTER TABLE child ADD CONSTRAINT ...` line is long enough that it wraps in source rendering; consider a fenced block to match Pattern A/B style. Non-blocking.

### Verified OK

- PG version cited correctly (PG 16, matches `postgres:16-alpine` in `docker-compose.yml`).
- Two-step rollout (NOT VALID → backfill orphans → VALIDATE) is correctly noted as still required.
- `VALIDATE CONSTRAINT` lock semantics for FK (`SHARE UPDATE EXCLUSIVE` on child + `ROW SHARE` on parent) match the PG 16 docs verbatim.
- PR template (`.github/pull_request_template.md`) already reads "CHECK / FK / NOT NULL" and references `docs/migration-patterns.md` — no template change is needed, agent's claim is correct.
- Zero Attribution Policy: no Claude/Anthropic/AI mentions in the patch.
- CI: Build / Test / Typecheck all green.

blamechris:
Fix pushed in aa8f719:
- Replaced the self-contradictory `SHARE ROW EXCLUSIVE` bullet with the verbatim PG 16 conflict list (`ROW EXCLUSIVE`, `SHARE UPDATE EXCLUSIVE`, `SHARE`, `SHARE ROW EXCLUSIVE`, `EXCLUSIVE`, `ACCESS EXCLUSIVE`) and called out that plain `INSERT`/`UPDATE`/`DELETE` are briefly blocked (the win vs CHECK is reads-keep-flowing + millisecond catalog write, not freedom-from-write-block).
- Dropped the incorrect `VACUUM FULL` / "concurrent DDL" examples for `ROW EXCLUSIVE` holders.
- Added the practical-risk note about the lock on the referenced parent being the load-bearing concern.
- Fixed the `VALIDATE` bullet to specify `SHARE UPDATE EXCLUSIVE` is on the child + `ROW SHARE` on the parent.

CI: Build / Test / Typecheck all green on the new commit.

PR #1192, opened 2026-05-25
fix(portal): epic-launch-banner spoofing guard (#1186)

### LINKED ISSUE #1186: fix(portal): epic-launch-banner trusts URL params even when server context is null (mild spoof vector)
## Context

\`apps/clinician-portal/src/components/epic-launch-banner.tsx\` activates when the URL has \`?epic_patient=<fhir-id>\` and renders \"Launched from <org> — Viewing Epic patient FHIR id <epic_patient>\".

It queries \`epicAuth.getLaunchContext\` for server-confirmed context, but explicitly chooses to render even when the server returns \`null\`:

\`\`\`tsx
// Optional: when the URL says we launched but the server says no valid
// connection, still show the banner using URL data (the connection
// may have just expired between the redirect and now — clinically it's
// safer to surface the patient context than to silently drop it).
const ctx = contextQuery.data ?? null;
const orgLabel = ctx?.epic_org_iss ?? \"Epic\";
\`\`\`

The reasoning is clinical-safety driven (don't silently drop launch context). But this means an attacker who can get a clinician to click

\`\`\`
https://carebridge.example/patients/abc?epic_patient=spoofed-fhir-id&epic_encounter=fake
\`\`\`

…will see \"Launched from Epic — Viewing Epic patient FHIR id spoofed-fhir-id\". No PHI leak (the patient page still uses the \`[id]\` route param), but it could be combined with social-engineering (\"your launch context expired, please re-authorise\").

## Required

Choose ONE of:

1. **Strict mode**: only render the banner when \`epicPatient === ctx?.epic_patient_fhir_id\`. Render nothing when ctx is null. (Trades clinical-safety for security.)
2. **Tagged mode**: render the banner but add an \"(unverified)\" or warning sub-line when ctx is null, and don't echo the FHIR id verbatim — replace with a generic placeholder.

## Acceptance

- [ ] Manually crafted URL with \`epic_patient\` but no live Epic connection renders either nothing or a clearly-unverified state.
- [ ] Test: launch banner with URL-only param and \`getLaunchContext\` returning null asserts the chosen behavior.

## References

- PR #1184

blamechris:
## Summary

The banner no longer renders Epic context from URL params alone — it requires `epicAuth.getLaunchContext` to return a non-null server-confirmed context. If the server says no context and the URL says yes, the banner stays hidden (and emits a console.warn for mismatch).

Preserves the clinical-safety intent for genuine launches: when server context is present, the banner uses server values exclusively.

## Test plan

- [ ] Server null + URL set → renders null
- [ ] Server set + URL absent → banner renders (genuine post-callback)
- [ ] Server set + URL mismatch → renders null + warn
- [ ] Server set + URL match → banner renders, uses server values

Closes #1186

### DIFF
--- a/apps/clinician-portal/src/__tests__/epic-launch-banner.test.tsx
+++ b/apps/clinician-portal/src/__tests__/epic-launch-banner.test.tsx
@@ -85,7 +85,12 @@ afterEach(() => {
 });
 
 describe("EpicLaunchBanner", () => {
-  it("returns null when no epic_patient param is on the URL", () => {
+  it("returns null when no URL params AND server has no launch context", () => {
+    getLaunchContextQuery.mockReturnValueOnce({
+      data: null,
+      isLoading: false,
+      isError: false,
+    });
     const { storage } = makeStorage();
     render(<EpicLaunchBanner storage={storage} />);
     expect(screen.queryByTestId("epic-launch-banner")).toBeNull();
@@ -95,6 +100,17 @@ describe("EpicLaunchBanner", () => {
     searchParamsMock = new URLSearchParams({
       epic_patient: "eP123",
       epic_encounter: "eE456",
+    });
+    getLaunchContextQuery.mockReturnValueOnce({
+      data: {
+        epic_org_iss: "https://fhir.epic.example/api/FHIR/R4",
+        epic_practitioner_fhir_id: "prac-1",
+        epic_patient_fhir_id: "eP123",
+        launch_encounter_fhir_id: "eE456",
+        expires_at: "2099-01-01T00:00:00Z",
+      },
+      isLoading: false,
+      isError: false,
     });
     const { storage } = makeStorage();
     render(<EpicLaunchBanner storage={storage} />);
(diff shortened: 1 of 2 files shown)


blamechris (commented):
## CareBridge Inspector — Security Review

**Verdict: APPROVE**

### Trust model — VERIFIED
- URL params (`urlEpicPatient`, `urlEpicEncounter`) are pulled at lines 54–55 but used **only** for equality comparison (lines 96, 103–106). They are never rendered.
- The rendered values (`orgLabel`, `patientId`, `encounterId` at lines 116–118) come exclusively from the server payload `ctx`.
- Pre-`ctx` early returns (loading/error/null-ctx at lines 85–89) ensure the banner cannot escape the gate when the server has no context — even with a fully-spoofed URL.

### Suppress-key safety — VERIFIED
- Read at line 75 compares storage value against `serverPatient` (server-confirmed FHIR id), not URL.
- Write at line 123 uses `patientId` (server-derived). A spoofed URL can never produce a write here because the function is only reachable past the `!ctx` gate, where `patientId === ctx.epic_patient_fhir_id`.
- Lifecycle: dismiss(server-FHIR-X) → close-tab clears flag on logout/new tab → re-launch(server-FHIR-X) → banner re-shows only if server context says X.
- Cross-spoof attempt (URL-only without server context): the effect early-returns on `!serverPatient` (line 72), so a spoofed URL can't even read the key.

### Mismatch warning — REASONABLE
- Fires only when `urlEpicPatient && urlEpicPatient !== ctx.epic_patient_fhir_id`. Same for encounter.
- Filed nit below for rate-limiting (React strict-mode double-render + frequent re-renders could fire the warn multiple times for the same tamper signal).

### Edge cases — all six handled correctly
1. Server null + no URL → null  (lines 67, 89)
2. Server ctx + no URL → renders from server  (lines 96 short-circuits when `urlEpicPatient` is falsy)
3. Server null + URL has epic_patient → null (anti-spoof gate at line 89; no warn since `!ctx` exits before the mismatch check — that's fine since there's no comparison to make)
4. Server ctx + URL mismatch → null + warn  (lines 96–101)
5. Server ctx + URL match → renders, uses server values
6. Server loading → null  (line 85)

### Stacked child #1198 compatibility — VERIFIED
All hooks (`useRouter`, `usePathname`, `useSearchParams`, `useQuery`, `useState`, `useEffect`) are unconditionally called at the top of the component before any early return. Replacing `useState(false)` + post-mount `useEffect` with a lazy `useState(() => readSync())` in #1198 is fully compatible.

### Tests — 8 passing locally
- Test "stays hidden when sessionStorage flag matches current patient" relies on the default mock returning `epic_patient_fhir_id: "eP123"` matching the storage flag `"eP123"`. Now reads as: server-confirmed eP123 + suppress flag eP123 → hidden. Still correct under the new semantics.
- The 4 new tests cover the four key branches enumerated in the PR's test plan.

### Zero attribution — VERIFIED
Component header, PR body, and commit messages contain no Claude/Anthropic attribution.

### One non-blocking nit (will file from-review)
`console.warn` at lines 98 and 109 fires on every render where the mismatch holds. In React 18 strict mode dev builds, that's already 2x per "render." If the user navigates with sticky URL params, this could fire many times for the same tamper signal. Consider memoizing/guarding with a `useRef` "already warned for this mismatch" or use `useEffect` so it fires once per (urlPatient, serverPatient) pair. Non-blocking for security correctness.

LGTM — fix correctly converts URL params from "render source" to "comparator-only", and the suppress key is now bound to server-confirmed identity.

PR #1193, opened 2026-05-25
fix(gateway): /auth/epic/authorize open-redirect (#1185)

### LINKED ISSUE #1185: fix(gateway): validate `redirect` query param on /auth/epic/authorize to prevent open redirect
## Context

The Epic OAuth standalone-launch flow in `services/api-gateway/src/routes/epic-auth.ts` accepts a `redirect` query param on `GET /auth/epic/authorize`:

```ts
postLaunchRedirect: request.query.redirect ?? "/",
```

That value is stored in launch state and later fed verbatim into `reply.redirect(finalUrl)` from the callback handler. No allowlist or scheme/path check is applied.

The clinician-portal UI from #1184 hardcodes `redirect=/settings/integrations` so users of the UI are safe today. The risk lives at the route level: a logged-in clinician clicking a malicious link like

```
https://carebridge.example/auth/epic/authorize?iss=https://fhir.epic.example/api/FHIR/R4&redirect=https://attacker.example/phish
```

…would, after completing the OAuth dance, be redirected to the attacker-controlled URL — a classic open redirect, useful for phishing follow-ons.

## Required

- Validate `redirect` to be a same-origin path (`startsWith("/")` and no `//`/`\\`/protocol prefix), OR
- Maintain an allowlist of permitted redirect paths.

## Acceptance

- [ ] `?redirect=https://attacker.example` returns 400 from `/auth/epic/authorize`.
- [ ] `?redirect=/settings/integrations` and other same-origin paths continue to work.
- [ ] Unit test for the validator.

## References

- PR #1184 (the UI surface)
- Issue #392 (parent SMART App Launch)

blamechris:
## Summary

Validates the optional `redirect` query parameter on `GET /auth/epic/authorize` to be a same-origin path. Rejects `https://*`, `//`, `javascript:`, `data:`, and other absolute or protocol-relative forms with 400.

UI today hardcodes `redirect=/settings/integrations` so users hitting the OAuth flow via the portal are unaffected. This closes the route-level open-redirect surface for hand-crafted phishing URLs.

## Test plan

- [ ] Same-origin paths preserved through the flow
- [ ] Absolute / protocol-relative / scheme-injection rejected with 400
- [ ] Missing redirect defaults to "/"

Closes #1185

### DIFF
--- a/services/api-gateway/src/__tests__/epic-auth-redirect-validation.test.ts
+++ b/services/api-gateway/src/__tests__/epic-auth-redirect-validation.test.ts
@@ -244,4 +244,33 @@ describe("GET /auth/epic/authorize — redirect validation (#1185)", () => {
     expect(beginLaunchMock).toHaveBeenCalledTimes(1);
     expect(capturedBeginLaunchArgs[0]?.postLaunchRedirect).toBe("/");
   });
+
+  // Exercises the explicit `/\` prefix guard. Some URL parsers normalise
+  // `/\foo` into the same-origin path `/\foo`, others promote the
+  // backslash to a path separator. We reject up-front so neither shape
+  // can slip through if the parser ever changes behaviour.
+  it("rejects mixed-slash ?redirect=/\\attacker.example with 400", async () => {
+    const response = await server.inject({
+      method: "GET",
+      url: `/auth/epic/authorize?iss=${encodeURIComponent(ISS)}&redirect=${encodeURIComponent("/\\attacker.example")}`,
+      headers: { "x-test-auth": "1" },
+    });
+
+    expect(response.statusCode).toBe(400);
+    expect(beginLaunchMock).not.toHaveBeenCalled();
+  });
+
+  // Whitespace-prefixed candidate — fails the `startsWith("/")` guard.
+  // Important because raw spaces survive `decodeURIComponent` and could
+  // hide downstream parser quirks if we accepted them.
+  it("rejects whitespace-prefixed ?redirect=%20/legit with 400", async () => {
+    const response = await server.inject({
+      method: "GET",
+      url: `/auth/epic/authorize?iss=${encodeURIComponent(ISS)}&redirect=${encodeURIComponent(" /legit")}`,
+      headers: { "x-test-auth": "1" },
+    });
+
+    expect(response.statusCode).toBe(400);
+    expect(beginLaunchMock).not.toHaveBeenCalled();
+  });
 });
--- a/services/api-gateway/src/__tests__/epic-auth-redirect-validation.test.ts
+++ b/services/api-gateway/src/__tests__/epic-auth-redirect-validation.test.ts
@@ -1,0 +1,247 @@
+/**
+ * Open-redirect hardening on GET /auth/epic/authorize (issue #1185).
+ *
+ * The handler previously stored `request.query.redirect` verbatim in launch
+ * state; the callback later did `reply.redirect(finalUrl)` on whatever came
+ * back. Anyone crafting the URL by hand could land a logged-in clinician on
+ * an attacker-controlled origin after the OAuth dance.
+ *
+ * We now validate the redirect on entry: it must parse as a same-origin
+ * path (starts with `/`, no scheme, no protocol-relative `//`, no Windows
+ * `\\` trick). Reject with 400 otherwise.
+ *
+ * These tests boot a minimal Fastify server with a stub user preHandler
+ * and stub the epic-connector orchestrator so beginLaunch doesn't try to
+ * fetch SMART configuration over the network.
+ */
+import {
+  describe,
+  it,
+  expect,
+  vi,
+  beforeAll,
+  afterAll,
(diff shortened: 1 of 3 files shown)


blamechris:
## CareBridge Inspector — security review (PR #1193, issue #1185)

**Verdict: APPROVE.** The fix correctly closes the open-redirect surface on `/auth/epic/authorize` with a defence-in-depth posture and adequate test coverage. No blocking findings.

---

### 1. Validator soundness — `validateRedirect`

I enumerated every shape I could think of and ran each through the actual function (Node 22 URL parser) plus the downstream callback URL re-parse. Behaviour matches the contract documented in the function header:

| Input | Result | Notes |
|---|---|---|
| `undefined` / `""` | accepts → `"/"` | covered by tests #8, #9 |
| `https://attacker.example` | reject (`origin` mismatch) | test #2 |
| `//attacker.example` | reject (prefix guard) | test #3 |
| `\\attacker.example` | reject (no leading `/`) | test #5 |
| `/\attacker.example` | reject (`/\` prefix guard) | not directly tested |
| `javascript:alert(1)` | reject (no leading `/`) | test #4 |
| `data:text/html,…` | reject (no leading `/`) | test #6 |
| `/legitimate/path` | accept | test #1 |
| `/../../etc/passwd` | accept (out of scope) | test #7, comment explains why |
| `/path?q=1#frag` | accept | fragment is dropped by the callback's URL re-parse, but that's a UX nit, not a security issue |
| `/legit\r\nLocation: https://evil` | accept here, BUT the callback re-parses via `new URL(...)` which strips control chars from the pathname (`%0A%0D` would be percent-encoded, raw CR/LF is stripped). No header-injection vector. |
| `/\t/x`, `/\n/x` | reject — URL parser treats raw control chars as making it not start-with-`/` after normalisation (confirmed empirically) |
| ` /legit` (leading space) | reject (no leading `/`) | |
| `:evil` | reject (no leading `/`) | |

The combination of (a) explicit `startsWith("/")` and `startsWith("//")` / `startsWith("/\\")` short-circuits, plus (b) the `URL`-origin equality check, is robust. The explicit prefix guards are belt-and-braces against future URL-parser quirks (good defensive style).

### 2. Defence in depth — callback also re-validates

`/auth/epic/callback` reads `postLaunchRedirect` from launch-state (Redis or in-memory) and runs `new URL(result.postLaunchRedirect, "http://_dummy")`. If launch state were ever poisoned, the `target.origin === "http://_dummy"` check forces `reply.redirect()` to use only `target.pathname + target.search`, which is by construction same-origin. **Two independent checks both have to fail to land an attacker URL.** Strong.

### 3. Same-origin vs absolute-URL design choice

EHR App Launch always loops back through the gateway's own `/auth/epic/callback`, so post-launch redirects are always inside the CareBridge origin. Rejecting absolute URLs is correct — there's no legitimate cross-origin destination in this flow. No false positives.

### 4. Test coverage

9 tests, mapped to acceptance criteria from #1185:
- ✓ Same-origin path forwarded to launch state (`/settings/integrations`)
- ✓ `https://attacker.example` → 400
- ✓ `//attacker.example` → 400
- ✓ `javascript:alert(1)` → 400
- ✓ `\\attacker.example` → 400
- ✓ `data:text/html,…` → 400
- ✓ Path-traversal accepted (correctly scoped out)
- ✓ Missing → defaults to `/`
- ✓ Empty string → defaults to `/`

Each test verifies BOTH the status code AND that `beginLaunch` was (not) called — so it's testing the gateway behaviour, not just the validator helper. Solid.

**Suggested additional cases (non-blocking, can be follow-up):** `/\attacker.example` (exercises the `/\` prefix guard branch) and ` /legit` with leading whitespace. I confirmed both reject correctly, but the dedicated `/\` guard branch is currently uncovered by an assertion.

### 5. `vitest.config.ts` aliases

Adding `@carebridge/epic-connector` → `services/epic-connector/src/index.ts` and `@carebridge/logger` → `packages/logger/src/index.ts`. Verified:
- Both source files exist.
- The only other api-gateway test that touches `@carebridge/logger` (`audit-write-failure-visibility.test.ts`) uses `vi.mock("@carebridge/logger", …)`, so the alias is overridden by the mock and changes nothing for that test.
- No other api-gateway test imports `@carebridge/epic-connector`.
- Local run: `pnpm --filter @carebridge/api-gateway test` → 37/37 test files, 539/539 tests pass after dep dist artifacts are built (CI builds them; pre-existing unrelated failures otherwise).

Aliases are scoped correctly and don't break the build-artefact path elsewhere.

### 6. Zero Attribution

Diff and tests checked — no Claude / Anthropic / "Generated with" mentions. Clean.

### 7. Issue scope

Issue #1185 acceptance criteria all met:
- [x] `?redirect=https://attacker.example` → 400 from `/auth/epic/authorize` ✓
- [x] `?redirect=/settings/integrations` continues to work ✓
- [x] Unit test for the validator ✓ (9 of them)

---

**Recommendation: merge.** Optional follow-up to add the two extra edge-case assertions noted above; not blocking.

Issue #1182, opened 2026-05-25
feat(portal): Connect to Epic UI + launch-context banner (clinician portal) to complete #392


blamechris:
## Context

Discovered while auditing #392 (SMART on FHIR App Launch). The **backend** is fully shipped — \`/auth/epic/authorize\` and \`/auth/epic/callback\` routes, full \`services/epic-connector/src/app-launch/\` module (authorize, pkce, token-exchange, state-store, connection-repo, launch-service), PKCE+S256, migration 0049 \`epic_user_connections\`.

What's missing is the **clinician portal UI surface** — there's no "Connect to Epic" button, no launch-context banner, no patient-pre-selection UX on EHR launch. Today the flow is fully driven by curl / direct URL navigation against the backend routes.

### 1. Settings page — "Connect to Epic" button

- Location: \`apps/clinician-portal/src/app/settings/\` (or wherever the current settings live — read the source first)
- Button kicks off \`GET /auth/epic/authorize\` (no \`launch\` param → standalone flow)
- After callback, show connection state (connected to \`<Epic org name>\`, \`last connected at\`, "Disconnect" action that calls a backend revoke endpoint if one exists; otherwise just clears the local row in \`epic_user_connections\`)
- Support multiple Epic orgs per user (already DB-modeled in 0049)

### 2. Launch-context banner

- When the user enters via \`/auth/epic/authorize?iss=<epic-base>&launch=<token>\` (EHR launch) the callback resolves to a CareBridge route with patient context
- Show a banner at the top of the portal: \`Launched from Epic — viewing <Patient Name>\` with a "Switch to portal mode" affordance
- Patient should be pre-selected (URL param or context provider)

### 3. Backend integration

- Wire the tRPC \`epicSync.triggerSync\` mutation into a "Sync now" action on the patient detail page (admin-gated)
- Show \`epicSync.getSyncStatus\` in a status widget (last_synced_at, error count)

## Acceptance criteria

- [ ] Standalone Launch button visible in settings, end-to-end flow works against Epic sandbox
- [ ] EHR Launch banner renders when arriving with a launch context
- [ ] Patient pre-selected from \`patient\` token-exchange field
- [ ] Disconnect / re-connect works idempotently
- [ ] \`epicSync.triggerSync\` button on patient page (admin-only)

## Out of scope (already done)

- Backend routes — shipped (\`services/api-gateway/src/routes/epic-auth.ts\`)
- PKCE + S256 — shipped
- Token exchange — shipped
[...]

PR #1195, opened 2026-05-25
feat(portal): forward epic_patient URL param to EpicSyncCard (#1188)

### LINKED ISSUE #1188: feat(portal): pass epic_patient_fhir_id from launch URL into EpicSyncCard on patient page
## Context

When a clinician arrives on \`/patients/{id}\` via the Epic launch flow, the URL carries \`?epic_patient=<fhir-id>\`. \`EpicSyncCard\` accepts an optional \`epicPatientFhirId\` prop that, when set, lets the \`triggerSync\` mutation skip the stored-mapping lookup. But the patient page currently mounts the card without forwarding the value:

\`\`\`tsx
// apps/clinician-portal/app/patients/[id]/page.tsx (line 253)
<EpicSyncCard patientId={patientId} />
\`\`\`

Result: when a clinician hits \"Sync from Epic now\" right after launching from Epic, the backend has to round-trip to look up the Epic FHIR id from the connection row even though the launch context already carries it.

## Required

Read \`useSearchParams().get(\"epic_patient\")\` in the patient page and forward to the card:

\`\`\`tsx
const epicPatientFhirId = searchParams.get(\"epic_patient\") ?? undefined;
<EpicSyncCard patientId={patientId} epicPatientFhirId={epicPatientFhirId} />
\`\`\`

## Acceptance

- [ ] When the URL has \`?epic_patient=eP1\`, the sync mutation is called with \`epic_patient_fhir_id: \"eP1\"\`.
- [ ] Test extends the existing patient-chart-tabs render or adds a small standalone integration test.

## References

- PR #1184

blamechris:
## Summary

`/patients/[id]` now reads `?epic_patient` from `useSearchParams` and passes it to `EpicSyncCard` as `epicPatientFhirId`. The card already knew how to consume the prop — this just wires the URL→prop path so clinicians arriving via the Epic launch flow get a faster Sync-Now (no round-trip to the stored mapping).

Input is validated to FHIR-id shape (`[A-Za-z0-9.-]{1,64}`) before forwarding.

## Test plan

- [ ] URL param present → forwarded
- [ ] URL param absent → not forwarded
- [ ] URL param malformed → not forwarded

Closes #1188

### DIFF
--- a/apps/clinician-portal/app/patients/[id]/page.tsx
+++ b/apps/clinician-portal/app/patients/[id]/page.tsx
@@ -32,6 +32,7 @@ import {
 } from "@/lib/lab-display";
 import { PrescriptionForm } from "./PrescriptionForm";
 import { EpicSyncCard } from "@/components/epic-sync-card";
+import { parseEpicPatientParam } from "@/lib/epic-launch";
 import type { Medication } from "@carebridge/shared-types";
 
 const tabs = [
@@ -85,7 +86,13 @@ function stalenessStyles(tier: StalenessTier): {
   }
 }
 
-function OverviewTab({ patientId }: { patientId: string }) {
+function OverviewTab({
+  patientId,
+  epicPatientFhirId,
+}: {
+  patientId: string;
+  epicPatientFhirId?: string;
+}) {
   const patientQuery = trpc.patients.getById.useQuery({ id: patientId });
   const diagnosesQuery = trpc.patients.diagnoses.getByPatient.useQuery({ patientId });
   const allergiesQuery = trpc.patients.allergies.getByPatient.useQuery({ patientId });
@@ -249,8 +256,14 @@ function OverviewTab({ patientId }: { patientId: string }) {
         )}
       </div>
 
-      {/* Epic sync status + admin-only Sync Now action (#1182). */}
-      <EpicSyncCard patientId={patientId} />
+      {/* Epic sync status + admin-only Sync Now action (#1182).
+          When the user arrived via the Epic SMART launch flow the URL
+          carries `?epic_patient=<fhir-id>` (#1188) — forward it so the
+          backend can skip the stored-mapping lookup. */}
+      <EpicSyncCard
+        patientId={patientId}
+        epicPatientFhirId={epicPatientFhirId}
+      />
     </div>
   );
 }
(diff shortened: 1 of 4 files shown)


blamechris:
## CareBridge Inspector — Review

**Verdict: APPROVE** — ship it.

### Regex correctness (FHIR R4 §2.36.1.4)
- Pattern `/^[A-Za-z0-9.\-]+$/` is properly anchored at both ends. Mentally + executed: `<script>` → rejected, `eP-1` → accepted, empty → rejected, 65-char → rejected, whitespace/newline/tab/unicode → rejected. All boundary cases (`64` accepted, `65` rejected, empty rejected) covered by tests. Centralising `FHIR_ID_MAX_LENGTH = 64` is the right move.
- Defence-in-depth note in the helper's JSDoc about *not* `decodeURIComponent`-ing the value is correct — `URLSearchParams.get()` already decodes.

### Helper isolation
- `parseEpicPatientParam` is a pure, exported function in `src/lib/epic-launch.ts` — easy to unit-test, easy to reuse if other routes start consuming launch params later. Good shape.

### Forwarding logic
- Validation runs synchronously on each render via `searchParams.get(...)` → `parseEpicPatientParam(...)`. No useEffect lag, so "click Sync now" cannot fire before validation. Verified by reading `PatientChartContent` at lines 961-973 + the `<OverviewTab epicPatientFhirId={...} />` mount at line 1033.

### EpicSyncCard consumption + backend fallback
- `EpicSyncCard` already accepts `epicPatientFhirId` (#1182 baseline) and forwards it into the `triggerSync` mutation payload as `epic_patient_fhir_id` (snake_case) — verified at `src/components/epic-sync-card.tsx:65-69`.
- Backend: `services/api-gateway/src/routers/epic-sync.ts:82-134` accepts `epic_patient_fhir_id: z.string().optional()` for all three modes and passes it through to `enqueueIncrementalSync` / `enqueueFullSync` / `enqueueSingleResourceSync`.
- Fallback path: `services/epic-connector/src/sync/sync-jobs.ts:534-538` — when `epicPatientFhirId` is absent, `fetchResources` falls back to `args.patientId` for non-Patient resources (and short-circuits to `[]` for the Patient resource itself). Missing param is genuinely a no-op-or-stored-mapping path, not a crash.

### Pre-existing TS errors
- Confirmed by running `tsc --noEmit` against `origin/main`: the six `layout-skip-nav.test.tsx` TS2339 errors exist on main and are unchanged by this PR. The PR's claim that they are pre-existing and unrelated is accurate.

### Tests
- 15 new tests across `epic-launch-helper.test.ts` (10 unit) + `patient-chart-epic-param.test.tsx` (5 integration). Full suite: 25 files / 243 tests passing locally.
- The integration test asserts the entire URL→prop→mutation chain by spying on `triggerSync.mutate`, including the malformed-param security path and the 64-char boundary. Coverage is sufficient.

### Style / Zero attribution
- No Claude / AI attribution anywhere. Commit message clean.

### Suggestion (non-blocking)
- The same FHIR-id grammar lives implicitly in `services/api-gateway/src/routers/epic-sync.ts:88,93,99` as bare `z.string().optional()`. If you want to harden the backend too, a small `epicPatientFhirId: z.string().regex(/^[A-Za-z0-9.\-]{1,64}$/)` on the backend would mean a malformed query string would also be rejected over the wire even if a future frontend forgot to validate. Optional follow-up.

Issue #1191, opened 2026-05-25
feat(epic): dedup Epic-synced resources against pre-existing CareBridge rows


blamechris:
## Context

Discovered during review of #1183 (Encounter sync wiring). The Epic sync correctly dedups *re-syncs of the same Epic resource* via the `fhir_resources` mapping table — `findMapping(resourceType, fhirId)` finds the existing internal row and updates in place rather than inserting a duplicate.

But the mapping is **Epic-id → CareBridge-id**, not **logical-encounter → CareBridge-id**. If the same logical row exists in CareBridge from the internal-creation flow (no `fhir_resources` mapping row exists for it), and then an Epic sync arrives carrying the same encounter / diagnosis / medication / etc., a duplicate row lands in the target table.

This applies to all Epic-synced resources, not just Encounter:
- `persistEncounter` (`services/epic-connector/src/sync/persistence.ts:654`)
- `persistCondition` (line 489)
- `persistAllergy` (line 569)
- `persistMedicationRequest` (line 240)
- `persistObservation` (line 344)
- `persistPatient` (line 165) — likely safest because MRN-via-mrn_hmac is a unique index

## Why it matters

A clinician documents an emergency-room visit in CareBridge → 2 hours later the Epic sync pulls the same visit from Epic's EHR → the chart now shows two encounters for the same event. AI oversight reviews both, which means duplicate flags. Patient-portal timelines show duplicates.

## Possible approaches (out of scope for the first cut)

1. **Composite-key dedup before insert.** For Encounter: `(patient_id, start_time, encounter_type)`. Fragile for boundary cases (encounter starts at 23:59, Epic recorded it as 00:00 next day) but covers the common path.
2. **Epic-id-on-row column** (depends on #1190). Once `encounters.epic_encounter_id` exists, the dedup query can be `SELECT id FROM encounters WHERE patient_id = ? AND epic_encounter_id = ?` — much narrower.
3. **Reconciliation worker** that finds suspected duplicates by composite-key match and merges them, surfacing the merge in audit_log.

Option (2) layered on top of (1) is probably the right answer: Epic-id match is the canonical dedup, composite-key is the fallback for the first-import case where the same logical encounter pre-existed in CareBridge.

## Acceptance criteria

- [ ] Dedup query implemented in each `persist*` function before the INSERT branch
- [ ] Test case: pre-existing CareBridge encounter + Epic sync of same logical encounter → no duplicate row created
[...]

PR #1183, opened 2026-05-25
feat(epic): Encounter converter to complete #390 (#1181)

blamechris:
## Summary

Closes the last gap from #390 — Epic FHIR `searchEncounters` had no companion converter or sync wiring. This PR adds:

- `epicEncounterToRow(fhirEncounter, patientId)` in `services/epic-connector/src/converters.ts` mapping FHIR Encounter → `encounters` table row with `source_system: "epic"`
- `persistEncounter` in `sync/persistence.ts` with the same upsert/conflict/mapping pattern as the other resources
- Encounter arm in the sync worker per-resource fan-out (`SUPPORTED_RESOURCE_TYPES`, `persistOne`, `eventTypeFor`)
- `encounter.created` / `encounter.updated` clinical-event types in `@carebridge/shared-types`
- `epic_sync_state` row tracking for the `Encounter` resource type via the existing `markRunning` / `markOk` / `markFailed` path

### Schema choices

The `encounters` table is minimal (`id`, `patient_id`, `encounter_type`, `status`, `start_time`, `end_time`, `provider_id`, `location`, `reason`, `notes`, `created_at`). It has no `epic_encounter_id` or `source_system` columns today, so the converter return type (`EpicEncounterRow`) surfaces both for downstream callers but the writer only inserts columns that exist — matching the convention used by `epicConditionToRow` (the `diagnoses` table also doesn't carry `source_system`). The FHIR resource id round-trips through the `fhir_resources` mapping table as it does for every other resource type. A future migration can lift those fields onto the table without changing the converter.

`provider_id` is deliberately not mapped — FHIR `Encounter.participant.individual.reference` points at a Practitioner resource, and Practitioner→`users.id` reconciliation needs a separate sync flow (out of scope for #1181).

### Status codeset

FHIR R4 Encounter.status values (`planned | arrived | triaged | in-progress | onleave | finished | cancelled | entered-in-error | unknown`) flow through unchanged. Any value outside the codeset is coerced to `"unknown"` so downstream filters can rely on the contract.

## Test plan

- [x] Minimal Encounter (status + subject + period) → row with `source_system: "epic"`
- [x] Full Encounter (CSN identifier, reasonCode text, participant, location display) → all optional fields populated
- [x] Missing `subject` → returns `null`
- [x] Missing `status` → returns `null`
- [x] Missing `period.start` → returns `null` (start_time is notNull)
- [x] Unknown status code → falls through to `"unknown"`
- [x] Every FHIR R4 status code preserved verbatim
[...]

PR #1196, opened 2026-05-25
chore(db): source_system columns for encounters/diagnoses/allergies (#1190)

### LINKED ISSUE #1190: chore(db): add source_system column to encounters/diagnoses/allergies (Epic provenance gap)
## Context

Discovered during review of #1183 (Encounter converter, the last gap of #390). The Epic converters surface `source_system: "epic"` on the row shape they return (and the test suite asserts it), but three target tables don't have the column:

- `encounters` — `packages/db-schema/src/schema/encounters.ts` (no `source_system`, no `epic_encounter_id`)
- `diagnoses` — `packages/db-schema/src/schema/patients.ts:32` (no `source_system`)
- `allergies` — `packages/db-schema/src/schema/patients.ts:47` (no `source_system`)

Compare to `medications`, `vitals`, `lab_panels`, `procedures` (all in `packages/db-schema/src/schema/clinical-data.ts`) which carry `source_system: text("source_system").default("internal")`.

## Why it matters

Today, an Epic-sourced encounter / diagnosis / allergy is indistinguishable from a CareBridge-sourced one at the row level. To answer "where did this row come from?" requires a JOIN through `fhir_resources` keyed by `(resource_type, internal_record_id)`. That's a non-trivial query and a perf hazard for chart-listing pages that filter by source.

Per-row `source_system` also unlocks:
- `WHERE source_system = 'epic'` reporting (admin tooling, conflict dashboards)
- The source-conflict guard in `persistence.ts` (`existingSourceSystem != "epic"` → log conflict, refuse Epic overwrite) — currently relies on the mapping-row's `source_system`, which is set by the *first* writer. Adding the column to the target row would make the conflict check defendable against direct SQL inserts that bypass the mapping table.

## Proposed migration

Add `source_system text default 'internal'` to:
- `encounters`
- `diagnoses`
- `allergies`

Backfill is trivial: existing rows are CareBridge-originated by definition, so the default covers them. Epic-sync writers (`persistence.ts`) should then set `source_system: EPIC_SOURCE_SYSTEM_TAG` on insert and stop relying on the `fhir_resources.source_system` field for that signal.

## Out of scope

`epic_encounter_id` as a first-class column on `encounters` (Epic CSN — Contact Serial Number). Today it round-trips through the `fhir_resources` mapping row. Adding it would let dashboards link directly to Epic's encounter view by CSN without a JOIN — file separately if needed.

## Acceptance criteria

- [ ] Drizzle migration adds `source_system text default 'internal'` to `encounters`, `diagnoses`, `allergies`
- [ ] Epic persistence writers (`persistEncounter`, `persistCondition`, `persistAllergy`) set the column on insert
- [ ] Source-conflict guard reads from the target row's `source_system` (not just the mapping row)
- [ ] Existing tests pass; add a test that re-running Epic sync on a CareBridge-originated diagnosis logs a conflict and does NOT overwrite

## References

- #1183 (Encounter converter — surfaced this gap)
- #390 (Epic FHIR client — parent epic)
- `packages/db-schema/src/schema/clinical-data.ts` — the pattern to follow

blamechris:
## Summary

Adds `source_system TEXT NOT NULL DEFAULT 'internal'` to `encounters`, `diagnoses`, and `allergies` tables. Brings the schema in line with `medications` / `vitals` / `lab_panels` / `procedures` which already carry the column.

Closes the gap surfaced during #1183 review — Epic converters were always returning `source_system: "epic"` on the row shape, but for these three tables the value was being silently dropped on insert because the column did not exist. The persistence layer now threads `row.source_system` through to the insert.

## Migration

Single additive migration (`drizzle/0054_source_system_columns.sql`, journal idx 58). The `DEFAULT 'internal'` populates every existing row in the same statement — Pattern A from `docs/migration-patterns.md`, special case where pre-clean is unnecessary because the default IS the cleanup. No `NOT VALID + VALIDATE` required.

```sql
ALTER TABLE encounters ADD COLUMN source_system TEXT NOT NULL DEFAULT 'internal';
ALTER TABLE diagnoses  ADD COLUMN source_system TEXT NOT NULL DEFAULT 'internal';
ALTER TABLE allergies  ADD COLUMN source_system TEXT NOT NULL DEFAULT 'internal';
```

## Files touched

- `packages/db-schema/drizzle/0054_source_system_columns.sql` — new migration
- `packages/db-schema/drizzle/meta/_journal.json` — idx 58 entry
- `packages/db-schema/src/schema/encounters.ts` — add column
- `packages/db-schema/src/schema/patients.ts` — add column to `diagnoses` + `allergies`
- `packages/db-schema/src/__tests__/migration-0054-source-system.test.ts` — SQL + Drizzle column shape tests
- `services/epic-connector/src/sync/persistence.ts` — pass `source_system: EPIC_SOURCE_SYSTEM_TAG` on insert for the three tables (the converters already returned it; the writer was dropping it)
- `services/epic-connector/src/__tests__/persistence-source-system.int.test.ts` — gated integration test (TEST_DATABASE_URL) asserting end-to-end that `persistEncounter` / `persistCondition` / `persistAllergy` write `source_system: "epic"` and the migration default is `'internal'`
- `services/fhir-gateway/src/__tests__/us-core-meta-profile.test.ts` — fixture rows updated to include `source_system: "internal"` (context-field drift, per CLAUDE.md memory)

## Unlocks

- `WHERE source_system = 'epic'` reporting on encounters/diagnoses/allergies
- Cleaner cross-source dedup (see #1191, in flight)
- Row-level provenance without a JOIN through `fhir_resources`

## Test plan

- [x] `pnpm typecheck` clean
- [x] `pnpm lint` clean
- [x] `pnpm --filter @carebridge/db-schema test` — 50/50 pass (new SQL + Drizzle column tests green)
- [x] `pnpm --filter @carebridge/epic-connector test` — 271 pass / 4 skipped (integration tests gated on TEST_DATABASE_URL)
- [x] `pnpm --filter @carebridge/fhir-gateway test` — 500/500 pass (fixture drift fixed)
- [x] `pnpm test` — all 41 workspaces green
- [ ] CI runs migration against fresh DB and the new int test in `persistence-source-system.int.test.ts` verifies `persistEncounter`/`persistCondition`/`persistAllergy` write `source_system: "epic"` against real Postgres

Closes #1190

### DIFF
--- a/packages/db-schema/drizzle/0054_source_system_columns.sql
+++ b/packages/db-schema/drizzle/0054_source_system_columns.sql
@@ -1,0 +1,23 @@
+-- #1190: source_system provenance columns for encounters/diagnoses/allergies.
+--
+-- The medications, vitals, lab_panels, and procedures tables already carry
+-- a source_system text column (DEFAULT 'internal') so callers can answer
+-- "where did this row come from?" without joining through fhir_resources.
+-- These three tables were the only remaining gaps; Epic-sourced rows here
+-- are currently indistinguishable from CareBridge-originated rows at the
+-- row level, which blocks WHERE source_system = 'epic' reporting and
+-- cross-source dedup.
+--
+-- Pattern A from docs/migration-patterns.md, special case: the DEFAULT
+-- 'internal' populates every existing row in the same ADD COLUMN
+-- statement, so the pre-clean step is unnecessary — the default IS the
+-- cleanup. No CHECK constraint needed today; downstream code already
+-- treats source_system as an open-set tag ("internal", "epic",
+-- "fhir_import").
+--
+-- Pure additive — no NOT VALID + VALIDATE dance required. Safe on a
+-- populated DB: PostgreSQL records the default in pg_attribute and
+-- existing rows are filled in-place.
+ALTER TABLE encounters ADD COLUMN source_system TEXT NOT NULL DEFAULT 'internal';
+ALTER TABLE diagnoses  ADD COLUMN source_system TEXT NOT NULL DEFAULT 'internal';
+ALTER TABLE allergies  ADD COLUMN source_system TEXT NOT NULL DEFAULT 'internal';
--- a/packages/db-schema/drizzle/meta/_journal.json
+++ b/packages/db-schema/drizzle/meta/_journal.json
@@ -407,6 +407,13 @@
       "when": 1747930200000,
       "tag": "0053_npi_nucc_check",
       "breakpoints": true
+    },
+    {
+      "idx": 58,
+      "version": "7",
+      "when": 1747930260000,
+      "tag": "0054_source_system_columns",
+      "breakpoints": true
     }
   ]
 }
--- a/packages/db-schema/src/__tests__/migration-0054-source-system.test.ts
+++ b/packages/db-schema/src/__tests__/migration-0054-source-system.test.ts
@@ -1,0 +1,120 @@
+import { describe, it } from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+import { fileURLToPath } from "node:url";
+import { getTableColumns } from "drizzle-orm";
+import { encounters, diagnoses, allergies } from "../schema/index.js";
+
+const __dirname = fileURLToPath(new URL(".", import.meta.url));
+const MIGRATION_PATH = join(
+  __dirname,
+  "..",
+  "..",
+  "drizzle",
+  "0054_source_system_columns.sql",
+);
+
+describe("migration 0054_source_system_columns (#1190)", () => {
+  const sql = readFileSync(MIGRATION_PATH, "utf8");
+
+  it("adds source_system NOT NULL DEFAULT 'internal' to encounters", () => {
+    // Pure additive ADD COLUMN — the DEFAULT 'internal' populates every
+    // existing row in the same statement, so no NOT VALID + VALIDATE
+    // dance is required (Pattern A from docs/migration-patterns.md
+    // without the pre-clean step because the default IS the cleanup).
+    assert.match(
+      sql,
+      /ALTER\s+TABLE\s+encounters\s+ADD\s+COLUMN\s+source_system\s+TEXT\s+NOT\s+NULL\s+DEFAULT\s+'internal'/i,
+      "migration must add source_system NOT NULL DEFAULT 'internal' to encounters",
+    );
+  });
+
+  it("adds source_system NOT NULL DEFAULT 'internal' to diagnoses", () => {
+    assert.match(
+      sql,
+      /ALTER\s+TABLE\s+diagnoses\s+ADD\s+COLUMN\s+source_system\s+TEXT\s+NOT\s+NULL\s+DEFAULT\s+'internal'/i,
+      "migration must add source_system NOT NULL DEFAULT 'internal' to diagnoses",
+    );
+  });
+
+  it("adds source_system NOT NULL DEFAULT 'internal' to allergies", () => {
+    assert.match(
+      sql,
+      /ALTER\s+TABLE\s+allergies\s+ADD\s+COLUMN\s+source_system\s+TEXT\s+NOT\s+NULL\s+DEFAULT\s+'internal'/i,
+      "migration must add source_system NOT NULL DEFAULT 'internal' to allergies",
+    );
+  });
+
+  it("does not touch any other tables (single-purpose migration)", () => {
+    // Defends against accidental scope creep — the migration must only
+    // alter the three tables documented in the PR. ALTER TABLE statements
+    // for any other table should fail this check.
+    const alterMatches = Array.from(
+      sql.matchAll(/ALTER\s+TABLE\s+(\w+)/gi),
+      (m) => m[1].toLowerCase(),
+    );
+    const allowed = new Set(["encounters", "diagnoses", "allergies"]);
+    const unexpected = alterMatches.filter((t) => !allowed.has(t));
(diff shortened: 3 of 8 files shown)


blamechris (commented):
## Review verdict: APPROVE-WITH-FIX (CI fix landed via push)

**Migration safety (Pattern A — special case):** Clean. The migration is `ADD COLUMN ... NOT NULL DEFAULT 'internal'` ×3, which on Postgres ≥11 is the metadata-only fast path — the default lands in `pg_attribute.attmissingval` and existing rows are NOT rewritten under `ACCESS EXCLUSIVE`. Atomic per statement, so no nullable window. Migration is appropriately self-documenting; cites `docs/migration-patterns.md` and explains why the pre-clean step from canonical Pattern A is unnecessary (the default IS the cleanup). No `NOT VALID + VALIDATE` dance needed.

**Journal idx (58):** Verified — previous max was 57 (`0053_npi_nucc_check`). Single contributor on this PR so no collision risk per the parallel-subagent migration-collision memory. Schema test `does not reference SQL files that do not exist on disk` and `has strictly increasing idx values starting at 0` both pass.

**Schema/SQL parity:** Three Drizzle schemas updated with `text("source_system").notNull().default("internal")` matching the SQL exactly. Type via `$inferSelect` will surface `source_system: string` (non-null) on row reads — distinct from the older `medications`/`vitals`/`lab_panels`/`procedures` columns which are nullable (`.default("internal")` without `.notNull()`). That distinction is fine — the new tables get the stricter, correct shape. Schema parity tests (`encounters/diagnoses/allergies.source_system is NOT NULL with default 'internal'`) green.

**Persistence layer:** `persistEncounter`/`persistCondition`/`persistAllergy` correctly stamp `source_system: EPIC_SOURCE_SYSTEM_TAG` on insert. On UPDATE (re-sync of an Epic-owned row), `source_system` is intentionally NOT updated — correct, because the column is provenance and the row stays Epic-owned. Internal-creation flows are unmodified — they get `'internal'` via the column default. No need to touch CareBridge-originated writers.

**Fixture drift fix (`us-core-meta-profile.test.ts`):** Correct fix per the CLAUDE.md context-field-drift memory. The three `makeDiagnosis`/`makeAllergy`/`makeEncounter` helpers build literal-object rows and the test uses `as Diagnosis`/`as Allergy`/`as EncounterRow` casts. Adding `source_system: "internal"` makes them semantically accurate — that's what the DB would return for an internal-originated row. Other fhir-gateway test files (`allergy-intolerance.test.ts`, `encounter.test.ts`) use `Partial<X>` cast-spread which side-steps the missing field at the type level and is unaffected at runtime since generators don't read `source_system`.

**Acceptance criteria coverage from #1190:**
- [x] Drizzle migration adds `source_system text default 'internal'` ✓
- [x] Epic persistence writers set the column on insert ✓
- [ ] **Source-conflict guard reads from the target row's `source_system` (not just the mapping row)** — DEFERRED. The conflict guard still reads only `existingSourceSystem` from `fhir_resources` (the mapping row). The issue calls this out as defending against direct SQL inserts that bypass the mapping table. **Recommend filing a follow-up issue** for this, since the PR body says "Closes #1190" but only delivers 2 of 4 ACs. Alternatively, expand scope on this PR.
- [ ] **Add a test that re-running Epic sync on a CareBridge-originated diagnosis logs a conflict and does NOT overwrite** — DEFERRED. Same followup.

**Audit-log untouched:** Confirmed. `git diff --stat` shows no audit-log files. Append-only triggers from migration 0012 remain intact.

**Coordination with #1197 (cross-source dedup):** Verified independent. #1197's PR body explicitly says "Independent of #1190 — works with or without those columns. The fingerprint queries operate only on plaintext-indexed columns (codes, timestamps, MRN HMAC)." This PR doesn't change `source_system` semantics, just adds the column where it was missing.

**Reversibility:** Trivially reversible via `DROP COLUMN source_system` — pure additive metadata change.

**Zero attribution:** Clean — git log shows no Claude/Anthropic/Co-Authored-By.

---

## Finding — CI test failure (fixed in this push)

The new `persistence-source-system.int.test.ts` fails on CI because `seedPatient()` writes `patients.name` which is an `encryptedText` column, and CI's `test` job doesn't set `PHI_ENCRYPTION_KEY`. Local run reproduces the same `PHI_ENCRYPTION_KEY environment variable is not set` error when the key is absent.

Pushed a fix that sets a per-process throwaway `PHI_ENCRYPTION_KEY` at the top of the test file (gated on `TEST_URL`, set BEFORE the dynamic `@carebridge/db-schema` import since `getKey()` memoises). Verified locally: `pnpm vitest run src/__tests__/persistence-source-system.int.test.ts` against the local Postgres → 4/4 pass.

Self-contained-test approach over plumbing real `PHI_ENCRYPTION_KEY` into CI workflow env, since (a) the workflow doesn't need the production secret for any other reason, (b) the test only needs *some* key to round-trip through the encrypted column, and (c) it mirrors the pattern in `packages/db-schema/src/__tests__/encryption.test.ts` (generates `TEST_KEY` via `randomBytes(32)`).

## Follow-ups

1. **#1190 AC #3 + #4 not delivered** — recommend filing a follow-up issue for "source-conflict guard reads from target row's `source_system`" + the corresponding conflict-on-CareBridge-row test. Either that, or amend the PR body to say "Refs #1190" instead of "Closes #1190" per the `verify-issue-state-vs-code` memory (so #1190 stays open after merge).
2. `makeDiagnosis` in `us-core-meta-profile.test.ts` includes `updated_at: ...` (line 71) but the `diagnoses` schema has no `updated_at` column. Pre-existing, suppressed by the `as Diagnosis` cast. Out of scope for this PR — minor cleanup target.

PR #1197, opened 2026-05-25
feat(epic): cross-source dedup for Epic-synced resources (#1191)

### LINKED ISSUE #1191: feat(epic): dedup Epic-synced resources against pre-existing CareBridge rows

blamechris:
## Summary

Closes the latent duplication gap surfaced during #1183 review: when a CareBridge-originated row exists for a logical entity (encounter, diagnosis, etc.) and an Epic sync later arrives for the same entity, the sync was inserting a duplicate because the mapping lookup is keyed by Epic FHIR ID, not by clinical fingerprint.

This PR adds a fingerprint-based fallback in each `persist*` function in `services/epic-connector/src/sync/persistence.ts`. When the Epic-id mapping misses, the writer does a second lookup by clinical fingerprint — if a CareBridge-originated row matches, the sync updates in place AND creates a `fhir_resources` mapping row so future Epic re-syncs land on the same record. Every fingerprint hit is logged to `audit_log` with `action="epic_sync_dedup_match"`.

Independent of #1190 (source_system columns) — works with or without those columns. The fingerprint queries operate only on plaintext-indexed columns (codes, timestamps, MRN HMAC).

## Fingerprints

| Resource | Fingerprint |
|---|---|
| Patient | `mrn_hmac` (already unique-indexed; verified pre-existing safe) |
| Encounter | `patient_id + start_time + encounter_type` |
| Condition | `patient_id + (icd10_code OR snomed_code) + onset_date` |
| Allergy | `patient_id + (rxnorm_code OR snomed_code)` |
| MedicationRequest / Statement | `patient_id + rxnorm_code + started_at` |
| Observation (vital branch) | `patient_id + loinc_code + recorded_at` |

### Safety guards

- Lookups short-circuit to `null` when no coded dimension is present on the inbound Epic resource — better to tolerate the occasional duplicate than risk a false-merge on a dimensionally-weak fingerprint.
- Encrypted columns (`allergen`, `medications.name`, `lab_results.test_name`) are never compared directly; only their associated plaintext codes are.
- Lab branch of Observation: deferred to a follow-up — the `lab_results` leaf lacks `patient_id`, and the synthesised panel makes the dimensional weight of a fingerprint identical to the existing mapping-only path.

## Test plan

- [x] Per resource: pre-existing CareBridge row → Epic sync dedups in place, mapping row added (6 resources covered)
- [x] Per resource: no matching row → new insert, no spurious match
- [x] Patient (MRN-based) already-safe path unchanged
- [x] Audit log entry on every fingerprint match with `action="epic_sync_dedup_match"`
- [x] `findMapping` Epic-id fast-path still works (smoke test asserts only one SELECT when mapping is present, no fingerprint SELECT issued)
- [x] `pnpm typecheck && pnpm lint` — green
- [x] `pnpm --filter @carebridge/epic-connector test -- persistence` — 13 / 13 pass
- [x] Full repo `pnpm -r test` — all suites pass

Closes #1191

### DIFF
--- a/services/epic-connector/src/__tests__/persistence.test.ts
+++ b/services/epic-connector/src/__tests__/persistence.test.ts
@@ -1,0 +1,547 @@
+/**
+ * Cross-source dedup tests for Epic persistence layer (#1191).
+ *
+ * The existing path keys mapping by Epic FHIR id (resource_type +
+ * resource_id) and updates in place on re-sync. But if a CareBridge-
+ * originated row already exists for the same logical entity (no
+ * `fhir_resources` mapping row), an Epic sync of the same entity used
+ * to insert a duplicate. These tests pin the fix: each persist*
+ * function does a fingerprint fallback when the Epic-id mapping
+ * misses, updates the matched CareBridge row in place, creates the
+ * mapping row, and logs an audit entry.
+ *
+ * Uses the shared createMockDb harness so we stay decoupled from
+ * Postgres + Drizzle internals (see sync-state-repo.test.ts for the
+ * established pattern in this service).
+ */
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { createMockDb, type MockDb } from "@carebridge/test-utils";
+
+let db: MockDb;
+
+vi.mock("@carebridge/db-schema", () => ({
+  getDb: () => db,
+  patients: {
+    id: "id",
+    mrn_hmac: "mrn_hmac",
+    patient_id: "patient_id",
+  },
+  medications: {
+    id: "id",
+    patient_id: "patient_id",
+    rxnorm_code: "rxnorm_code",
+    started_at: "started_at",
+  },
+  vitals: {
+    id: "id",
+    patient_id: "patient_id",
+    loinc_code: "loinc_code",
+    recorded_at: "recorded_at",
+  },
+  labPanels: {},
+  labResults: {},
+  diagnoses: {
+    id: "id",
+    patient_id: "patient_id",
+    icd10_code: "icd10_code",
+    snomed_code: "snomed_code",
+    onset_date: "onset_date",
+  },
+  allergies: {
+    id: "id",
+    patient_id: "patient_id",
+    rxnorm_code: "rxnorm_code",
+    snomed_code: "snomed_code",
+  },
+  encounters: {
+    id: "id",
+    patient_id: "patient_id",
+    start_time: "start_time",
+    encounter_type: "encounter_type",
+  },
+  fhirResources: {
+    id: "id",
+    resource_type: "resource_type",
+    resource_id: "resource_id",
+    internal_record_id: "internal_record_id",
+  },
+  auditLog: {},
+  hmacForIndex: (value: string) => `hmac:${value}`,
+}));
+
+const {
+  persistPatient,
+  persistMedicationRequest,
+  persistObservation,
+  persistCondition,
+  persistAllergy,
+  persistEncounter,
+} = await import("../sync/persistence.js");
+
+const PATIENT_ID = "11111111-1111-1111-1111-111111111111";
+
+beforeEach(() => {
+  vi.clearAllMocks();
+  db = createMockDb();
+});
+
+// ── helpers ─────────────────────────────────────────────────────
+
+/**
+ * Collect every recorded INSERT and tag it by the table reference
+ * passed in. Useful because a single persist call can fan out to
+ * several tables (e.g. fhir_resources mapping, audit_log).
+ */
+function insertsByTable(): Array<{ tableArg: unknown; values: unknown }> {
+  return db.insert.calls.map((c) => ({
+    tableArg: c.args[0],
+    values: c.chainArgs[0]?.[0],
+  }));
+}
+
+function findInsertFor(
+  tableSentinel: Record<string, unknown>,
+): Record<string, unknown> | undefined {
+  const hit = insertsByTable().find((i) => i.tableArg === tableSentinel);
+  return hit?.values as Record<string, unknown> | undefined;
+}
+
+function findAllInsertsFor(
+  tableSentinel: Record<string, unknown>,
+): Array<Record<string, unknown>> {
+  return insertsByTable()
+    .filter((i) => i.tableArg === tableSentinel)
+    .map((i) => i.values as Record<string, unknown>);
+}
+
+// ── Encounter ───────────────────────────────────────────────────
+
+describe("persistEncounter — cross-source dedup (#1191)", () => {
+  const RESOURCE = {
+    resourceType: "Encounter",
+    id: "epic-enc-1",
+    status: "finished",
+    class: { code: "AMB" },
+    subject: { reference: "Patient/p-1" },
+    period: { start: "2026-05-20T09:00:00Z", end: "2026-05-20T09:45:00Z" },
+  };
+
+  it("dedups against an existing CareBridge-originated encounter by patient_id+start_time+encounter_type and inserts a mapping row", async () => {
+    const existingInternalId = "internal-enc-1";
+    // findMapping → no Epic mapping yet
+    db.willSelect([]);
+    // findByFingerprint → existing internal row matches the fingerprint
+    db.willSelect([
+      {
+        id: existingInternalId,
+        patient_id: PATIENT_ID,
+        start_time: "2026-05-20T09:00:00Z",
+        encounter_type: "AMB",
+      },
+    ]);
+    // UPDATE the matched encounters row
+    db.willUpdate();
+    // UPSERT the fhir_resources mapping row
+    db.willInsert();
+    // audit_log insert (epic_sync_dedup_match)
+    db.willInsert();
+
+    const result = await persistEncounter(RESOURCE, PATIENT_ID);
+
+    expect(result.internalId).toBe(existingInternalId);
+    expect(result.kind).toBe("encounter");
+    expect(result.inserted).toBe(false);
+    // No new encounters row was inserted — only the mapping + audit.
+    expect(findAllInsertsFor({} as never)).toBeDefined();
+
+    // Verify mapping row was upserted pointing at the existing internal id.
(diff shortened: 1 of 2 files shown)


blamechris (commented):
## CareBridge Inspector — clinical-safety + data-integrity review (#1197)

Verdict: **approve with follow-ups** (no blockers). The fingerprint-based fallback is clean, well-instrumented, and the audit-log integration is solid. Comments below capture two semantic risks worth filing as follow-ups before merge, plus one nit on test coverage.

### Strengths

- **Audit-log integration is correct.** Every fingerprint hit writes a `epic_sync_dedup_match` row with `success: true`, distinguishing it from `source_system_conflict` writes (`success: false`). Payload carries `epic_fhir_id`, the matched internal_id, and the fingerprint dimensions — enough to recover the link manually if needed. Append-only constraint (migration 0012) is respected.
- **Independence from #1190 is verified.** `encounters` / `diagnoses` / `allergies` tables have no `source_system` column on `main` today (confirmed against `packages/db-schema/src/schema/encounters.ts` and `patients.ts`); the defensive `(hit as { source_system?: string | null }).source_system ?? null` casts in `findEncounterByFingerprint` / `findDiagnosisByFingerprint` / `findAllergyByFingerprint` work without the columns. Typecheck + lint + 284 tests pass locally.
- **Safety short-circuits.** All non-MRN fingerprints return `null` when no coded dimension is present (`!args.icd10Code && !args.snomedCode`, `!args.rxnormCode`, `!args.loincCode`) — the writer falls through to insert rather than risk a false-merge on a dimensionally-weak fingerprint. This is the right trade-off for clinical data.
- **Patient MRN path is a no-op regression-wise.** The MRN unique index already enforces this dedup at the DB layer; the new `findPatientByFingerprint` is a belt-and-braces optimisation to avoid a unique-violation round-trip. `computeMrnHmacIfPresent` returns null on HMAC helper failure so the lookup degrades safely.
- **Test harness matches the established pattern.** `createMockDb` is the same fluent mock used by `sync-state-repo.test.ts` and `flag-push.test.ts` — consistent with how this service tests Drizzle chains today.
- **Encounter / Condition / Medication / Vital fingerprint dimensions are clinically defensible.** Different start_time = different encounter; different rxnorm + started_at = restart of a med; different LOINC + recorded_at = different observation event. Onset_date nullable widening (line 274-280) correctly handles the case where CareBridge row has no onset but Epic does.

### Findings worth filing follow-ups for

#### 1) Source-system policy gap in the fingerprint-hit branch (medium)

The header doc (lines 21-25) describes a `source_system != "epic"` conflict check that refuses to overwrite CareBridge-originated rows. That check exists on the Epic-id mapping path (line 413-430, 561-578, etc.) but **NOT on the fingerprint-hit path**. When fingerprint matches a CareBridge row, the writer unconditionally UPDATEs encounter_type / status / end_time / location / reason from the Epic payload (e.g. line 1196-1204), overwriting fields a clinician may have edited.

This is also inconsistent with the policy in the issue acceptance criteria: "When dedup matches, the resource is treated as an UPDATE (or skipped with `inserted: false`)". Treating-as-update with full Epic-side overwrite is one valid interpretation, but it's not what the conflict-check comment promises.

Two possible resolutions:
- A) Add a `source_system != "epic"` guard on the fingerprint branch that either skips the UPDATE (mapping-only + audit) or merges with a field-level last-writer policy.
- B) Document explicitly that fingerprint-merge always overwrites CareBridge fields, and remove the "Epic does not get to overwrite CareBridge-originated data" wording from the header (which currently reads as a guarantee).

I'd suggest filing a follow-up and going with (B) for now if (A) is too disruptive — but flag this is a clinical-safety surface; a clinician-entered encounter note shouldn't silently lose its location/reason on Epic re-sync.

#### 2) Fingerprint-hit on Epic-mapped row treated as silent merge (medium)

If the same `patient_id + start_time + encounter_type` row is **already mapped** to a different Epic FHIR id (e.g., Epic re-issued a new resource id for the same encounter), the fingerprint hit at line 1194 returns that row and `upsertMapping` writes a SECOND `fhir_resources` row pointing at the same internal_id. The two Epic FHIR ids now both round-trip through this internal row, which is probably the desired behaviour for "Epic re-issued the ID" but may also indicate two genuinely-distinct logical entities collapsing.

The brief from the issue says this case should "log to audit_log + create a new row". Today the implementation logs to audit_log via `epic_sync_dedup_match` (good) but never creates a new row (just merges). Worth a follow-up to detect this — `findByFingerprint` could also check `fhirResources.resource_type=X AND internal_record_id=hit.id` and return that count so the writer can decide.

#### 3) Allergy fingerprint has no temporal dimension (PR-acknowledged)

Already flagged in PR description and code comment (line 1054-1056). Two allergies re-evaluated for the same SNOMED/RxNorm code (e.g. patient initially flagged penicillin allergy then re-confirmed years later under a new entry) will merge. For allergies this is arguably correct — same substance, same allergy — but `verification_status` and `severity` may diverge between the two records and Epic's payload will silently overwrite.

#### 4) Observation tight-time-window collision risk (low)

`(patient_id, loinc_code, recorded_at)` with `recorded_at` precise to seconds means two clinicians charting the same vital from different devices ~1 second apart could either dedupe (if Epic rounds) or duplicate (if clocks differ). Acceptable for v1; document or add a ± tolerance later.

#### 5) Lab branch fingerprint dedup deferred (acknowledged)

Lines 806-814 document this in code. PR body mentions follow-up. Worth filing as a tracked issue rather than only an inline comment.

### Performance

Fingerprint lookups add one SELECT per row when the Epic-id mapping misses. For a 10K-row backfill the cost is bounded — existing indexes cover the patient_id leading column on every table (`idx_encounters_patient`, `idx_diagnoses_patient`, `idx_allergies_patient`, `idx_medications_patient`, `idx_vitals_patient_type`). The composite `(patient_id, type, recorded_at)` index on vitals is the best fit. No critical missing index. Worth filing a follow-up to add a `(patient_id, loinc_code, recorded_at)` index on vitals and a `(patient_id, rxnorm_code, started_at)` index on medications if backfills get slow.

### Tests

- 13 tests, 6 resources, positive + negative + audit + Epic-id fast-path. Each meaningful.
- Uses `createMockDb` (the established mock harness, NOT a raw `vi.fn` mock) — matches sibling test files.
- The fast-path test (`findMapping fast-path remains intact`) asserts exactly one `select` call when mapping is present, pinning the optimisation. Good.

Note: tests rely on `db.willSelect([])` queue ordering matching the call order in production code. Adding a future SELECT (e.g. for the source-system check in finding 1) will break these tests subtly — worth a note for whoever does the follow-up.

### Zero-attribution

Verified. No co-author lines or Claude references in commits, code, or PR body.

### Summary

LGTM with the follow-ups above. Findings 1 + 2 are the higher-priority clinical-safety items; findings 3-5 are tracked-and-documented but worth a tagged issue each.

blamechris:
Filed follow-ups from the review:

- #1200 — source_system guard on fingerprint-merge path (medium / clinical-safety)
- #1201 — detect Epic-id conflict on fingerprint merge (medium)
- #1202 — lab branch fingerprint dedup (deferred branch — tracked)
- #1203 — allergy fingerprint temporal dimension (low / data-integrity)
- #1204 — composite indexes for fingerprint dedup lookups (performance)

None of these are blockers for #1197 — the current PR is a strict improvement over the latent-duplicate behaviour it replaces. Recommend merging and addressing the source_system guard (#1200) next.

PR #1198, opened 2026-05-25
fix(portal): epic-launch-banner flash on suppressed re-renders (#1187)

### LINKED ISSUE #1187: fix(portal): epic-launch-banner flashes briefly on suppressed re-renders
## Context

\`epic-launch-banner.tsx\` checks the sessionStorage suppression flag inside \`useEffect\`, which only fires after mount. The component returns \`null\` only when \`suppressed === true\`, but \`suppressed\` starts as \`false\`. Result: the banner paints on every navigation that carries \`?epic_patient=...\` even after the user clicked \"Switch to portal mode\", because the suppression-flag read happens post-paint.

\`\`\`tsx
const [suppressed, setSuppressed] = useState(false);
useEffect(() => {
  if (!inLaunchContext) return;
  try {
    const ss = storage ?? globalThis.sessionStorage;
    if (ss?.getItem(SUPPRESS_KEY) === epicPatient) {
      setSuppressed(true);
    }
  } catch { /* ... */ }
}, [inLaunchContext, epicPatient, storage]);
\`\`\`

In practice the \"Switch to portal mode\" handler also strips the query params via \`router.replace\`, so the banner only flashes briefly between Suspense resolution and effect run — but on direct deep-link re-visits with the params still present, the banner will flash.

## Required

Compute initial suppression synchronously during render using a \`useState\` lazy initializer that reads \`sessionStorage\` once (guarded by \`typeof window !== \"undefined\"\`), so the banner is never visible to a suppressed user.

## Acceptance

- [ ] Visit \`/patients/abc?epic_patient=X\` after clicking \"Switch to portal mode\" once: banner does not flash.
- [ ] Test: pre-populate sessionStorage with the suppression flag, render with the launch params on the URL, assert the banner is not in the DOM at all (including initial mount).

## References

- PR #1184

blamechris:
## Summary

Replaces the post-mount `useEffect` sessionStorage read in `epic-launch-banner.tsx` with a `useState` lazy initializer so the banner doesn't paint between Suspense/query resolution and the effect run. Builds on #1186's server-context restructuring.

- Lazy `useState` initializer reads sessionStorage during the first render, before paint
- `typeof window === "undefined"` guard keeps SSR-safe
- `try/catch` covers Safari private mode and sandboxed-iframe blocked-storage
- Suppress key uses the server-confirmed `epic_patient_fhir_id` (matches #1186)
- The follow-up `useEffect` stays in place for the cold-mount case (server data resolves after first render) and the cross-tab edge case

## Test plan

- [x] Existing 8 banner tests still pass (no regressions)
- [x] New TDD test asserts sessionStorage `getItem` is consulted during the synchronous render path (via a render-order probe), not deferred to a post-mount effect
- [x] With suppress flag set, banner never renders (no flash)
- [x] Without suppress flag, banner renders normally
- [x] SSR-safe (`typeof window` guard)

Closes #1187
Depends on #1192

### DIFF
--- a/apps/clinician-portal/src/__tests__/epic-launch-banner.test.tsx
+++ b/apps/clinician-portal/src/__tests__/epic-launch-banner.test.tsx
@@ -245,4 +245,64 @@ describe("EpicLaunchBanner", () => {
     expect(banner).toBeInTheDocument();
     expect(banner.textContent).toMatch(/abc/);
   });
+
+  // Issue #1187 — suppression flag must be read synchronously during render
+  // (lazy useState initializer), not in a post-mount useEffect. Reading it
+  // post-mount causes the banner to paint once before the effect hides it,
+  // producing a visible flash on deep-link re-visits with launch params still
+  // present in the URL.
+
+  it("reads sessionStorage suppress flag during initial render (lazy init, not via effect)", () => {
+    // Contract: with the lazy useState initializer, getItem must be called
+    // as part of the synchronous render path. Without the fix, getItem is
+    // only called inside useEffect — which fires AFTER the first paint.
+    //
+    // The probe below renders AFTER the banner in tree order. React calls
+    // function components top-down during render, so by the time Probe's
+    // render function executes, the banner's render function has already
+    // run to completion (including any lazy useState initializer). If
+    // getItem was called by then, the lazy-init path was exercised; if not,
+    // the read is deferred to an effect and the banner has already painted.
+    searchParamsMock = new URLSearchParams({
+      epic_patient: "eP123",
+    });
+    getLaunchContextQuery.mockReturnValueOnce({
+      data: {
+        epic_org_iss: "https://fhir.epic.example/api/FHIR/R4",
+        epic_practitioner_fhir_id: "prac-1",
+        epic_patient_fhir_id: "eP123",
+        launch_encounter_fhir_id: null,
+        expires_at: "2099-01-01T00:00:00Z",
+      },
+      isLoading: false,
+      isError: false,
+    });
+
+    const { store, storage } = makeStorage();
+    store.set("carebridge_epic_launch_suppressed", "eP123");
+
+    const getItemSpy = vi.fn((k: string) => storage.getItem(k));
+    const spyStorage: Pick<Storage, "getItem" | "setItem" | "removeItem"> = {
+      getItem: getItemSpy,
+      setItem: (k, v) => storage.setItem(k, v),
+      removeItem: (k) => storage.removeItem(k),
+    };
+
+    let getItemCallsSeenByProbe = 0;
+    function Probe() {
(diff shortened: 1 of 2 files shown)


blamechris (commented):
**Delta-only review against PR #1192 base (`fix/issue-1186-banner-trust`).**

## Verdict: APPROVE

Clean, surgical fix that targets the exact failure path called out in #1187 without disturbing the security restructuring from #1186.

## What was reviewed (delta vs #1192 base)
- `apps/clinician-portal/src/components/epic-launch-banner.tsx` — replaces `useState(false)` with `useState<boolean>(() => …)` lazy initializer.
- `apps/clinician-portal/src/__tests__/epic-launch-banner.test.tsx` — adds one TDD test using a render-order probe to assert `sessionStorage.getItem` is called during synchronous render.

## Correctness
- **Lazy init**: `useState(() => …)` closure runs exactly once on first render. Verified the closure:
  - returns `false` under SSR via `typeof window === "undefined"` guard.
  - returns `false` when `serverPatient` not yet resolved (the cold-tRPC-mount path) — the retained `useEffect` then handles that case on re-render.
  - reads `SUPPRESS_KEY === serverPatient`, where `serverPatient = ctx?.epic_patient_fhir_id` — the server-confirmed id introduced by #1186, not the URL value. Correct.
  - wraps `sessionStorage` access in `try/catch` for Safari private mode + sandboxed iframes.
- **Effect retention**: the follow-up `useEffect` (lines 91–101) is unchanged and still catches the warm-server / cold-cache mount where `serverPatient` resolves on render N+1.
- **No regression to anti-spoof gating**: all the #1186 invariants (server-truth, URL mismatch warns, no-ctx → null) are untouched.

## Test methodology
- Verified the new test fails on the unfixed code (asserted `getItemCallsSeenByProbe).toBeGreaterThan(0)` produces `expected 0 to be greater than 0` against the parent base), and passes on the fixed code.
- The probe-after-banner trick is a clean way to assert "render-time read" without depending on internal hook telemetry. All 9 tests pass.

## Scope honesty
- The inline comment correctly acknowledges the lazy init cannot fix the cold-tRPC-mount flash (no `serverPatient` available on render 0). The PR explicitly scopes itself to the deep-link re-visit case from #1187 where `staleTime: 60_000` gives the query a warm hit, and that's where it works. Comment language ("storage events would be needed for true cross-tab sync") is honest about the residual cross-tab gap — out of scope for this PR.

## Stack discipline
- Base: `fix/issue-1186-banner-trust` ✅
- Body: "Depends on #1192" ✅
- No accidental changes to parent files.

## Zero Attribution
- No Claude / Anthropic / Co-Authored-By mentions in commits or diff.

## Nits (non-blocking)
- None worth holding the PR for. The added comment block is generous and accurate.

Ready to merge once #1192 lands.

PR #1205, opened 2026-05-26
feat(portal): dedupe epic-launch-banner mismatch console.warn (#1199)

### LINKED ISSUE #1199: feat(portal): rate-limit epic-launch-banner mismatch console.warn
## Context

Discovered while reviewing PR #1192 (the #1186 anti-spoof banner fix).

`apps/clinician-portal/src/components/epic-launch-banner.tsx` emits a `console.warn` whenever the URL `epic_patient` (or `epic_encounter`) disagrees with the server-confirmed launch context:

```tsx
if (urlEpicPatient && urlEpicPatient !== ctx.epic_patient_fhir_id) {
  console.warn("[epic-launch-banner] URL epic_patient does not match server launch context; hiding banner.");
  return null;
}
```

These guards run inline in the render body, so each re-render of the banner component re-fires the warn. In practice:

- React 18 dev strict-mode renders the component twice → warn fires 2x per logical render.
- If a clinician is on a page with sticky launch params and React Query re-fetches `getLaunchContext` (or any parent re-renders), the warn fires again.
- If anything triggers frequent re-renders (e.g. parent state changes, route transitions), the dev console can fill with the same tamper signal.

This is a developer-experience / alarm-fatigue concern, not a correctness bug. The warn is functionally correct.

## Required

Rate-limit / dedupe the warn so it fires once per unique (urlPatient, serverPatient) mismatch pair per mount. Options:

1. **`useRef` guard**: `const warnedFor = useRef<string | null>(null);` — compare `${urlEpicPatient}|${ctx.epic_patient_fhir_id}` before warning.
2. **`useEffect` instead of inline**: move the warn into a `useEffect([urlEpicPatient, ctx.epic_patient_fhir_id])` so React's effect-deduping covers same-value re-renders. (Note: still fires on strict-mode double-mount; combine with option 1 if that matters.)
3. **Module-level `Set<string>`**: dedupe across the component lifetime entirely. Heavier — only worth it if mismatches are expected to bounce between values.

Option 1 is the smallest delta and likely sufficient.

## Acceptance

- [ ] Re-rendering the banner 10x with the same mismatched URL+server pair emits the warn at most once (or once per unique pair, dev strict-mode double-render OK).
- [ ] Test asserts `warnSpy` called exactly N times for N unique mismatches across re-renders.

## References

- PR #1192
- Lines 96-112 of `apps/clinician-portal/src/components/epic-launch-banner.tsx`

blamechris:
## Summary
- Adds a useRef-based dedup so the URL-vs-server mismatch console.warn fires at most once per unique (urlEpicPatient, serverPatient) pair per mount
- Applies same dedup to the encounter mismatch warn path

## Test plan
- [x] Re-rendering with same mismatch pair → exactly 1 warn
- [x] Switching to a different mismatch pair → second warn fires
- [x] No regression on the suppression / matching-context paths

Closes #1199

### DIFF
--- a/apps/clinician-portal/src/__tests__/epic-launch-banner.test.tsx
+++ b/apps/clinician-portal/src/__tests__/epic-launch-banner.test.tsx
@@ -305,4 +305,102 @@ describe("EpicLaunchBanner", () => {
     // And the steady-state DOM has no banner either.
     expect(screen.queryByTestId("epic-launch-banner")).toBeNull();
   });
+
+  // Issue #1199 — the mismatch console.warn runs inline in render, so React
+  // strict-mode double-renders and parent re-renders fired the same tamper
+  // warning repeatedly in dev. Dedup by tracking the last warned
+  // (urlEpicPatient, serverPatient) pair in a useRef and only re-warning
+  // when the pair changes.
+
+  it("warns only once when re-rendered repeatedly with the same patient mismatch", () => {
+    searchParamsMock = new URLSearchParams({ epic_patient: "spoof" });
+    getLaunchContextQuery.mockReturnValue({
+      data: {
+        epic_org_iss: "https://fhir.epic.example/api/FHIR/R4",
+        epic_practitioner_fhir_id: "prac-1",
+        epic_patient_fhir_id: "abc",
+        launch_encounter_fhir_id: null,
+        expires_at: "2099-01-01T00:00:00Z",
+      },
+      isLoading: false,
+      isError: false,
+    });
+
+    const warnSpy = vi.spyOn(console, "warn").mockImplementation(() => {});
+    const { storage } = makeStorage();
+    const { rerender } = render(<EpicLaunchBanner storage={storage} />);
+    for (let i = 0; i < 10; i++) {
+      rerender(<EpicLaunchBanner storage={storage} />);
+    }
+
+    expect(warnSpy).toHaveBeenCalledTimes(1);
+    warnSpy.mockRestore();
+  });
+
+  it("emits a fresh warn when the mismatch pair changes between renders", () => {
+    searchParamsMock = new URLSearchParams({ epic_patient: "spoof-1" });
+    getLaunchContextQuery.mockReturnValue({
+      data: {
+        epic_org_iss: "https://fhir.epic.example/api/FHIR/R4",
+        epic_practitioner_fhir_id: "prac-1",
+        epic_patient_fhir_id: "abc",
+        launch_encounter_fhir_id: null,
+        expires_at: "2099-01-01T00:00:00Z",
+      },
+      isLoading: false,
+      isError: false,
+    });
+
+    const warnSpy = vi.spyOn(console, "warn").mockImplementation(() => {});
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #1205

### Summary
Adds a `useRef`-based pair-key dedup so the URL-vs-server mismatch `console.warn` in `epic-launch-banner.tsx` fires at most once per unique `(urlValue, serverValue)` pair per mount. Implements exactly Option 1 from issue #1199 (`useRef` guard, smallest delta). Symmetric treatment for the patient-mismatch and encounter-mismatch branches.

### Strengths
- **Implementation matches the linked issue precisely.** Issue #1199 listed three options and recommended option 1 (`useRef` guard, compare `${url}|${server}`). PR ships exactly that.
- **Symmetric refs for both warn paths** (`warnedPatientPairRef` + `warnedEncounterPairRef`) — easy to reason about, no shared state between the two unrelated tamper signals.
- **Refs reset on remount semantics are intentional and documented** in the inline comment ("a fresh launch attempt will still log once") — preserves the "noise-floor signal" while removing dev-console spam.
- **Hook ordering is correct.** Both `useRef`s are unconditionally called before any conditional `return null`, so Rules of Hooks holds.
- **Test coverage is on-point and reads the way the issue's acceptance criteria asked for:**
  - 10x re-render same pair → exactly 1 warn (issue's acceptance bullet)
  - Pair change → fresh warn fires (boundary case the issue called out)
  - Encounter branch covered symmetrically
- **Comment quality.** The 7-line `Issue #1199:` block above the refs explains *why* the dedup exists (strict-mode double-render + parent re-renders) and the remount-reset trade-off. This is the right altitude — anyone reading this file later won't wonder why two refs sit between effects and early returns.

### Issues Found

#### Critical (Must Fix)
None.

None worth gating the PR on.

#### Nitpicks (Optional)
- **`pairKey` separator collision (theoretical).** `${urlEpicPatient}|${ctx.epic_patient_fhir_id}` would collide if a value contained `|`. FHIR resource IDs are restricted to `[A-Za-z0-9\-\.]{1,64}` per the R4 spec, so this is structurally impossible — flagging only for completeness. No action needed.
- **Encounter test's `"real-enc" as unknown as null` cast.** The comment explains the cast (vi.fn's inferred return type pins `launch_encounter_fhir_id` to `null` from the module-level default). An alternative would be widening the module default to `string | null`, but that touches a lot of unrelated tests and isn't worth it for one assertion. Current approach is fine.

### Architecture Notes
- Single-file client-component change. No tRPC, no Drizzle, no BullMQ, no AI-prompt versioning touched.
- Doesn't affect the `getLaunchContext` query or any server-side anti-spoof gate — only the dev-console rendering side-effect.
- Closes the `from-review` follow-up (#1199) created during the #1192 anti-spoof review. Tidy paper trail.

### Verdict
- [x] Approve — Ready to merge

CI is green (Build/Test/Typecheck all pass on HEAD `da645c3`). Implementation is minimal, well-commented, and the tests cover the exact acceptance criteria from #1199.

PR #1207, opened 2026-05-26
feat(epic): panel-level lab fingerprint dedup (#1202)

### LINKED ISSUE #1202: feat(epic): lab branch fingerprint dedup (#1197 follow-up)
Follow-up to #1197. The lab branch of `persistObservation` in `services/epic-connector/src/sync/persistence.ts` (line 806-870) deferred fingerprint dedup. Reason documented inline:

> Lab fingerprint dedup is also deferred — the leaf (lab_results) doesn't carry patient_id directly; matching would require a JOIN against the synthesised lab_panels row whose collected_at is itself derived from recorded_at, so the dimensional weight of the fingerprint is identical to the panel lookup.

Worth tracking explicitly rather than only as a code comment.

## Options to consider

1. **Panel-level fingerprint.** Fingerprint by `lab_panels(patient_id, panel_name, collected_at)`. Each panel has its own UUID; multi-result panels arriving from Epic as multiple Observations would have a single CareBridge panel row, and the existing lab_results rows would hang off it. Risk: Epic single-result Observations get a synthesised one-row panel, so the fingerprint key includes the test_name as panel_name — collision risk for the same test repeated under different panel names.
2. **lab_results.patient_id denormalisation.** Add `patient_id` column to lab_results, populate via trigger or app-side. Enables direct fingerprint by (patient_id, test_code, recorded_at). Bigger migration.
3. **DiagnosticReport-driven dedup.** When Epic sends a DiagnosticReport, all its result Observations get grouped into one CareBridge panel. Fingerprint at the report level. Out of scope until #391 followup wires DiagnosticReport ingest.

(1) is the most achievable today.

## Acceptance criteria

- [ ] Lab branch in `persistObservation` does a fingerprint lookup before INSERT
- [ ] Decision: panel-level vs. result-level (recommend panel-level for v1)
- [ ] Tests cover positive + negative + audit-log entry consistent with other resources
- [ ] Inline comment block updated or removed

## References

- #1197 (this PR — deferred the branch)
- #1191 (parent)
- #391 (Epic FHIR client — DiagnosticReport ingest, future enabler)

blamechris:
## Summary
- Adds `findLabPanelByFingerprint(patient_id, panel_name, collected_at)` helper in `services/epic-connector/src/sync/persistence.ts`
- Wires it into the lab branch of `persistObservation` so repeated panel syncs reuse the existing `lab_panels` row instead of inserting duplicates
- Removes the inline deferral comment from #1197

## Test plan
- [x] Positive: pre-existing panel + Epic re-sync → no new panel, lab_results attached to existing
- [x] Negative: no pre-existing panel → new panel created
- [x] Audit log entry `epic_sync_dedup_match` written on match
- [x] Single-result Observation: dedupe by test_name as panel_name

Closes #1202
Follows: #1197

### DIFF
--- a/services/epic-connector/src/__tests__/persistence.test.ts
+++ b/services/epic-connector/src/__tests__/persistence.test.ts
@@ -38,8 +38,16 @@ vi.mock("@carebridge/db-schema", () => ({
     loinc_code: "loinc_code",
     recorded_at: "recorded_at",
   },
-  labPanels: {},
-  labResults: {},
+  labPanels: {
+    id: "id",
+    patient_id: "patient_id",
+    panel_name: "panel_name",
+    collected_at: "collected_at",
+  },
+  labResults: {
+    id: "id",
+    panel_id: "panel_id",
+  },
   diagnoses: {
     id: "id",
     patient_id: "patient_id",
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #1207

### Summary
Adds panel-level fingerprint dedup for lab Observations synced from Epic. A new `findLabPanelByFingerprint(patient_id, panel_name, collected_at)` helper sits between the existing FHIR-mapping lookup and the fresh-insert branch of `persistObservation`, so repeat Epic syncs of the same single-result lab attach a new `lab_results` row to the existing `lab_panels` row rather than minting a duplicate panel each time. Mirrors the vital fingerprint pattern from #1191 and emits the standard `epic_sync_dedup_match` audit row on a hit.

### Strengths
- Pattern alignment is excellent — same `FingerprintHit` shape, same `logDedupMatch` helper, same `DEDUP_MATCH_ACTION` constant. Nothing bespoke about the audit-log handling.
- Inline `// #1202` comment on the helper calls out the panel_name=test_name collapse risk, which is the exact concern the PR brief highlighted.
- Tests cover the three meaningful cases: panel hit, no hit, and the single-result-as-panel collapse. Three inserts are asserted (lab_results, mapping, audit) — no off-by-one risk in the mock counts.
- The deferral comment removed from #1197's persistence.ts is gone cleanly; nothing dangling.
- Typecheck and CI all green.

### Issues Found

#### Critical (Must Fix)

| `services/epic-connector/src/sync/persistence.ts` | 887, 935 | **Mapping `internal_record_id` is inconsistent between the new dedup-fallback branch and the existing new-panel branch.** The new-panel branch (line 935-941) stores `resultId` (the `lab_results.id`) as the mapping's internalId, but the dedup-fallback branch (line 887-893) stores `panelFingerprintHit.internalId` (the `lab_panels.id`). On the very next Epic re-sync of an Observation that took the dedup path, `findMapping` returns the panel id, then the leaf-update branch (line 839-862) runs `db.update(labResults).where(eq(labResults.id, mapping.internalId))` — that matches **zero rows** (the leaf's id is the per-result UUID, not the panel id). The Epic-side value silently never refreshes, while `persistObservation` still returns `{ inserted: false }`. | Pick one identity convention and use it everywhere in the lab branch. Cleanest: in the fingerprint-hit branch, `upsertMapping` with `internalId: resultId` (the new `lab_results.id` you just inserted), and return `internalId: resultId`. That makes round-2 mapping hits land on the leaf-update path correctly. The dedup audit row can still reference the panel id in the `details` payload if you want the merge target preserved for ops. |

| `services/epic-connector/src/sync/persistence.ts` | 379-388 | The `collectedAt: null` branch widens the fingerprint to `(patient_id, panel_name)`, which on lab data is dimensionally weak — a stale Hemoglobin from 2020 could swallow a 2026 Hemoglobin. Today this path is unreachable because `mapFhirObservationToLabResultRow` returns null when `effectiveDateTime` is missing (observation-mapper.ts:370), so `findLabPanelByFingerprint` never sees a null `collectedAt` via `persistObservation`. | Either drop the null branch (and let the caller fall through to insert on the rare null case) or add an explicit comment that this widening is defensive-only and the caller currently guarantees non-null. The medication helper has the same pattern but a stronger key (rxnorm code) — the analogy is weaker here. |
| `services/epic-connector/src/sync/persistence.ts` | 873-910 | When the panel is reused, `lab_panels.reported_at` and `lab_panels.source_system` keep whatever the existing row had — so a CareBridge-originated panel reused by an Epic sync will still report `source_system != epic`. That's almost certainly the right call (preserves provenance) but it's not asserted or commented anywhere. | One-liner comment near the fingerprint-hit branch: "we attach the new result to the existing panel without mutating panel.source_system — the panel keeps its origin tag and the audit row records the Epic merge." |
| `services/epic-connector/src/__tests__/persistence.test.ts` | 469-598 | No test for the round-2 mapping-hit case described in the critical issue above — i.e. after a fingerprint dedup, what happens when the same Epic Observation is re-synced and `findMapping` returns a hit. | Add a test that primes a `findMapping` hit returning the panel id, then asserts a `lab_results` update by id actually executes (or, after the fix, that the result-id round-trip is preserved). |

#### Nitpicks (Optional)
- Test `dedups single-result Observations using test_name as panel_name` (line 583-600) uses the same `RESOURCE` fixture as the first test, so the assertion that "single-result deduping works" is structurally identical to the first case rather than exercising the *two-distinct-tests-collapsing-to-one-panel* scenario the title implies. Consider either renaming the test for clarity or adding a second different-test-name resource to actually stress the collapse risk.

(See "Issues" column in summary table.)

### Architecture Notes
This is a textbook addition to the cross-source dedup framework set up in #1191 / #1197 — same helper shape, same audit semantics, same predicate-builder pattern with nullable-dimension handling. The wiring sits in the correct place in `persistObservation` (after mapping lookup, before fresh-insert). The mapping-id inconsistency is the only structural concern.

### Verdict
- [x] Request Changes — the mapping-id inconsistency causes silent update misses on round-2 syncs of dedup-merged labs. Once fixed (and ideally tested), this is ready to merge.

blamechris:
### Deferred Items (Follow-Up Issues) — links

| Drop or document defensive null-collected_at branch | #1212 | Path is unreachable from current caller; safe to defer |
| Comment lab_panels.source_system provenance on fingerprint merge | #1213 | Pure documentation/clarity, no behaviour change |
| Strengthen single-result test to actually exercise distinct-test collapse | #1214 | Existing test still covers the happy path; strengthens, not unblocks |

blamechris:
Critical mapping-id inconsistency fixed in a7d136a.

**Root cause confirmed exactly as called out**: the fingerprint-hit branch in `persistObservation`'s lab path stored `panelFingerprintHit.internalId` (the reused `lab_panels.id`) as the mapping's `internal_record_id`, while inserting a fresh `lab_results` row under a new UUID. Round-2 syncs of the same Epic Observation hit the mapping, then the leaf-update branch ran `update(labResults).where(labResults.id = mapping.internalId)` — that WHERE never matched a row, the SET was a silent no-op, and the function still returned `inserted: false`. Epic-side result changes never landed.

**Fix**: in the fingerprint-hit branch, mint the `resultId` for the inserted `lab_results` row (already done) and pass it to `upsertMapping` and `logDedupMatch`, then return `{ internalId: resultId, kind: "lab", inserted: false }`. The leaf id is now stable across both the panel-reuse and panel-create paths — round-2 mapping hits land on the correct `lab_results` row and the UPDATE takes.

**Tests**:
- Added `round-2 lab Observation sync via dedup path updates the leaf row (#1207)`: pins the full round-trip — round 1 fingerprint hit inserts a new `lab_results` row and stores its id in the mapping; round 2 with the same Epic Observation id hits the mapping, runs UPDATE against `labResults`, and the modified value reaches the SET clause. This was the missing test called out in the Suggestions section.
- Updated the two existing lab dedup test assertions that previously pinned the buggy contract (return value and audit `resource_id` were asserted as the panel id; now asserted as the new `lab_results.id`).
- All three failed pre-fix with `expected '<panel-id>' to be '<uuid>'` — the exact symptom the review predicted.

Test count: 287 -> 288 (one net new round-2 test; the other two assertions are updates to existing cases). `pnpm typecheck` and `pnpm lint` both pass.

Suggestions #2 (defensive null-collected_at comment) and the nitpick on the duplicate test fixture are not addressed in this commit — happy to fold in or defer to a follow-up depending on review preference.

PR #1208, opened 2026-05-26
feat(epic): source_system guard on fingerprint-merge path (#1200)

### LINKED ISSUE #1200: feat(epic): source_system guard on fingerprint-merge path (#1197 follow-up)
Follow-up to #1197. The fingerprint-merge path in `services/epic-connector/src/sync/persistence.ts` unconditionally UPDATEs CareBridge-originated rows with Epic-payload values when `findByFingerprint` hits, without the `source_system != \"epic\"` guard that the Epic-id mapping path has.

## Why it matters

A clinician edits an encounter (location, reason, end_time) directly in CareBridge → Epic syncs the same logical encounter → the fingerprint match overwrites the clinician's edits with Epic's values. The conflict-check guard on the Epic-id mapping path (`persistence.ts` lines 413-430 / 561-578 / etc.) exists precisely to prevent this; the fingerprint path was added later in #1197 without an equivalent guard.

The header doc (lines 21-25) currently promises \"Epic does not get to overwrite CareBridge-originated data through the Epic-id path; we keep CareBridge's row and log the attempted overwrite for manual review\" — which is true for the Epic-id path but not for the fingerprint path.

## Options

1. Add the source_system check on the fingerprint path: when the matched row has `source_system != \"epic\"` (and != null), skip the UPDATE, write the `fhir_resources` mapping row only, and log to `audit_log` as a source-system conflict.
2. Keep the merge but log it explicitly — the current `epic_sync_dedup_match` audit row would need a `fields_overwritten` field listing what changed, so a HIPAA reviewer can reconstruct the clinician's original values from the audit trail.
3. Field-level last-writer policy: keep CareBridge-set fields, only fill in NULLs from Epic.

(1) is the safest interpretation of \"Epic doesn't overwrite CareBridge\". (3) is closer to ideal but adds complexity.

## Acceptance criteria

- [ ] When fingerprint matches a row with `source_system != \"epic\"` (or NULL on tables with no column), the writer either (a) skips the UPDATE + maps + logs, or (b) records the overwritten fields in the audit entry.
- [ ] Test covers the case: pre-existing CareBridge encounter with non-Epic source → Epic sync of same fingerprint → CareBridge row's location/reason are NOT silently lost.

## References

- #1197 (this PR)
- #1191 (parent dedup issue)
- #1190 (source_system columns — would tighten the type assertion once landed)

blamechris:
## Summary
- Adds a `source_system != 'epic'` guard on the fingerprint-merge path in `services/epic-connector/src/sync/persistence.ts`
- When the fingerprint hit is a CareBridge-originated row, the writer skips the UPDATE, writes the fhir_resources mapping, and emits an `epic_sync_source_conflict` audit entry
- Closes the gap between the Epic-id mapping path (already guarded) and the fingerprint path (added in #1197 without an equivalent guard)
- Updates the persistence.ts header doc to cover both paths

## Why this matters (clinical safety)
A clinician edits a CareBridge encounter (location, reason, end_time) -> Epic later syncs the same logical encounter -> fingerprint would match and overwrite the clinician's edits. This PR prevents that.

## Test plan
- [x] Pre-existing internal encounter + Epic fingerprint match -> fields preserved, audit row written
- [x] Pre-existing internal diagnosis + Epic fingerprint match -> fields preserved
- [x] Pre-existing internal allergy + Epic fingerprint match -> fields preserved
- [x] Pre-existing internal medication + Epic fingerprint match -> fields preserved
- [x] Pre-existing internal vital + Epic fingerprint match -> fields preserved
- [x] Pre-existing epic-sourced row + Epic fingerprint match -> existing dedup-match behavior intact
- [x] source_system=null treated as non-epic (defensive) -> fields preserved
- [x] Clinical-safety scenario from issue: location='ER Bay 4', reason='chest pain' survives Epic sync with location='ICU', reason='unstable'

Closes #1200
Follows: #1197

### DIFF
--- a/services/epic-connector/src/__tests__/persistence-source-guard.test.ts
+++ b/services/epic-connector/src/__tests__/persistence-source-guard.test.ts
@@ -1,0 +1,573 @@
+/**
+ * Source-system guard on the fingerprint-merge path (#1200).
+ *
+ * #1197 added cross-source fingerprint dedup that UPDATEs the matched
+ * row with the Epic payload regardless of who originally created it.
+ * The Epic-id mapping path has guarded against this since #1183 ("Epic
+ * does not get to overwrite CareBridge-originated data"). This test
+ * suite pins the equivalent guard for the fingerprint path:
+ *
+ *  - When findByFingerprint hits a row whose source_system != 'epic'
+ *    (or null on a table that has the column — treated as non-epic):
+ *      • SKIP the UPDATE of clinical fields
+ *      • Still write the fhir_resources mapping (Epic id round-trips
+ *        to the same internal row)
+ *      • Write an audit_log row with action='epic_sync_source_conflict'
+ *        carrying the matched row id, the Epic FHIR id, and the
+ *        source_system value of the matched row.
+ *  - When the matched row has source_system='epic' (or table lacks the
+ *    column and we treat absence as 'epic-equivalent' per resource —
+ *    medications/vitals where the column was added long ago and pre-
+ *    column rows are effectively internal — we still treat null as
+ *    non-epic to be safe): existing dedup-match behaviour preserved.
+ *
+ * Uses the shared createMockDb harness — same pattern as the cycle-5
+ * tests in persistence.test.ts.
+ */
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { createMockDb, type MockDb } from "@carebridge/test-utils";
+
+let db: MockDb;
+
+vi.mock("@carebridge/db-schema", () => ({
+  getDb: () => db,
+  patients: {
+    id: "id",
+    mrn_hmac: "mrn_hmac",
+    patient_id: "patient_id",
+  },
+  medications: {
+    id: "id",
+    patient_id: "patient_id",
+    rxnorm_code: "rxnorm_code",
+    started_at: "started_at",
+  },
+  vitals: {
+    id: "id",
+    patient_id: "patient_id",
+    loinc_code: "loinc_code",
+    recorded_at: "recorded_at",
+  },
+  labPanels: {},
+  labResults: {},
+  diagnoses: {
+    id: "id",
+    patient_id: "patient_id",
+    icd10_code: "icd10_code",
+    snomed_code: "snomed_code",
+    onset_date: "onset_date",
+  },
+  allergies: {
+    id: "id",
+    patient_id: "patient_id",
+    rxnorm_code: "rxnorm_code",
+    snomed_code: "snomed_code",
+  },
+  encounters: {
+    id: "id",
+    patient_id: "patient_id",
+    start_time: "start_time",
+    encounter_type: "encounter_type",
+  },
+  fhirResources: {
+    id: "id",
+    resource_type: "resource_type",
+    resource_id: "resource_id",
+    internal_record_id: "internal_record_id",
+  },
+  auditLog: {},
+  hmacForIndex: (value: string) => `hmac:${value}`,
+}));
+
+const {
+  persistMedicationRequest,
+  persistObservation,
+  persistCondition,
+  persistAllergy,
+  persistEncounter,
+} = await import("../sync/persistence.js");
+
+const PATIENT_ID = "11111111-1111-1111-1111-111111111111";
+
+beforeEach(() => {
+  vi.clearAllMocks();
+  db = createMockDb();
+});
+
+// ── helpers ─────────────────────────────────────────────────────
+
+function findAuditInsert(
+  action: string,
+): Record<string, unknown> | undefined {
+  const hit = db.insert.calls.find((c) => {
+    const v = c.chainArgs[0]?.[0] as Record<string, unknown> | undefined;
+    return v?.action === action;
+  });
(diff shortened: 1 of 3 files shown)


blamechris:
## Code Review: PR #1208 — source_system guard on fingerprint-merge path

### Summary
Closes the protection gap between the Epic-id mapping path (guarded since #1183) and the fingerprint path (added in #1197 without a guard). For Encounter, Condition, AllergyIntolerance, MedicationRequest, and Observation/vital, a fingerprint hit on a row whose `source_system != "epic"` now: (a) skips the UPDATE, (b) writes the `fhir_resources` mapping so the Epic FHIR id still round-trips to the same internal row, and (c) emits `audit_log` row with `action="epic_sync_source_conflict"`. The matched-`epic` path keeps its existing `epic_sync_dedup_match` semantics.

### Strengths
- **Per-resource symmetry is uniform.** All five guarded paths follow the same shape: `isEpicOwned()` check → conditional mapping-only-write + audit + early-return → otherwise UPDATE + mapping + dedup-audit. No copy-paste drift between resource types.
- **Mapping IS still written on conflict.** This is the right call — without it, the next Epic sync of the same FHIR id would re-fingerprint-match and re-trip the guard, and the Epic↔CareBridge id link would never be persisted. The audit row also carries `internal_record_id` so a reviewer can correlate.
- **Audit payload completeness.** `details` JSON carries `reason`, `attempted_source`, `existing_source`, `epic_fhir_id`, `fingerprint` (the dimensions used), and `resolution="keep_carebridge_value"`. Distinct `action` values (`epic_sync_source_conflict` vs `epic_sync_dedup_match` vs Epic-id-path `action="update", success=false`) keep the three failure/merge modes queryable.
- **Null treated as non-epic.** Defensive policy is the safer interpretation; pre-#1190 rows that bypassed the default still get protected.
- **Test rename is well-justified.** The 5 existing dedup tests in `persistence.test.ts` now tag their fixture rows with `source_system: "epic"` (otherwise the new guard would flip them to `conflict`). Each retitle has an inline note pointing at `persistence-source-guard.test.ts` for the CareBridge-protection variant. The 6th dedup test (`persistPatient` MRN) is correctly NOT renamed — `patients` has no `source_system` column and `persistPatient` doesn't invoke the guard.
- **Clinical-safety scenario is end-to-end.** The `location='ER Bay 4', reason='chest pain'` survives-Epic-sync test in `persistence-source-guard.test.ts` is the canonical scenario from the issue and asserts: result shape, `db.update` never called, mapping written, audit emitted with the right provenance.

### Issues Found

#### Critical (Must Fix)
None.

| `services/epic-connector/src/sync/persistence.ts` | 1–37 (header doc) | Header doc says "Refuses to overwrite CareBridge-originated data on BOTH the Epic-id path AND the fingerprint path." Patient is silently exempt (no `source_system` column on `patients`; `persistPatient` fingerprint hit merges unconditionally). Either narrow the claim ("for resources whose table carries `source_system`: Encounter, Condition, AllergyIntolerance, MedicationRequest, Observation") or add a one-line "Patient identity is treated as Epic-owned at first import; see `findPatientByFingerprint`" caveat. | The header is the contract for downstream reviewers / next PR authors. Today the claim slightly over-reaches the implementation. |
| `services/epic-connector/src/sync/persistence.ts` | 311–314, 343–346, 367–370 | The `(hit as { source_system?: string \| null }).source_system ?? null` casts in `findEncounterByFingerprint`, `findDiagnosisByFingerprint`, `findAllergyByFingerprint` were defensively-typed when those tables had no column. Migration 0054 (#1190) now makes `source_system` NOT NULL on all three, so `hit.source_system` is a direct property read like medications/vitals. Cast can be dropped in a follow-up. Comment on line 308-309 ("encounters has no source_system column today (#1190 may add one)") is also stale. | Pure cleanup — no behaviour change. Worth doing once after a soak window so 0054 has time to settle. |

#### Nitpicks (Optional)
- `SOURCE_CONFLICT_ACTION` and `DEDUP_MATCH_ACTION` are file-local `const`s. If stacked PRs need to reference them by name, they will need the `export` keyword added. Currently the test suite uses the raw string literals so there is no consumer today — no need to pre-emptively export.

| Header doc clarifies Patient exemption | https://github.com/blamechris/carebridge/issues/1215 | Doc-only; not safety-critical and the inline `findPatientByFingerprint` comment already explains the semantics. |
| Drop stale `(hit as {...}).source_system` casts post-#1190 | https://github.com/blamechris/carebridge/issues/1216 | Pure cleanup; runs after #1190's migration 0054 has soaked. |

### Architecture Notes
- Persistence layer remains the right home — the guard lives at the write boundary where `findByFingerprint` results meet the UPDATE, not in the converter or sync-worker layers.
- BullMQ retry-safety preserved: `epic_sync_source_conflict` audit rows are append-only by design (migration 0012); re-running the same Epic FHIR resource emits another conflict row on each retry, but the underlying clinical row is never touched. Audit-log growth on persistent conflicts is bounded by Epic's retry policy — acceptable.
- No new tRPC procedures; no new BullMQ events; no schema change; no `@carebridge/validators` impact. Surface area is contained to one service.

### Verdict
**Approve** — Ready to merge. The clinical-safety contract is correct and tested end-to-end; deferred items are doc + dead-code cleanup, not gating.

PR #1209, opened 2026-05-26
feat(epic): capture pre-update field values in allergy dedup audit (#1203)

### LINKED ISSUE #1203: feat(epic): allergy fingerprint temporal dimension (#1197 follow-up)
Follow-up to #1197. The allergy fingerprint in `services/epic-connector/src/sync/persistence.ts` (`findAllergyByFingerprint`, line 291) uses only `(patient_id, rxnorm_code | snomed_code)` — no temporal dimension. This is acknowledged in the PR body and the inline comment (line 1054-1056):

> No timestamp dimension on allergies (allergens are typically open-ended), and the free-text allergen column is encrypted, so we rely on the coded substance for a safe match.

## Risk

Two re-evaluated allergy records for the same SNOMED/RxNorm code merge. Example: patient flagged penicillin allergy in 2020 (severity: moderate, verification_status: unconfirmed) → re-evaluated and re-confirmed in 2024 (severity: severe, verification_status: confirmed). Today both rows would be the same fingerprint and Epic's payload would silently overwrite the older record's severity / verification_status fields, losing the clinical history.

For allergies this is often correct (same substance = same allergy clinically) but the silent overwrite of severity / reaction / verification_status fields without an audit-readable record of the prior values is the data-integrity gap.

## Options

1. Document the merge behaviour explicitly in code + accept the trade-off.
2. Add `recorded_date` or `verification_status_changed_at` column to allergies and fingerprint by it.
3. Pair with #1200 (source_system guard): when fingerprint hits a CareBridge-originated allergy with non-trivial fields (severity, reaction), skip the field-level overwrite and only insert the mapping.
4. Capture the overwritten fields in the `epic_sync_dedup_match` audit entry payload so prior values are recoverable.

(4) is the lightest change and addresses the data-loss concern.

## Acceptance criteria

- [ ] Decide: accept-and-document, add-column, or capture-overwritten-fields
- [ ] If (4): the audit-log row carries the pre-update field values for any field Epic overwrote
- [ ] Test covers: pre-existing allergy with severity=\"severe\" + Epic sync with severity=\"moderate\" → audit row contains the original \"severe\" value

## References

- #1197 (this PR)
- #1191 (parent)
- #1200 (source_system guard follow-up — overlaps)

blamechris:
## Summary
- Extends `persistAllergy`'s dedup-merge audit payload (in `services/epic-connector/src/sync/persistence.ts`) with `overwritten_fields: { fieldName: priorValue, ... }` capturing the pre-update value of every field the Epic merge changes
- Only populated when at least one field actually changed (no audit weight on no-op merges)
- Preserves clinical history for re-evaluated allergies whose substance code is identical but whose severity/verification_status/reaction were updated by a clinician downstream

## Stacked on #1208
This PR branches from `feat/1200-source-system-guard`. Retarget to main after #1208 merges. #1200's source_system guard handles the case where the matched allergy is CareBridge-originated; this PR handles the Epic-on-Epic re-evaluation case.

## Fields captured
The snapshot mirrors exactly the columns `persistAllergy`'s UPDATE rewrites today: `allergen`, `severity`, `reaction`. Widening the UPDATE in a follow-up automatically widens the snapshot via the same shared diff helper (`computeAllergyOverwrittenFields`). `verification_status` is on the row but not in the current UPDATE set, so it isn't captured here — a follow-up that adds verification_status to the merge surface should add it to PriorAllergyRow + the diff helper together.

## Test plan
- [x] Pre-existing severity='severe' + inbound severity='moderate' → audit payload.overwritten_fields = { severity: 'severe' }
- [x] All three fields change → all three captured (allergen, severity, reaction)
- [x] No-op merge (identical inbound) → no overwritten_fields key
- [x] Partial overwrite (only severity changes) → only severity captured
- [x] Existing #1197 / #1208 allergy tests still pass (30/30 in persistence + persistence-source-guard + persistence-allergy-audit)

Depends on #1208
Closes #1203

### DIFF
--- a/services/epic-connector/src/__tests__/persistence-allergy-audit.test.ts
+++ b/services/epic-connector/src/__tests__/persistence-allergy-audit.test.ts
@@ -1,0 +1,315 @@
+/**
+ * Pre-update field snapshot in the allergy dedup-merge audit (#1203).
+ *
+ * `findAllergyByFingerprint` keys on (patient_id, rxnorm_code | snomed_code)
+ * with no temporal dimension — a re-evaluated allergy (same SNOMED/RxNorm,
+ * updated severity/reaction) merges silently. #1200's source_system guard
+ * prevents Epic from overwriting CareBridge-originated rows, but when the
+ * matched row IS Epic-sourced the merge proceeds and the clinician-edited
+ * prior values are lost without provenance.
+ *
+ * #1203 (this file) pins the pre-update snapshot: the `epic_sync_dedup_match`
+ * audit row's payload now carries
+ *   { overwritten_fields: { severity: priorValue, reaction: priorValue, ... } }
+ * for every field whose value actually changed on the UPDATE. No-op merges
+ * write no `overwritten_fields` key — the audit weight is only earned when
+ * a clinically meaningful field flipped.
+ *
+ * The set of fields captured matches what `persistAllergy`'s UPDATE writes
+ * (allergen, severity, reaction). Widening the UPDATE in a follow-up
+ * automatically widens the snapshot via the same shared computation.
+ */
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { createMockDb, type MockDb } from "@carebridge/test-utils";
+import type { FhirResource } from "../fhir-types.js";
+
+let db: MockDb;
+
+vi.mock("@carebridge/db-schema", () => ({
+  getDb: () => db,
+  patients: {
+    id: "id",
+    mrn_hmac: "mrn_hmac",
+    patient_id: "patient_id",
+  },
+  medications: {
+    id: "id",
+    patient_id: "patient_id",
+    rxnorm_code: "rxnorm_code",
+    started_at: "started_at",
+  },
+  vitals: {
+    id: "id",
+    patient_id: "patient_id",
+    loinc_code: "loinc_code",
+    recorded_at: "recorded_at",
+  },
+  labPanels: {},
+  labResults: {},
+  diagnoses: {
+    id: "id",
+    patient_id: "patient_id",
+    icd10_code: "icd10_code",
+    snomed_code: "snomed_code",
+    onset_date: "onset_date",
+  },
+  allergies: {
+    id: "id",
+    patient_id: "patient_id",
+    rxnorm_code: "rxnorm_code",
+    snomed_code: "snomed_code",
+  },
+  encounters: {
+    id: "id",
+    patient_id: "patient_id",
+    start_time: "start_time",
+    encounter_type: "encounter_type",
+  },
+  fhirResources: {
+    id: "id",
+    resource_type: "resource_type",
+    resource_id: "resource_id",
+    internal_record_id: "internal_record_id",
+  },
+  auditLog: {},
+  hmacForIndex: (value: string) => `hmac:${value}`,
+}));
+
+const { persistAllergy } = await import("../sync/persistence.js");
+
+const PATIENT_ID = "11111111-1111-1111-1111-111111111111";
+
+beforeEach(() => {
+  vi.clearAllMocks();
+  db = createMockDb();
+});
+
+function findAuditInsert(
+  action: string,
+): Record<string, unknown> | undefined {
+  const hit = db.insert.calls.find((c) => {
+    const v = c.chainArgs[0]?.[0] as Record<string, unknown> | undefined;
+    return v?.action === action;
+  });
+  return hit?.chainArgs[0]?.[0] as Record<string, unknown> | undefined;
+}
+
+/**
+ * Build a FHIR AllergyIntolerance resource with caller-controlled
+ * severity + reaction so we can drive the merge with explicit "new"
+ * values. SNOMED 373270004 = Penicillin (same fingerprint as the
+ * existing #1191/#1200 allergy tests).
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #1209

### Summary

Extends the `epic_sync_dedup_match` audit payload with a `overwritten_fields: { fieldName: priorValue, ... }` snapshot whenever a fingerprint-matched Epic allergy merge actually rewrites a clinically meaningful field. The snapshot is captured in `findAllergyByFingerprint`, diffed by a new `computeAllergyOverwrittenFields` helper, and threaded through `logDedupMatch` via a widened signature. No-op merges (identical inbound values) omit the key entirely so the audit weight only lands when a real change happened. The change is HIPAA-relevant: a reviewer can now reconstruct the clinician's pre-merge state from the audit row.

### Strengths

- **Defensive null handling.** `findAllergyByFingerprint` coerces missing columns via `?? null`, and `computeAllergyOverwrittenFields` works on `string | null` everywhere it needs to. The strict-equality diff is correct for plaintext columns.
- **Audit payload shape is preserved.** All existing keys (`reason`, `attempted_source`, `epic_fhir_id`, `fingerprint`, `resolution`) are still emitted unchanged; `overwritten_fields` is purely additive. HIPAA reviewers won't see a contract break.
- **No-op merges stay clean.** `logDedupMatch` filters via `Object.keys(args.overwrittenFields).length > 0` so the audit payload doesn't carry an empty object on a no-op merge.
- **Helper scope matches UPDATE scope.** `PriorAllergyRow` carries exactly the three columns the UPDATE writes — no incidental snapshotting of unrelated columns on every fingerprint lookup.
- **Documentation pins the invariant.** Both `PriorAllergyRow` and `computeAllergyOverwrittenFields` docstrings call out the single-source-of-truth rule: widen the UPDATE -> widen `PriorAllergyRow` + helper together. Cuts the future-drift risk.
- **`priorAllergyRow` is optional on `FingerprintHit`.** Other resource finders (encounter, medication, diagnosis) aren't forced to populate it; allergy-specific scope is respected.
- **Test coverage is thorough.** 5 tests cover severity-only flip, all-three-fields change, no-op, partial overwrite, and the issue's acceptance scenario (with verification_status explicitly documented as out-of-scope). All 30 epic-connector persistence tests pass locally; typecheck clean.

### Issues Found

#### Critical (Must Fix)

None.

| `services/epic-connector/src/sync/persistence.ts` | 438 | Loosen `incoming.allergen` from `string` to `string \| null` in `computeAllergyOverwrittenFields`. | `PriorAllergyRow.allergen` is already `string \| null`. Matching both sides keeps the diff helper safe if `MappedAllergyRow.allergen` ever loosens (today the upstream type is `string`, but the helper would be a footgun if that changes). Purely a future-proofing tightening. |
| `services/epic-connector/src/__tests__/persistence-allergy-audit.test.ts` | tests 2-5 | Optionally add one assertion that the legacy payload keys (`reason`, `epic_fhir_id`, `fingerprint`, `resolution`, `attempted_source`) are still present alongside `overwritten_fields`. | Test 1 covers this for the severity case; the other four only assert `overwritten_fields`. The existing #1191 tests likely already cover the shape, but a single line in this file would lock the contract against accidental regression. Not blocking. |

#### Nitpicks (Optional)

- `priorAllergyRow` sits on the shared `FingerprintHit` interface but is only set by the allergy finder. Pure tidiness — a per-resource refinement (`AllergyFingerprintHit extends FingerprintHit`) would be cleaner, but the current shape works fine and matches the established style in the file.
- The deferred-verification-status follow-up is explicitly called out in the PR body; tracked as a follow-up issue below so it doesn't get lost.

| Widen the dedup-merge UPDATE + `PriorAllergyRow` + helper to include `verification_status`. | [#1220](https://github.com/blamechris/carebridge/issues/1220) | Out of scope per the PR body; widening the UPDATE surface is a separate change with its own clinical-data implications. |

### Architecture Notes

- Cleanly slots into the existing dedup-merge path established by #1191 (cross-source dedup) and #1200 (source_system guard). #1200 short-circuits the merge when the matched row is CareBridge-originated; this PR handles the Epic-on-Epic re-evaluation case that #1200 explicitly leaves through.
- Audit row uses the file-local `DEDUP_MATCH_ACTION` const introduced in the stack; no constant duplication.
- No PHI leak risk in the audit payload: `details` is encrypted at rest the same way the existing `attempted_source` / `epic_fhir_id` fields are.
- Closes #1203 (allergy fingerprint temporal dimension).

### Verdict

**Approve** — Ready to merge once #1208 (the stacked base) lands and this PR retargets to main. No critical findings; the two suggestions are nice-to-haves, not blockers. Follow-up issue #1220 tracks the deferred `verification_status` widening.

blamechris:
## Review Comments Addressed

No inline review comments to process — the agent-review on this PR posted an issue-level comment (verdict: Approve) with two non-blocking suggestions and one follow-up issue ([#1220](https://github.com/blamechris/carebridge/issues/1220) for the deferred `verification_status` widening).

**Total:** 0 inline comments addressed
- Fixed: 0
- False positives: 0
- Follow-up issues created: 1 (#1220 — already linked in the review)
- Existing issues closed: 0

PR is ready for merge once #1208 (stacked base) lands and this PR retargets to main.

Issue #391, opened 2026-04-13
feat(fhir): implement Epic data sync worker with AI oversight integration


blamechris:
## Context
This is the capstone feature: automatically pull patient data from Epic, store it in CareBridge, and feed changes into the AI oversight engine. This is CareBridge differentiator — Epic data flows in, AI catches cross-specialty gaps that Epic alone misses.

#### Sync Orchestration
- New BullMQ queue: "epic-sync"
- Job types: full-sync (initial patient import), incremental-sync (delta updates), single-resource-sync
- Configurable sync interval per patient (default: every 15 minutes for active patients)
- Use Epic FHIR client (#390) for data retrieval

#### Clinical Event Emission
- After importing/updating any clinical data from Epic, emit to the existing "clinical-events" BullMQ queue
- Event types: vitals.recorded, lab.resulted, medication.prescribed, diagnosis.added, encounter.created
- This triggers the existing AI oversight worker (deterministic rules + LLM review)
- Include `source_system: "epic"` in event metadata

#### Conflict Resolution
- When Epic data conflicts with existing CareBridge data (e.g., medication status changed):
  - Epic is source of truth for Epic-originated data
  - CareBridge is source of truth for CareBridge-originated data
  - Conflicts logged to audit_log with details for manual review

#### Sync Dashboard Data
- Track sync status per patient: last_synced_at, sync_errors, resources_synced_count
- Expose via tRPC query for clinician portal (read-only)

### tRPC Router
- `epicSync.triggerSync` — admin mutation to trigger manual sync for a patient
- `epicSync.getSyncStatus` — query sync state for a patient
- `epicSync.listSyncErrors` — query recent sync failures

## Acceptance criteria
- Full and incremental sync working against Epic sandbox
- Clinical events emitted for each imported/updated resource
- AI oversight worker processes Epic-sourced events (verified end-to-end)
- Conflict resolution logged
- Sync status queryable via tRPC
- Audit trail for all Epic data imports

## Dependencies
- #390 (Epic FHIR client)
- #389 (SMART on FHIR auth)

blamechris:
Closing — the Epic data sync worker is shipped. Audited against this issue's acceptance criteria:

**Shipped**:
- BullMQ sync worker: \`services/epic-connector/src/workers/sync-worker.ts\`
[...]

PR #1210, opened 2026-05-26
feat(epic): detect Epic-id conflict on fingerprint merge (#1201)

### LINKED ISSUE #1201: feat(epic): detect Epic-id conflict on fingerprint merge (#1197 follow-up)
Follow-up to #1197.

When the fingerprint match in `services/epic-connector/src/sync/persistence.ts` lands on an internal row that is **already mapped** to a different Epic FHIR id (via a pre-existing `fhir_resources` row with a different `resource_id`), `upsertMapping` writes a second mapping row pointing at the same internal id. The two Epic FHIR ids now both round-trip to the same CareBridge record.

This may be correct (Epic re-issued the resource ID for the same logical entity) or it may collapse two genuinely distinct entities. Today the code can't tell.

## What the original brief said

> When fingerprint matches an Epic-originated row that's mapped to a DIFFERENT epic_id → this is a true conflict (e.g., Epic re-issued a different ID for the same logical encounter). The issue's policy is \"log to audit_log + create a new row\".

Today: dedup-merges + logs (no new-row option). Worth detecting the case explicitly.

## Suggested approach

`findByFingerprint` returns the matched row id. Before treating it as a dedup target, check `fhir_resources` for any existing mapping whose `internal_record_id` = the matched id and `resource_id` != the inbound Epic id:

```ts
const conflictingMapping = await db
  .select()
  .from(fhirResources)
  .where(and(
    eq(fhirResources.resource_type, resourceType),
    eq(fhirResources.internal_record_id, hit.internalId),
    ne(fhirResources.resource_id, fhirId),
  ))
  .limit(1);

if (conflictingMapping.length > 0) {
  // True ambiguity. Either insert a new row + log, or skip + log + bail.
}
```

## Acceptance criteria

- [ ] Detect when a fingerprint hit lands on an internal row already mapped to a different Epic id
- [ ] Policy decision: insert a new row OR refuse the merge (issue suggests \"insert + log\")
- [ ] Audit-log entry distinguishes \"epic_sync_dedup_match\" (clean merge) from \"epic_sync_dedup_conflict\" (ambiguous)
- [ ] Test fixture: pre-existing Epic-mapped encounter + Epic sync of same fingerprint with different FHIR id → triggers the conflict branch

## References

- #1197 (this PR)
- #1191 (parent)

blamechris:
## Summary
- Adds `findConflictingMapping(db, resourceType, internalId, fhirId)` helper in `services/epic-connector/src/sync/persistence.ts`
- Inserts a new check stage between fingerprint lookup and the source-system guard: when the matched internal row already has a `fhir_resources` mapping pointing at a different Epic FHIR id, skip the merge, fall through to INSERT, and emit `epic_sync_dedup_conflict` audit row
- New audit action constant `DEDUP_CONFLICT_ACTION = "epic_sync_dedup_conflict"` distinguishes ambiguous merges from clean dedup-matches (`epic_sync_dedup_match`) and from source-conflict skips (`epic_sync_source_conflict` from #1200)

## Stacked on #1208
This PR branches from `feat/1200-source-system-guard` and depends on the audit-action constant pattern landed there. Retarget to main after #1208 merges.

## Test plan
- [x] Pre-existing Epic-mapped encounter + same fingerprint, different Epic id -> new row inserted, conflict audit row written
- [x] Same case for diagnoses, medications
- [x] Pre-existing Epic-mapped row + same fingerprint, SAME Epic id -> unchanged dedup-match behavior

Depends on #1208
Closes #1201

### DIFF
--- a/services/epic-connector/src/__tests__/persistence-id-conflict.test.ts
+++ b/services/epic-connector/src/__tests__/persistence-id-conflict.test.ts
@@ -1,0 +1,438 @@
+/**
+ * Epic-id conflict detection on the fingerprint-merge path (#1201).
+ *
+ * Background: #1191 introduced fingerprint dedup so that a CareBridge-
+ * originated row created before Epic sync was wired doesn't get
+ * duplicated when Epic later syncs the same logical entity. #1200
+ * tightened that path so the dedup-merge UPDATE is skipped when the
+ * matched row is CareBridge-originated.
+ *
+ * But there's a *separate* failure mode: the fingerprint match lands on
+ * an internal row that ALREADY has an Epic FHIR id mapping pointing at
+ * it — and the inbound Epic FHIR id is *different*. That means either
+ * Epic has re-issued an ID for the same logical entity, or two
+ * genuinely distinct Epic entities collide on our fingerprint key. The
+ * old dedup-merge path silently wrote a SECOND `fhir_resources` row
+ * pointing at the same internal id, so two Epic FHIR ids both
+ * round-tripped to one CareBridge record with no signal to a reviewer.
+ *
+ * This suite pins the new behaviour: when the matched internal row
+ * already has a conflicting Epic mapping, the writer skips the merge,
+ * falls through to a plain INSERT (a new internal row + a new mapping
+ * for the inbound Epic id), and writes ONE audit row with
+ * action='epic_sync_dedup_conflict' carrying both Epic ids.
+ *
+ * Pipeline order, per the issue:
+ *   1. fingerprint lookup → hit?
+ *   2. (NEW) conflict check on fhir_resources → if conflict, fall
+ *      through to plain INSERT + audit `epic_sync_dedup_conflict`.
+ *   3. source_system guard (#1200).
+ *   4. dedup-merge.
+ *
+ * The conflict check is FIRST after the fingerprint hit because it's
+ * about identity ambiguity, not data ownership — once two Epic ids both
+ * point at one row we've lost the ability to safely round-trip either.
+ */
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { createMockDb, type MockDb } from "@carebridge/test-utils";
+
+let db: MockDb;
+
+vi.mock("@carebridge/db-schema", () => ({
+  getDb: () => db,
+  patients: {
+    id: "id",
+    mrn_hmac: "mrn_hmac",
+    patient_id: "patient_id",
+  },
+  medications: {
+    id: "id",
+    patient_id: "patient_id",
+    rxnorm_code: "rxnorm_code",
+    started_at: "started_at",
+  },
+  vitals: {
+    id: "id",
+    patient_id: "patient_id",
+    loinc_code: "loinc_code",
+    recorded_at: "recorded_at",
+  },
+  labPanels: {},
+  labResults: {},
+  diagnoses: {
+    id: "id",
+    patient_id: "patient_id",
+    icd10_code: "icd10_code",
+    snomed_code: "snomed_code",
+    onset_date: "onset_date",
+  },
+  allergies: {
+    id: "id",
+    patient_id: "patient_id",
+    rxnorm_code: "rxnorm_code",
+    snomed_code: "snomed_code",
+  },
+  encounters: {
+    id: "id",
+    patient_id: "patient_id",
+    start_time: "start_time",
+    encounter_type: "encounter_type",
+  },
+  fhirResources: {
+    id: "id",
+    resource_type: "resource_type",
+    resource_id: "resource_id",
+    internal_record_id: "internal_record_id",
+  },
+  auditLog: {},
+  hmacForIndex: (value: string) => `hmac:${value}`,
+}));
+
+const {
+  persistMedicationRequest,
+  persistCondition,
+  persistEncounter,
+} = await import("../sync/persistence.js");
+
+const PATIENT_ID = "11111111-1111-1111-1111-111111111111";
+
+beforeEach(() => {
+  vi.clearAllMocks();
+  db = createMockDb();
+});
+
+// ── helpers ─────────────────────────────────────────────────────
+
+function findAuditInsert(
+  action: string,
+): Record<string, unknown> | undefined {
+  const hit = db.insert.calls.find((c) => {
+    const v = c.chainArgs[0]?.[0] as Record<string, unknown> | undefined;
+    return v?.action === action;
+  });
+  return hit?.chainArgs[0]?.[0] as Record<string, unknown> | undefined;
+}
+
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #1210

### Summary
Adds a third stage to the persistence pipeline (#391, #1191, #1200, #1201): between fingerprint match and source_system guard, the writer now checks `fhir_resources` for a pre-existing Epic mapping on the matched internal id with a DIFFERENT inbound `resource_id`. On hit, it falls through to a fresh INSERT and emits an `epic_sync_dedup_conflict` audit row carrying both Epic IDs and the fingerprint. Implemented for Patient, Encounter, Condition, MedicationRequest, Observation, AllergyIntolerance. 5 new tests cover positive + same-id negative across Encounter, Condition, MedicationRequest. 23 of 24 epic-connector test files (301 tests) pass, typecheck clean.

### Strengths
- **Pipeline order is correct per the issue spec.** Conflict-check runs BEFORE the #1200 source_system guard in every writer — verified at persistence.ts:1575-1583 (Encounter), 632-639 (Patient), and all four other writers. Module-doc step 6 also calls this out explicitly.
- **`findConflictingMapping` query is exactly what the issue spec calls for**: `resource_type AND internal_record_id AND resource_id != fhirId`, with `LIMIT 1`. The `ne(...)` predicate ensures the same-id re-sync case is correctly NOT flagged.
- **Edge case "fingerprint hit IS the inbound Epic id's existing mapping" is correctly handled** — the `ne(resource_id, inboundFhirId)` filter excludes that mapping at SQL level, and the same-id negative tests (`epic-enc-A` / `epic-med-A`) pin the behavior.
- **Audit payload is complete and useful for manual review**: `new_epic_fhir_id`, `existing_epic_fhir_id`, `matched_internal_id`, `fingerprint`, `reason`, `resolution`. `success=false` flags the row for review. `resource_id` ties to the NEW internal id (with `matched_internal_id` in details) — defensible choice and the inline comment explains why.
- **No silent double-mapping.** The conflict path falls through to `db.insert(<resource_table>)` + `upsertMapping(...new id...)`, never calls `upsertMapping` on the matched id. Tests at lines 196-212 pin both: a mapping for `epic-enc-B → new id` exists, and no mapping for `epic-enc-B → existing id` exists.
- **Audit-action constant naming is consistent** with #1191 / #1200: `epic_sync_dedup_match` / `epic_sync_source_conflict` / `epic_sync_dedup_conflict` — three distinct ambiguity modes, three distinct actions.
- **Defensive coding** for unit-test mocks: `rows?.[0]` handles both empty arrays and under-queued mock returns of `undefined`. Inline comment explains why.
- **Module doc and the `findConflictingMapping`/`logDedupConflict` JSDoc both explain the WHY** (identity ambiguity vs data ownership), the pipeline ordering rationale, and the resolution policy. Good narrative trail for the next reader.

### Issues Found

#### Critical (Must Fix)
None.

| services/epic-connector/src/sync/persistence.ts | 1673, 698, etc. | Consider surfacing `conflict: \"epic_id_conflict\"` (or similar) on the returned object when the conflict path fires. | The #1200 path already returns `conflict: \"source_system_conflict\"` so callers can distinguish ambiguity from clean inserts. The #1201 path returns `{ inserted: true }` with no caller signal — only the audit row records it. If any downstream caller is metrics-gathering on conflict counts (or needs to alert on identity ambiguity), they would currently have to query audit_log. Not blocking — audit-only is a defensible policy — but worth tracking. |
| services/epic-connector/src/__tests__/persistence-id-conflict.test.ts | n/a | No test for Patient or AllergyIntolerance conflict paths. | The 5 tests cover Encounter (positive + negative), Condition (positive), MedicationRequest (positive + negative). Patient and AllergyIntolerance use the same `findConflictingMapping` + `logDedupConflict` helpers so the risk is low, but coverage is asymmetric. Patient is called out in the inline comment as \"high-stakes\" (Epic merge/unmerge cycles), so it's the one most worth pinning. |
| services/epic-connector/src/sync/persistence.ts | 285 | `LIMIT 1` is sufficient for the conflict detection, but loses information when MULTIPLE different Epic ids already map to the same internal id. | Today this is presumably impossible (mapping is keyed on `(resource_type, resource_id)`), but if it ever happens (eg. a prior bug, manual DB intervention), the audit row only surfaces one of N. A `LIMIT 5` + array in `details.existing_epic_fhir_ids` would future-proof. Nit; not blocking. |

#### Nitpicks (Optional)
- `ConflictingMapping.internalRecordId` is set but never consumed by callers — they already have `fingerprintHit.internalId`. Could be dropped from the interface to keep it minimal. (services/epic-connector/src/sync/persistence.ts:267-268, 294-295)
- The `if (conflictMapping && fingerprintHit)` guard at the end of each writer is technically redundant — `conflictMapping` can only be set inside `if (fingerprintHit)`. TypeScript narrowing would handle it without the explicit `&& fingerprintHit`, but the explicit check is more defensible for refactoring (and CI typecheck passes either way). Optional.
- Test file uses `findAllMappingInserts()` heuristic (`internal_record_id !== undefined || imported_at !== undefined`) to discriminate mapping rows from audit rows in the mock — fragile if a future audit schema gains `internal_record_id`. A `table === fhirResources` filter would be sturdier, but the mock infrastructure doesn't currently expose `table`. Not actionable today.

| Surface `conflict: \"epic_id_conflict\"` discriminator on writer return value | [#1217](https://github.com/blamechris/carebridge/issues/1217) | Symmetry with #1200's `conflict: \"source_system_conflict\"`. Audit-only is fine for the immediate ask. |
| Add Patient + AllergyIntolerance conflict-path test coverage | [#1218](https://github.com/blamechris/carebridge/issues/1218) | Helpers are shared, so the risk is low, but Patient identity ambiguity is high-stakes per the inline comment. |
| Future-proof `findConflictingMapping` to surface up to N existing Epic ids | [#1219](https://github.com/blamechris/carebridge/issues/1219) | Defensive — current schema makes multi-mapping impossible, but audit row only captures one row today. |

### Architecture Notes
- Pipeline ordering matches the issue spec (`fingerprint → conflict-check (NEW) → source_system guard → dedup-merge`), and the module-doc step 6 documents it clearly.
- Distinct audit-action constants (`epic_sync_dedup_match`, `epic_sync_source_conflict`, `epic_sync_dedup_conflict`) make downstream HIPAA-trail queries unambiguous: ambiguity (#1201), data-ownership skip (#1200), clean merge (#1191).
- No service boundary or contract changes; persistence layer only. No tRPC, no api-gateway, no BullMQ surface impact.
- `drizzle-orm` import expanded (`ne` added) — appropriate, persistence is the place that imports drizzle.
- Stacked on #1208 (approved). Retargeting to `main` after #1208 merges is documented in the PR body.

### Verdict
**Approve** — Ready to merge after #1208 lands.

Issue #1204, opened 2026-05-25
feat(epic): composite indexes for fingerprint dedup lookups (#1197 follow-up)


blamechris:
Follow-up to #1197.

The fingerprint dedup paths in `services/epic-connector/src/sync/persistence.ts` add one SELECT per row when the Epic-id mapping misses. Today's indexes cover the patient_id leading column on every table, which is adequate for moderate backfills but suboptimal once a single patient has thousands of historical records.

Current relevant indexes (`packages/db-schema/src/schema/`):
- encounters: `idx_encounters_patient` (patient_id) + `idx_encounters_start_time` — fingerprint filters on (patient_id, start_time, encounter_type), so today the planner uses patient_id idx then filters in-row.
- diagnoses: `idx_diagnoses_patient` (patient_id, status) — fingerprint filters on (patient_id, icd10_code OR snomed_code, onset_date).
- allergies: `idx_allergies_patient` (patient_id) — fingerprint adds rxnorm_code / snomed_code filter.
- medications: `idx_medications_patient` (patient_id, status) — fingerprint adds (rxnorm_code, started_at).
- vitals: `idx_vitals_patient_type` (patient_id, type, recorded_at) — fingerprint uses (patient_id, loinc_code, recorded_at). Not a perfect match — `type` (the CareBridge enum) ≠ `loinc_code`.

## Suggested indexes

- `idx_encounters_dedup` on (patient_id, start_time, encounter_type) — covers the Encounter fingerprint
- `idx_diagnoses_dedup_icd10` on (patient_id, icd10_code) WHERE icd10_code IS NOT NULL
- `idx_diagnoses_dedup_snomed` on (patient_id, snomed_code) WHERE snomed_code IS NOT NULL
- `idx_allergies_dedup_rxnorm` on (patient_id, rxnorm_code) WHERE rxnorm_code IS NOT NULL
- `idx_allergies_dedup_snomed` on (patient_id, snomed_code) WHERE snomed_code IS NOT NULL
- `idx_medications_dedup` on (patient_id, rxnorm_code, started_at) WHERE rxnorm_code IS NOT NULL
- `idx_vitals_dedup` on (patient_id, loinc_code, recorded_at) WHERE loinc_code IS NOT NULL

Partial indexes (`WHERE NOT NULL`) keep the index small — the fingerprint code returns null when the coded dimension is absent, so unindexed rows aren't queried.

## Acceptance criteria

- [ ] Migration adds the composite indexes listed above
[...]

PR #1222, opened 2026-05-26
feat(db): extend diagnoses dedup indexes to include onset_date (#1211)

### LINKED ISSUE #1211: Extend diagnoses dedup indexes to cover onset_date
## Context

Identified during review of PR #1206 (migration `0055_fingerprint_dedup_indexes.sql`).

## Description

`findDiagnosisByFingerprint` in `services/epic-connector/src/sync/persistence.ts` builds its predicate as:

```ts
const predicate = args.onsetDate
  ? and(
      eq(diagnoses.patient_id, args.patientId),
      codePredicate,
      eq(diagnoses.onset_date, args.onsetDate),
    )
  : and(eq(diagnoses.patient_id, args.patientId), codePredicate);
```

`codePredicate` is either `eq(diagnoses.icd10_code, ...)` or `eq(diagnoses.snomed_code, ...)`.

The two diagnoses dedup indexes added in #1206 are shaped as `(patient_id, icd10_code)` and `(patient_id, snomed_code)` — they cover the leading two-column predicate but not the trailing `onset_date` equality. The planner will still use them (it scans the prefix and filters), but a fully-covering shape `(patient_id, icd10_code, onset_date)` / `(patient_id, snomed_code, onset_date)` would let the planner satisfy the full predicate from the index alone — mirroring how `idx_medications_dedup` already includes `started_at` and `idx_vitals_dedup` includes `recorded_at`.

## Original Review Comment

> Consider including `onset_date` in `idx_diagnoses_dedup_icd10/_snomed` as the trailing column. `findDiagnosisByFingerprint` adds `eq(diagnoses.onset_date, args.onsetDate)` to the predicate when `onsetDate` is non-null. The current index will still be used (Postgres scans the patient+code prefix and filters), but a `(patient_id, icd10_code, onset_date)` shape would let the planner satisfy the full predicate from the index alone, mirroring how `idx_medications_dedup` includes `started_at`. Trade-off: slightly larger index for the common path.

## Acceptance Criteria

- [ ] New migration extends or replaces `idx_diagnoses_dedup_icd10` and `idx_diagnoses_dedup_snomed` to include `onset_date` as the trailing column (still partial on the code column).
- [ ] Migration test (`migration-XXXX-*.test.ts`) asserts the new column tuple.
- [ ] Confirm via `EXPLAIN ANALYZE` (or recorded plan) that the dedup query uses the new index shape end-to-end when `onsetDate` is supplied.
- [ ] Decide whether to drop the two-column variants (probably yes, since the three-column form remains usable for the two-column predicate via prefix scan).

## Refs
- #1204 (parent — leading-column shape gap)
- #1197 (introduced the fingerprint queries)
- #1191 (introduced fingerprint fallback)

blamechris:
## Summary
- Replaces 0055's `idx_diagnoses_dedup_icd10` / `idx_diagnoses_dedup_snomed` with three-column equivalents that include `onset_date` as the trailing column
- Three-column shape lets the planner cover `findDiagnosisByFingerprint`'s full predicate from the index; still usable for the two-column predicate via prefix scan
- Mirrors `idx_medications_dedup(patient_id, rxnorm_code, started_at)` and `idx_vitals_dedup(patient_id, loinc_code, recorded_at)`

## Test plan
- [x] New introspection test asserts the (patient_id, code, onset_date) shape

Closes #1211

### DIFF
--- a/packages/db-schema/drizzle/0056_diagnoses_dedup_indexes_with_onset.sql
+++ b/packages/db-schema/drizzle/0056_diagnoses_dedup_indexes_with_onset.sql
@@ -1,0 +1,35 @@
+-- #1211: extend the diagnoses dedup indexes added in 0055 (#1204) to
+-- carry `onset_date` as the trailing column, mirroring the
+-- (patient_id, coded-dim, temporal) shape used by
+--   idx_medications_dedup (patient_id, rxnorm_code, started_at)
+--   idx_vitals_dedup      (patient_id, loinc_code,  recorded_at)
+--
+-- The fingerprint query in services/epic-connector/src/sync/persistence.ts
+-- (findDiagnosisByFingerprint) is shaped as:
+--   WHERE patient_id = $1 AND <icd10|snomed>_code = $2 AND onset_date = $3
+-- The 0055 two-column index could only cover the leading two predicates;
+-- the planner had to recheck onset_date against the heap. The
+-- three-column shape lets the planner satisfy the full predicate from
+-- the index alone, and remains usable for the two-column predicate
+-- (icd10/snomed lookup without an onset filter) via a prefix scan.
+--
+-- DROP the 0055 variants first — CREATE INDEX IF NOT EXISTS would
+-- otherwise no-op when the narrower index of the same name is present.
+-- Partial WHERE NOT NULL is preserved so the index only covers rows
+-- the fingerprint query actually targets (Epic-sourced rows + any
+-- internal rows that happen to carry a code).
+--
+-- Plain CREATE INDEX (no CONCURRENTLY) matches the repo convention —
+-- see 0055/0034 for the rationale (Drizzle wraps each migration in
+-- a transaction; CONCURRENTLY cannot run inside one).
+
+DROP INDEX IF EXISTS "idx_diagnoses_dedup_icd10";
+DROP INDEX IF EXISTS "idx_diagnoses_dedup_snomed";
+
+CREATE INDEX IF NOT EXISTS "idx_diagnoses_dedup_icd10"
+  ON "diagnoses" USING btree ("patient_id", "icd10_code", "onset_date")
+  WHERE "icd10_code" IS NOT NULL;
+
+CREATE INDEX IF NOT EXISTS "idx_diagnoses_dedup_snomed"
+  ON "diagnoses" USING btree ("patient_id", "snomed_code", "onset_date")
+  WHERE "snomed_code" IS NOT NULL;
--- a/packages/db-schema/drizzle/meta/_journal.json
+++ b/packages/db-schema/drizzle/meta/_journal.json
@@ -421,6 +421,13 @@
       "when": 1747930320000,
       "tag": "0055_fingerprint_dedup_indexes",
       "breakpoints": true
+    },
+    {
+      "idx": 60,
+      "version": "7",
+      "when": 1747930380000,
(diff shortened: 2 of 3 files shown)


blamechris:
## Code Review: PR #1222

### Summary
Migration `0056_diagnoses_dedup_indexes_with_onset.sql` replaces the two-column diagnoses dedup indexes added in `0055` (`idx_diagnoses_dedup_icd10`, `idx_diagnoses_dedup_snomed`) with three-column variants that carry `onset_date` as the trailing column. This brings the diagnoses dedup shape into line with `idx_medications_dedup(patient_id, rxnorm_code, started_at)` and `idx_vitals_dedup(patient_id, loinc_code, recorded_at)`, and lets the planner fully cover `findDiagnosisByFingerprint`'s three-predicate path from the index alone.

### Strengths
- **Header comment is exemplary** — names the calling site, explains why DROP comes first (otherwise `CREATE INDEX IF NOT EXISTS` no-ops on the narrower index of the same name), and references `0055/0034` for the no-CONCURRENTLY rationale.
- **Idempotent and atomic** — `DROP INDEX IF EXISTS` + `CREATE INDEX IF NOT EXISTS` re-runs cleanly, and Drizzle wraps the entire `.sql` in a transaction so a partial DROP+failed CREATE rolls back together (no window where dedup runs unindexed).
- **Partial `WHERE <code> IS NOT NULL` preserved** — index footprint stays proportional to the rows the fingerprint query actually targets (Epic-sourced + any internal rows that happen to carry a code).
- **Prefix-scan compatibility** — three-column shape still serves the two-column fallback path in `findDiagnosisByFingerprint` (when `onsetDate` is null and the query degrades to `patient_id + code`).
- **Defensive introspection test** — `migration-0056-diagnoses-dedup-onset.test.ts` strips line comments before matching, asserts column order, asserts the `DROP` precedes the `CREATE`, asserts no `CONCURRENTLY` and no `ALTER TABLE`. Mirrors the 0055 test pattern.
- **Journal entry is well-formed** — `idx: 60`, `when: 1747930380000` (strictly greater than 0055's `1747930320000`), `breakpoints: true`. Caught by the existing journal-monotonicity test in `migration-journal.test.ts`.

### Issues Found

#### Critical (Must Fix)
None.

None — the migration is appropriately scoped and tested.

#### Nitpicks (Optional)
None.

### Verification

- `pnpm --filter @carebridge/db-schema test` — 66/66 pass, including the new 0056 introspection suite and the journal-monotonicity guard
- `gh pr checks 1222` — Build / Test / Typecheck all green on `17e7740`
- PR is `MERGEABLE`, no conflicts with `main`

### Audit Trail
- `idx 60` collision risk — none, this is the next available slot after `0055` (`idx 59`)
- DROP+CREATE atomicity — guaranteed by Drizzle's per-file transaction wrapper
- Partial WHERE clause — preserved verbatim from `0055`, matches the `<code> IS NOT NULL` shape on medications/vitals/allergies
- Journal monotonicity — `when` timestamp strictly increases (`1747930380000 > 1747930320000`), enforced by the existing test

### Architecture Notes
Migration is index-only (no `ALTER TABLE`), no behavioural changes to `findDiagnosisByFingerprint`, no schema-snapshot drift in `meta/0000_snapshot.json` (Drizzle's repo convention for this codebase only commits the initial snapshot — every later migration ships as a hand-written `.sql` + journal entry, see #1204 / #1197 / #1191).

### Verdict
**Approve** — Ready to merge.

PR #1225, opened 2026-05-26
test(epic): strengthen lab fingerprint test to cover distinct test_name (#1214)

### LINKED ISSUE #1214: epic: strengthen lab fingerprint test — exercise actual two-distinct-test collapse
## Context

Identified during review of PR #1207.

## Description

The new test `dedups single-result Observations using test_name as panel_name` (`services/epic-connector/src/__tests__/persistence.test.ts`) reuses the same `RESOURCE` fixture as the first test in the block, so it doesn't actually exercise the two-distinct-test-names-colliding-on-the-same-panel scenario its title implies. The current assertions only re-prove what the first test already covers.

## Original Review Comment

> Test `dedups single-result Observations using test_name as panel_name` uses the same `RESOURCE` fixture as the first test, so the assertion that "single-result deduping works" is structurally identical to the first case rather than exercising the *two-distinct-tests-collapsing-to-one-panel* scenario the title implies. Consider either renaming the test for clarity or adding a second different-test-name resource to actually stress the collapse risk.

## Acceptance Criteria

- [ ] Either rename the test to match what it actually asserts, OR
- [ ] Add a second `RESOURCE` with a different `code.text` and verify the two persist calls reach different panel rows (i.e. verify the dedup *only* fires on matching test_name, not just any single-result Observation)

blamechris:
## Summary
- Strengthens the lab fingerprint dedup test in `services/epic-connector/src/__tests__/persistence.test.ts` to actually exercise the two-distinct-test-name collapse risk implied by its title
- Adds a second fixture with a different `code.text` and asserts distinct test_names produce distinct panel rows

## Test plan
- [x] Test passes
- [x] Same-test_name still collapses (regression check)
- [x] Different-test_name now produces distinct panels (new assertion)

Closes #1214

### DIFF
--- a/services/epic-connector/src/__tests__/persistence.test.ts
+++ b/services/epic-connector/src/__tests__/persistence.test.ts
@@ -605,10 +605,17 @@ describe("persistObservation (lab) — panel-level dedup (#1202)", () => {
     expect(dedup).toBeUndefined();
   });
 
-  it("dedups single-result Observations using test_name as panel_name", async () => {
-    // Two single-test Epic Observations for the same patient at the same
-    // collected_at + test_name should resolve to the same lab_panels row.
-    const existingPanelId = "internal-panel-cbc-1";
+  it("dedups single-result Observations using test_name as panel_name — distinct test_names do NOT collapse (#1214)", async () => {
+    // The panel fingerprint is patient_id + panel_name + collected_at, and
+    // for single-result Epic Observations panel_name comes from test_name
+    // (= code.text). Two Observations with the SAME test_name MUST reuse a
+    // panel, while two Observations with DIFFERENT test_names at the same
+    // collected_at MUST NOT — otherwise we'd silently merge a Hemoglobin
+    // result and a Sodium result into one panel. Run both paths in one
+    // test so regressions on either side fail loudly.
+
+    // ── Path A: same test_name (Hemoglobin) → fingerprint HIT, reuse panel ──
+    const existingPanelId = "internal-panel-hemoglobin-1";
     db.willSelect([]); // mapping miss
     db.willSelect([
       {
@@ -622,7 +629,7 @@ describe("persistObservation (lab) — panel-level dedup (#1202)", () => {
     db.willInsert(); // mapping
     db.willInsert(); // audit
 
-    const result = await persistObservation(RESOURCE, PATIENT_ID);
+    const hemoglobinResult = await persistObservation(RESOURCE, PATIENT_ID);
 
     // Return value is the newly-inserted lab_results.id (the leaf), not
     // the reused panel id (#1207). Round-2 syncs need the leaf id to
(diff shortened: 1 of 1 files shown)


blamechris:
## Code Review: PR #1225

### Summary
Strengthens `services/epic-connector/src/__tests__/persistence.test.ts` to actually exercise the two-distinct-test_name dedup risk implied by the original test title. Adds a Sodium fixture (different `code.text`, identical `effectiveDateTime` to the existing Hemoglobin RESOURCE) and asserts that a NEW `lab_panels` row is inserted with `panel_name="Sodium"` — proving the panel fingerprint is keyed on `panel_name`, not just `patient + collected_at`.

### Strengths
- Test name now matches behaviour: "distinct test_names do NOT collapse" is exactly what the assertions verify.
- Comments explain *why* the test exists, including the pre-#1214 gap ("the same RESOURCE fixture was reused").
- Reuses the existing harness (`createMockDb`) — no new mocking patterns introduced.
- `db.reset()` between Path A and Path B is sound: confirmed `reset()` clears `insert.calls` (`packages/test-utils/src/mock-db.ts:249-265`), so Path B's `db.insert.calls.find(...)` only sees Sodium-path inserts.
- Spread-from-RESOURCE for `SODIUM_RESOURCE` keeps the divergence to only the dimension under test (`id`, `code`, `valueQuantity`) — `subject`, `category`, `effectiveDateTime` carry over so the fingerprint differs *only* by `panel_name`.
- Path B holds `effectiveDateTime` identical on purpose — explicitly proves the key includes `panel_name`. Good intent comment on line 678-679.

### Issues Found

#### Critical (Must Fix)
None.

| persistence.test.ts | 608 | Consider splitting Path A / Path B into two separate `it` blocks | Single-purpose tests fail more loudly (one failure mode per test). Current shape works, but Vitest stack traces will land on whichever path tripped first, hiding the other side if both regress. The `db.reset()` boundary already signals "two scenarios in one test" — making it two tests is mostly cosmetic but matches the file's prevailing one-scenario-per-`it` style. |
| persistence.test.ts | 696-700 | `panelInsertSodium` matches on `panel_name !== undefined`, which would also match the `lab_results` insert if it ever grew a `panel_name` field | Defensive: prefer matching on the table sentinel (`labPanels`) via `findInsertFor(labPanels)` — there's already a helper at line 110. Today `lab_results` has no `panel_name`, so the test passes, but the schema-coupling is brittle. |

#### Nitpicks (Optional)
- Line 654: `expect(panelInsertsAfterHemoglobin).toHaveLength(0)` would be slightly clearer as `expect(panelInsertsAfterHemoglobin).toEqual([])` — same guarantee, but the failure message prints the offending row(s) instead of just a count.
- Comment on line 661 reads "pre-#1214" — the PR title says it closes #1214, so "pre-this-PR" or "previously" would be tense-accurate when read on a future blame.

None — all suggestions are tightening on an already-passing test. The behaviour under test is the production guarantee from #1202 / #1207, and this PR's assertions correctly pin it.

### Architecture Notes
Pure test strengthening — no production code, no new mocks, no new dependencies. Sits cleanly on top of #1202 (panel-level dedup) and #1207 (leaf-id return). The `code.text → conversion.row.test_name → labPanels.panel_name` chain is verified end-to-end via the existing harness, and `services/fhir-gateway/src/observation-mapper.ts:376-395` confirms `test_name` derives from `code.text` first, so the `"Sodium"` fixture actually drives a distinct fingerprint. No PHI, no event-bus changes, no schema changes.

### Verdict
- [x] **Approve** — Ready to merge. Suggestions above are optional polish; the test now does what its title claims and forecloses the silent-collapse regression that motivated #1214.

PR #1228, opened 2026-05-26
feat(epic): surface epic_id_conflict discriminator on writer return (#1217)

### LINKED ISSUE #1217: epic-connector: surface 'epic_id_conflict' discriminator on writer return value
## Context

Identified during review of PR #1210 (#1201 — Epic-id conflict detection on fingerprint merge).

## Description

The #1200 source_system guard path returns `conflict: \"source_system_conflict\"` on the writer's result object so callers can distinguish ambiguity-skip from a clean insert. The #1201 conflict path returns `{ inserted: true }` with no caller-side discriminator — the conflict is only visible in the audit_log row.

Add a `conflict: \"epic_id_conflict\"` field (or analogous discriminator) to writer return values when the conflict path fires, mirroring the #1200 pattern. Useful for:

- Downstream metrics / alerting on conflict counts without audit_log queries
- Higher-level sync orchestrators that want to summarise per-resource sync outcomes
- Symmetry with the existing source-conflict discriminator

## Original Review Comment

> The #1200 path already returns `conflict: \"source_system_conflict\"` so callers can distinguish ambiguity from clean inserts. The #1201 path returns `{ inserted: true }` with no caller signal — only the audit row records it. If any downstream caller is metrics-gathering on conflict counts (or needs to alert on identity ambiguity), they would currently have to query audit_log. Not blocking — audit-only is a defensible policy — but worth tracking.

## Acceptance Criteria

- [ ] Writers return `conflict: \"epic_id_conflict\"` (or agreed name) on the dedup-conflict path
- [ ] Type union for the writer result widened to include the new discriminator
- [ ] Test coverage pins both presence on conflict and absence on clean insert / clean merge
- [ ] sync-jobs / sync-worker summary surfaces the new conflict count where appropriate

blamechris:
## Summary
- Adds `conflict: "epic_id_conflict"` discriminator to writer return values on the dedup-conflict path in `services/epic-connector/src/sync/persistence.ts`
- Widens the writer result type union to a literal union `"source_system_conflict" | "epic_id_conflict"` so call-sites get exhaustive narrowing
- Applies the discriminator across every persist* function with a fingerprint-conflict path (Patient, MedicationRequest/Statement, Observation/vital, Condition, Allergy, Encounter)

## Background
The #1200 source_system guard path returns `conflict: "source_system_conflict"` on the writer result so the orchestrator can distinguish ambiguity from a clean insert. The #1201 Epic-id conflict path (landed in #1210) returned a bare `{ inserted: true }` with no caller-side signal — the conflict was only visible by reading the freshly-written audit_log row. This PR mirrors the #1200 pattern.

## sync-worker / sync-jobs counter
The existing counter in `syncResourceType` is already discriminator-agnostic:

```
if (persisted.conflict) result.conflicts++;
else if (persisted.inserted) result.imported++;
else result.updated++;
```

So a `conflict: "epic_id_conflict"` return automatically counts toward `result.conflicts` (mirroring source-conflicts) and suppresses the clinical-event emit (`!persisted.conflict && persisted.internalId` gate). No structural code change to `sync-jobs.ts` was needed; a new explicit test in `sync-jobs.test.ts` pins the behaviour so a future counter refactor can't regress the wiring.

## Test plan
- [x] Conflict path returns `{ inserted: true, conflict: "epic_id_conflict" }` (asserted for Encounter, Condition, MedicationRequest in `persistence-id-conflict.test.ts`)
- [x] Clean merge / same-Epic-id path returns no `conflict` field (existing `toBeUndefined()` assertions preserved)
- [x] sync-jobs orchestrator counts `epic_id_conflict` toward `result.conflicts` and suppresses the clinical-event emit (new test)
- [x] Full epic-connector test suite (311 pass)
- [x] `pnpm typecheck` + `pnpm lint` clean

Closes #1217

### DIFF
--- a/services/epic-connector/src/__tests__/persistence-id-conflict.test.ts
+++ b/services/epic-connector/src/__tests__/persistence-id-conflict.test.ts
@@ -186,6 +186,11 @@ describe("persistEncounter — Epic-id conflict on fingerprint match (#1201)", (
     expect(result.kind).toBe("encounter");
     expect(result.inserted).toBe(true);
     expect(result.internalId).not.toBe(existingInternalId);
+    // #1217: caller-visible discriminator mirrors source_system_conflict
+    // so the sync orchestrator can distinguish a forked-insert (fingerprint
+    // collided on a different Epic id) from a clean import without
+    // re-reading the audit_log row.
+    expect(result.conflict).toBe("epic_id_conflict");
 
     // UPDATE must NOT have happened — the existing row is left alone.
     expect(db.update).not.toHaveBeenCalled();
@@ -326,6 +331,8 @@ describe("persistCondition — Epic-id conflict on fingerprint match (#1201)", (
     expect(result.kind).toBe("diagnosis");
     expect(result.inserted).toBe(true);
     expect(result.internalId).not.toBe(existingInternalId);
+    // #1217: discriminator surfaces the forked-insert reason to callers.
+    expect(result.conflict).toBe("epic_id_conflict");
     expect(db.update).not.toHaveBeenCalled();
 
     const audit = findAuditInsert("epic_sync_dedup_conflict");
@@ -391,6 +398,8 @@ describe("persistMedicationRequest — Epic-id conflict on fingerprint match (#1
     expect(result.kind).toBe("medication");
     expect(result.inserted).toBe(true);
     expect(result.internalId).not.toBe(existingInternalId);
+    // #1217: discriminator surfaces the forked-insert reason to callers.
+    expect(result.conflict).toBe("epic_id_conflict");
     expect(db.update).not.toHaveBeenCalled();
 
     const audit = findAuditInsert("epic_sync_dedup_conflict");
--- a/services/epic-connector/src/__tests__/sync-jobs.test.ts
+++ b/services/epic-connector/src/__tests__/sync-jobs.test.ts
@@ -194,6 +194,41 @@ describe("Epic sync-jobs (#391)", () => {
     expect(result.imported).toBe(0);
   });
 
+  // #1217 — epic_id_conflict is the fingerprint-collision-on-different-
+  // Epic-id forked-insert path. Persistence inserts a NEW internal row
+  // (so `inserted: true`) but the operator still needs the conflict
+  // signal: a row pair needs manual reconciliation. Treated identically
+  // to source_system_conflict by the orchestrator — counted as a
+  // conflict (NOT imported), event emit suppressed.
+  it("counts epic_id_conflict as a conflict and suppresses the clinical event", async () => {
+    persistMedicationRequest.mockResolvedValueOnce({
+      internalId: "internal-med-forked",
+      kind: "medication",
+      inserted: true,
+      conflict: "epic_id_conflict",
+    });
+    const client = makeFakeClient({
+      MedicationRequest: [
+        { resourceType: "MedicationRequest", id: "med-B" },
+      ],
+    });
+
+    const result = await runSingleResourceSync(
+      {
+        patientId: PATIENT_ID,
+        epicPatientFhirId: EPIC_PATIENT_FHIR_ID,
+        resourceType: "MedicationRequest",
+        watermark: null,
+      },
+      { client, emit },
+    );
+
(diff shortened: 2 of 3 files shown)


blamechris:
## Code Review: PR #1228

### Summary
Surfaces a `conflict: "epic_id_conflict"` discriminator on writer return values whenever the `persist*` family inserts a forked row because a fingerprint match landed on an internal row already mapped to a different Epic FHIR id. Mirrors the #1200 `source_system_conflict` pattern so the sync orchestrator (and any future consumer) can distinguish the forked-insert path from clean inserts/merges without re-reading the `audit_log` row.

### Strengths
- **Type union widened correctly.** `conflict?: string` becomes `conflict?: "source_system_conflict" | "epic_id_conflict"` — call-sites get exhaustive narrowing, and a typo on either string is now a compile-time error rather than a silent operational miss.
- **Symmetrical coverage.** All 6 persist paths with a fingerprint-collision INSERT branch carry the discriminator: Patient (L823), MedicationRow (L1039, driving both MedReq + MedStmt), Observation/vital (L1217), Condition (L1511), Allergy (L1692), Encounter (L1889). The discriminator return is inside the `if (conflictMapping && fingerprintHit)` branch on every path, immediately after the `logDedupConflict` call — fires iff the audit row is written.
- **Orchestrator stays discriminator-agnostic.** `sync-jobs.ts` L231 (`if (persisted.conflict)`) and L239 (`!persisted.conflict && persisted.internalId`) use truthiness, so a new discriminator string flows through `result.conflicts++` and the clinical-event suppression gate without any structural change. Documented in PR body, pinned by the new sync-jobs test.
- **Test pinning is thorough.** Conflict paths assert `conflict: "epic_id_conflict"` (Encounter L191, Condition L335, MedicationRequest L402); the clean-merge / same-Epic-id paths preserve the `toBeUndefined()` assertions (L268, L442) so regressions in either direction are caught. The sync-jobs test pins both presence-counts-as-conflict AND event-suppression in one go.
- **PersistResult doc-comment is genuinely helpful.** The block comment on the `conflict?` field explains exactly when each discriminator fires + what the corresponding `inserted` value will be — future maintainers reading the type get the full story without grepping.
- **Zero unrelated churn.** +101/-2 across exactly 3 files; nothing touched outside the conflict-discriminator concern. Closes #1217 cleanly.

### Issues Found

#### Critical (Must Fix)
None.

None — the change matches the established #1200 pattern exactly. The two discriminator strings + the orchestrator's truthiness check form a self-consistent contract.

#### Nitpicks (Optional)
- The `persistMedicationRow` discriminator return covers both `persistMedicationRequest` and `persistMedicationStatement` callers; the test only covers MedicationRequest. Statement is structurally identical (same `persistMedicationRow` delegate) so risk is near-zero, but a one-line dispatch test would close the loop. Non-blocking.
- `PersistResult.conflict` is now a literal union but `logDedupConflict` / `epic_sync_dedup_conflict` / `epic_sync_source_conflict` audit-event strings are still bare string literals elsewhere. Promoting both to a shared `const EPIC_CONFLICT_KIND = { ... } as const` would let callers (and future analytics queries) round-trip through one canonical map. Out of scope for #1217.

None — the nitpicks above are low-value polish and not worth tracking as issues.

### Architecture Notes
- Aligns with the #1200 contract precisely: orchestrator distinguishes "ambiguous skip" (`source_system_conflict`, `inserted: false`) from "forked insert" (`epic_id_conflict`, `inserted: true`), both counted as conflicts, both suppressing the clinical-event emit.
- BullMQ event-emission gate (`!persisted.conflict && persisted.internalId`) correctly suppresses for both discriminator values — a forked-insert row is operator-action territory, not something we want to auto-feed into the ai-oversight pipeline until reconciled.
- PersistResult is the single boundary type for the entire `persist*` family; widening it here is the right seam and keeps `sync-jobs.ts`/`sync-worker.ts` mechanically unchanged.

### Verdict

CI green (Typecheck + Build + Test all SUCCESS on dce27b0), MERGEABLE, zero attribution, all 6 writer paths covered with symmetric pattern matching #1200, tests pin both presence-on-conflict and absence-on-clean-merge plus orchestrator wiring.

PR #1230, opened 2026-05-26
feat(epic): surface up to N existing Epic-ids in dedup_conflict audit (#1219)

### LINKED ISSUE #1219: epic-connector: surface up to N existing Epic-ids in dedup_conflict audit detail
## Context

Identified during review of PR #1210 (#1201 — Epic-id conflict detection on fingerprint merge).

## Description

`findConflictingMapping` (services/epic-connector/src/sync/persistence.ts:271-296) uses `LIMIT 1` to find any pre-existing Epic mapping pointing at the matched internal id with a different `resource_id`. Today the `fhir_resources` schema makes multi-mapping impossible — `(resource_type, resource_id)` is unique and `internal_record_id` is a target — so in practice a row can only have one conflicting mapping at a time.

But if a prior bug, manual DB intervention, or future schema change ever results in N>1 conflicting mappings on the same internal id, the audit row would only surface ONE of them. A manual reviewer chasing the conflict would have to query `fhir_resources` directly to find the others.

Defensive improvement: switch `findConflictingMapping` to `LIMIT 5` (or similar), return an array, and emit `details.existing_epic_fhir_ids: string[]` in the audit row (alongside or instead of the single `existing_epic_fhir_id`).

## Original Review Comment

> `LIMIT 1` is sufficient for the conflict detection, but loses information when MULTIPLE different Epic ids already map to the same internal id. Today this is presumably impossible (mapping is keyed on `(resource_type, resource_id)`), but if it ever happens (eg. a prior bug, manual DB intervention), the audit row only surfaces one of N. A `LIMIT 5` + array in `details.existing_epic_fhir_ids` would future-proof. Nit; not blocking.

## Acceptance Criteria

- [ ] `findConflictingMapping` returns an array (capped at a small N) instead of a single row
- [ ] `logDedupConflict` payload includes `existing_epic_fhir_ids: string[]`
- [ ] Backwards-compat: keep `existing_epic_fhir_id` populated with the first entry for downstream consumers
- [ ] Test pins both single-conflict and (synthetic) multi-conflict behavior

blamechris:
## Summary
- `findConflictingMapping` (services/epic-connector/src/sync/persistence.ts) now returns an array of up to 5 conflicting Epic FHIR ids instead of a single row
- `logDedupConflict` emits `existing_epic_fhir_ids: string[]` in audit details, alongside the existing `existing_epic_fhir_id` (first entry, for backwards compat)
- Multi-conflict synthetic test exercises the array shape

## Test plan
- [x] Single-conflict: array length 1, first entry matches existing field
- [x] Multi-conflict: array length matches inserted rows (capped at 5)
- [x] All existing conflict tests still pass

Closes #1219

### DIFF
--- a/services/epic-connector/src/__tests__/persistence-multi-conflict.test.ts
+++ b/services/epic-connector/src/__tests__/persistence-multi-conflict.test.ts
@@ -1,0 +1,197 @@
+/**
+ * Multi-conflict audit surfacing on the fingerprint-merge dedup_conflict
+ * path (#1219).
+ *
+ * Background: #1201 introduced the `epic_sync_dedup_conflict` audit row
+ * for the case where a fingerprint hit lands on an internal row already
+ * mapped to a DIFFERENT Epic FHIR id. The original implementation used
+ * `LIMIT 1` and surfaced a single `existing_epic_fhir_id` in the audit
+ * details — sufficient for detection, but loses information when more
+ * than one Epic id maps to the same internal id (today impossible via
+ * the `(resource_type, resource_id)` unique key on `fhir_resources`, but
+ * possible after manual DB intervention or any future schema change
+ * that loosens the constraint).
+ *
+ * This suite pins the new behaviour:
+ *  - Single-conflict (existing case, exactly one row): the audit
+ *    details carry `existing_epic_fhir_ids` as a 1-element array AND
+ *    the back-compat `existing_epic_fhir_id` is the same string.
+ *  - Multi-conflict (synthetic 2-3 rows): `existing_epic_fhir_ids` is
+ *    the full array (capped at 5), and `existing_epic_fhir_id` is the
+ *    first entry.
+ */
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { createMockDb, type MockDb } from "@carebridge/test-utils";
+
+let db: MockDb;
+
+vi.mock("@carebridge/db-schema", () => ({
+  getDb: () => db,
+  patients: {
+    id: "id",
+    mrn_hmac: "mrn_hmac",
+    patient_id: "patient_id",
+  },
+  medications: {
+    id: "id",
+    patient_id: "patient_id",
+    rxnorm_code: "rxnorm_code",
+    started_at: "started_at",
+  },
+  vitals: {
+    id: "id",
+    patient_id: "patient_id",
+    loinc_code: "loinc_code",
+    recorded_at: "recorded_at",
+  },
+  labPanels: {},
+  labResults: {},
+  diagnoses: {
+    id: "id",
+    patient_id: "patient_id",
+    icd10_code: "icd10_code",
+    snomed_code: "snomed_code",
+    onset_date: "onset_date",
+  },
+  allergies: {
+    id: "id",
+    patient_id: "patient_id",
+    rxnorm_code: "rxnorm_code",
+    snomed_code: "snomed_code",
+  },
+  encounters: {
+    id: "id",
+    patient_id: "patient_id",
+    start_time: "start_time",
+    encounter_type: "encounter_type",
+  },
+  fhirResources: {
+    id: "id",
+    resource_type: "resource_type",
+    resource_id: "resource_id",
+    internal_record_id: "internal_record_id",
+  },
+  auditLog: {},
+  hmacForIndex: (value: string) => `hmac:${value}`,
+}));
+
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #1230

### Summary
Widens `findConflictingMapping` -> `findConflictingMappings` to return an array of up to 5 conflicting Epic FHIR ids (capped via SQL `LIMIT`), and emits both `existing_epic_fhir_ids: string[]` (new) and `existing_epic_fhir_id: string` (back-compat = `[0]`) in the `epic_sync_dedup_conflict` audit details. All 6 callsites (Patient, MedicationRow, Observation/vital, Condition, Allergy, Encounter) updated to the array-length branching pattern. Adds a synthetic 3-row multi-conflict test alongside the existing single-row coverage.

### Strengths
- `CAP_CONFLICTING_MAPPINGS = 5` is a named constant with a doc-block explaining the rationale (forensic value vs. payload bloat).
- Backwards-compat field `existing_epic_fhir_id` is preserved and explicitly contracted as `existingEpicFhirIds[0]` in the docstring, so consumers of #1201 audit rows keep working unchanged.
- Single-conflict assertions in `persistence-id-conflict.test.ts` remain valid because the scalar field is still emitted as the first array entry.
- Branching condition changed cleanly from `!conflictMapping` to `conflictingEpicIds.length === 0` (and inverse) at every callsite; no `null` checks left behind.
- The new test pins both shapes (1-element array + scalar match, 3-element array + scalar = first) and re-asserts the #1201 invariants (`new_epic_fhir_id`, `matched_internal_id`).
- Empty-array sentinel (`return []`) handles the under-queued mock case the same way `null` previously did — no test mocks need updating beyond the new file.
- `internalRecordId` is dropped from the return shape since all callers already had `fingerprintHit.internalId` in scope — no information loss, simpler signature.

### Issues Found

#### Critical (Must Fix)
None.

| services/epic-connector/src/sync/persistence.ts | ~310 | Defense-in-depth: belt-and-braces `.slice(0, CAP_CONFLICTING_MAPPINGS)` after the `.map()` | Today the cap is purely SQL-side (`.limit(5)`). If a future caller swaps the query builder, accidentally drops the limit, or a mock under-/over-queues, the audit payload silently grows. A no-op JS slice makes the cap a code-level invariant, not just a SQL-level one. |
| services/epic-connector/src/sync/persistence.ts | 363-364 | Consider emitting `existing_epic_fhir_id_count: number` alongside the array | Downstream consumers (alerts, dashboards) can filter on `count > 1` without parsing the array. The whole point of #1219 is to surface the multi-mapping case — making it grep-able is cheap. |

#### Nitpicks (Optional)
- The two consecutive JSDoc blocks at lines 270-296 (function doc + constant doc) read fine but the constant doc could move directly above `const CAP_CONFLICTING_MAPPINGS` so the rationale lives next to the value rather than between two unrelated doc blocks.
- `persistence-multi-conflict.test.ts` only exercises the Encounter path. The other 5 resource types share the helper so collapse-risk is low, but a tiny parametrized loop over `[persistPatient, persistMedicationRow, ...]` would lock the shape in for all consumers. Existing PR test plan acknowledges this is intentional — capturing as a follow-up.

| Defense-in-depth `.slice()` cap | [#1231](https://github.com/blamechris/carebridge/issues/1231) | Pure hardening, no current bug — keep this PR minimal. |
| Emit `existing_epic_fhir_id_count` | [#1232](https://github.com/blamechris/carebridge/issues/1232) | Schema-additive but out of scope; consumer side hasn't asked yet. |
| Parametrize multi-conflict test across all 6 resource types | [#1233](https://github.com/blamechris/carebridge/issues/1233) | Coverage tightening, not a correctness gap. |

### Architecture Notes
- Audit-log emission stays append-only and within the `epic_sync_dedup_conflict` action shape established by #1201. No new action, no schema change — purely a details-blob widening.
- No PHI exposure risk: only Epic FHIR ids (system identifiers, not patient identifiers) are emitted.
- `CAP_CONFLICTING_MAPPINGS` is module-private which is correct — callers should not be able to override the forensic cap.
- Pattern (`.map(row => row.resource_id)`) keeps the function pure and the return type narrow (`string[]`), which is what every callsite actually needs. The previous `ConflictingMapping` interface bundled `internalRecordId` that no caller used.

### Verdict
- [x] Approve - Ready to merge
- [ ] Request Changes
- [ ] Comment

LGTM. The deferred items are all hardening / observability, not blockers.

PR #1238, opened 2026-05-26
fix(epic): defense-in-depth slice cap on findConflictingMappings (#1231)

### LINKED ISSUE #1231: epic-connector: defense-in-depth .slice() cap on findConflictingMappings
## Context

Identified during review of PR #1230 (closes #1219).

## Description

`findConflictingMappings` in `services/epic-connector/src/sync/persistence.ts` enforces the 5-row forensic cap purely at the SQL layer via `.limit(CAP_CONFLICTING_MAPPINGS)`. If a future refactor swaps the query builder, accidentally drops the limit, or a test mock under/over-queues rows, the audit payload silently grows past the bound — defeating the cap's purpose (bounding the `epic_sync_dedup_conflict` audit payload size).

Adding a belt-and-braces `.slice(0, CAP_CONFLICTING_MAPPINGS)` after the `.map()` makes the cap a code-level invariant, not just a SQL-level one. It is a no-op in the happy path but cheap insurance.

## Original Review Comment

> Today the cap is purely SQL-side (.limit(5)). If a future caller swaps the query builder, accidentally drops the limit, or a mock under-/over-queues, the audit payload silently grows. A no-op JS slice makes the cap a code-level invariant, not just a SQL-level one.

## Acceptance Criteria

- [ ] `findConflictingMappings` returns `rows.map(...).slice(0, CAP_CONFLICTING_MAPPINGS)`
- [ ] Add a unit test that hands the function 7 mocked rows and asserts the returned array has length 5

blamechris:
## Summary
- Adds `.slice(0, CAP_CONFLICTING_MAPPINGS)` after the `.map()` in `findConflictingMappings` (services/epic-connector/src/sync/persistence.ts) so the 5-row cap is a code-level invariant, not just SQL-level
- Test: hand the function 7 mocked rows → returned array has length 5

## Test plan
- [x] New test pins the cap at the JS layer
- [x] Existing multi-conflict + id-conflict tests still pass

Closes #1231

### DIFF
--- a/services/epic-connector/src/__tests__/persistence-cap-conflicting-mappings.test.ts
+++ b/services/epic-connector/src/__tests__/persistence-cap-conflicting-mappings.test.ts
@@ -1,0 +1,154 @@
+/**
+ * Defense-in-depth cap on findConflictingMappings (#1231).
+ *
+ * Background: #1219 introduced a 5-row forensic cap on the conflicting
+ * Epic-mapping lookup used by the `epic_sync_dedup_conflict` audit row.
+ * The cap was enforced purely at SQL via `.limit(CAP_CONFLICTING_MAPPINGS)`.
+ * A future query-builder refactor, an accidentally-dropped limit, or a
+ * test mock that over-queues rows could let the audit payload grow past
+ * the bound.
+ *
+ * This suite pins the cap at the JS layer: hand the function 7 rows
+ * (bypassing the SQL `.limit()` via the mock DB) and assert the audit's
+ * `existing_epic_fhir_ids` array is truncated to exactly 5 entries — the
+ * first 5 in arrival order, with the back-compat scalar
+ * `existing_epic_fhir_id` still pointing at the first entry.
+ */
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { createMockDb, type MockDb } from "@carebridge/test-utils";
+
+let db: MockDb;
+
+vi.mock("@carebridge/db-schema", () => ({
+  getDb: () => db,
+  patients: {
+    id: "id",
+    mrn_hmac: "mrn_hmac",
+    patient_id: "patient_id",
+  },
+  medications: {
+    id: "id",
+    patient_id: "patient_id",
+    rxnorm_code: "rxnorm_code",
+    started_at: "started_at",
+  },
+  vitals: {
+    id: "id",
+    patient_id: "patient_id",
+    loinc_code: "loinc_code",
+    recorded_at: "recorded_at",
+  },
+  labPanels: {},
+  labResults: {},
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #1238

### Summary
Minimal, surgical defense-in-depth fix for #1231: adds `.slice(0, CAP_CONFLICTING_MAPPINGS)` after `.map()` in `findConflictingMappings` so the 5-row forensic cap on the `epic_sync_dedup_conflict` audit payload is a code-level invariant in addition to the existing SQL `.limit()`. Pinned by one new test that bypasses SQL via the mock DB and asserts the JS-side cap holds.

### Strengths
- **Correct placement** — `.slice()` chained immediately after `.map()` on line 347; uses the named constant `CAP_CONFLICTING_MAPPINGS` (not a magic `5`), so any future cap change in one place stays consistent across SQL limit + JS slice.
- **Inline comment cites #1231 and explains the three failure modes** it guards against (query-builder refactor, dropped limit, over-queuing mock). Future readers won't ask "why is this here, isn't the SQL limit enough?".
- **Test isolates the JS layer cleanly** — over-queues the mock with 7 rows (the mock's `willSelect` ignores `.limit()`, simulating the exact failure mode), then asserts `existing_epic_fhir_ids` is exactly 5 long.
- **Beyond AC**: also asserts (a) arrival-order preservation via `slice(0, 5)` equality, and (b) the back-compat scalar `existing_epic_fhir_id` still equals the first entry — locks down both contracts from #1219 in one shot.
- **Mock harness mirrors `persistence-multi-conflict.test.ts`** — same `vi.mock("@carebridge/db-schema", ...)` shape, same `createMockDb()` from `@carebridge/test-utils`, same `findAuditInsert()` helper pattern. Consistent with existing test style.
- **Zero production behaviour change in the happy path** — single-row and zero-row paths are unaffected; `.slice(0, 5)` on an array of length less-than-or-equal-to-5 is a no-op.

### Issues Found

#### Critical (Must Fix)
None.

None — the patch is the smallest possible change that satisfies both acceptance criteria from #1231, and CI is green on all three checks (Build, Test, Typecheck).

#### Nitpicks (Optional)
None.

None — the change is self-contained.

### Architecture Notes
Fits the established pattern in `services/epic-connector/src/sync/persistence.ts`: forensic audit payloads bounded by a named constant, with belt-and-braces enforcement at both the query layer and the application layer. No event emission, schema, or cross-service implications. Mirrors the prior #1219 work it extends.

### Verdict
- [x] **Approve** — Ready to merge.

Tiny patch, correct semantics, comprehensive test (length + ordering + back-compat scalar), clear inline rationale. Closes #1231 cleanly.

PR #1240, opened 2026-05-26
feat(epic): extend source_system guard to lab_panels persister (#1234)

### LINKED ISSUE #1234: feat(epic): extend source_system guard to lab_panels and procedures persisters
## Context

Identified during review of PR #1223 (which narrowed the `persistence.ts` header doc to name the five guarded resource tables).

## Description

The `persistence.ts` source_system conflict guard (Epic-id path #1183 + fingerprint path #1200) is wired into five resource persisters today: encounters, diagnoses, allergies, medications, vitals. The schema, however, carries a `source_system` column on two additional clinical tables:

- `lab_panels` (`packages/db-schema/src/schema/clinical-data.ts:102`)
- `procedures` (`packages/db-schema/src/schema/clinical-data.ts:137`)

The `persistObservation` lab branch (`services/epic-connector/src/sync/persistence.ts:1276-1306`) inserts a new `lab_panels` row tagged `source_system: EPIC_SOURCE_SYSTEM_TAG` but does NOT call `logSourceConflict` / `logFingerprintSourceConflict` when it reuses an existing panel via fingerprint hit (lines 1226-1274). A CareBridge-originated `lab_panels` row whose `(patient_id, panel_name, collected_at)` collides with an Epic Observation would be silently attached to with a new `lab_results` row, and no audit row would record the source-ownership conflict.

`procedures` has no persister in `persistence.ts` today — Epic Procedure resources are not synced. If/when they are added, the guard should be wired in from the start.

## Original Review Comment

> The header could also mention that `lab_panels` and `procedures` carry a `source_system` column today but are not yet wired into the conflict-check path. The existing inline comment on the lab branch already flags it (\"Existing lab_results rows are leaves — conflict logic on the panel level is out of scope for the first cut.\"). Tracking separately.

## Acceptance Criteria

- [ ] `persistObservation` lab-panel fingerprint-hit branch checks `lab_panels.source_system` before attaching a new `lab_results` row to a CareBridge-originated panel.
- [ ] On conflict: skip the attach, still upsert the `fhir_resources` mapping, emit an audit row with `action=\"epic_sync_source_conflict\"` (consistent with #1200).
- [ ] Test coverage in the epic-connector persistence test suite for the lab_panels conflict path.
- [ ] (If/when Procedure sync is added in a future PR) the guard is wired in from the start.

## References

- PR #1223 (narrowed the header doc)
- #1200 (source_system fingerprint guard for the five wired resources)
- #1183 (source_system Epic-id guard)
- #1190 (added source_system columns to encounters/diagnoses/allergies)

blamechris:
## Summary
- Adds source_system guard to `persistObservation`'s lab fingerprint-hit branch in `services/epic-connector/src/sync/persistence.ts`
- When matched `lab_panels` row has `source_system != 'epic'` (CareBridge-originated): skip lab_results attach, upsert fhir_resources mapping, emit `epic_sync_source_conflict` audit row
- When matched panel is Epic-sourced: existing dedup behavior preserved (lab_result attached)
- Procedures persister does not exist yet; out of scope per #1234

## Test plan
- [x] CareBridge panel + Epic Observation fingerprint match → no leaf attach + audit row
- [x] Epic panel + Epic Observation fingerprint match → leaf attached + dedup audit
- [x] Existing lab fingerprint dedup + source_system tests still pass

Closes #1234
Refs: #1200, #1183, #1190

### DIFF
--- a/services/epic-connector/src/__tests__/persistence-source-guard.test.ts
+++ b/services/epic-connector/src/__tests__/persistence-source-guard.test.ts
@@ -48,8 +48,16 @@ vi.mock("@carebridge/db-schema", () => ({
     loinc_code: "loinc_code",
     recorded_at: "recorded_at",
   },
-  labPanels: {},
-  labResults: {},
+  labPanels: {
+    id: "id",
+    patient_id: "patient_id",
+    panel_name: "panel_name",
+    collected_at: "collected_at",
+  },
+  labResults: {
+    id: "id",
+    panel_id: "panel_id",
+  },
   diagnoses: {
     id: "id",
     patient_id: "patient_id",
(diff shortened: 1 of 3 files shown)


blamechris:
## Code Review: PR #1240

### Summary
Extends the `source_system` fingerprint guard (#1200) from the five originally-wired persisters (encounters/diagnoses/allergies/medications/vitals) to the lab branch of `persistObservation`. When the panel-fingerprint lookup hits a `lab_panels` row whose `source_system != 'epic'` (or `null`, defensively treated as non-epic), the writer now: (a) skips the `lab_results` leaf attach so an Epic-sourced result cannot silently grow under a CareBridge-owned panel, (b) still upserts the `fhir_resources` mapping so subsequent syncs round-trip to the same panel and re-trip the guard, and (c) emits an `epic_sync_source_conflict` audit row carrying provenance (panel id, existing source_system, Epic FHIR id, resolution).

This is the largest PR in the cycle 8 marathon and the most clinically load-bearing of the lot — it plugs the last lab-side data-ownership hole #1234 called out.

### Strengths
- **Clinical-safety reasoning is sound.** The skip-attach behaviour matches the policy the parent-row writers have used since #1200 — CareBridge-originated clinical data is never overwritten/extended by Epic on a fingerprint collision. No PHI overwrite path was introduced.
- **Round-2 idempotency rationale is correct and verified.** The mapping intentionally points at the `panel_id` (not a leaf, because none was inserted). On a subsequent sync, control lands in the `mapping.internalId` branch at `persistence.ts:1266-1289`, which runs `UPDATE labResults WHERE id = <panel_id>` — that matches zero rows (panel ids are not leaf ids), so the UPDATE is a structural no-op and the guard re-fires every sync until the conflict is reconciled. The lengthy inline comment at `persistence.ts:1301-1317` accurately explains this; no further plumbing is needed.
- **The `null` source_system case is handled correctly.** `isEpicOwned` returns false for null, and the third test (`source_system=null → treated as non-epic, skip attach`) pins that policy down so a pre-#1190 lab_panels row defaults to the safer skip path.
- **`logFingerprintSourceConflict` payload is well-formed** — resource_type=Observation, resource_id=panel internalId, `reason="fingerprint_source_system_conflict"`, `existing_source` carries the actual value, `resolution="keep_carebridge_value"`. Symmetric with the vital/medication/allergy/diagnosis/encounter call sites.
- **Test fixture hygiene.** The three pre-existing `persistence.test.ts` fixtures tagged `source_system: 'epic'` with inline `#1234` comments are exactly what's needed to keep those tests pinned to the Epic-Epic dedup-merge path. Without the tag the panel would now trip the new guard and the dedup tests would fail for unrelated reasons.
- **Header doc updated** to reflect six guarded resource tables (was five).
- **Negative-path coverage.** The Epic-panel test explicitly asserts `findAuditInsert("epic_sync_source_conflict")).toBeUndefined()` and the CareBridge-panel test asserts `findAuditInsert("epic_sync_dedup_match")).toBeUndefined()`, so the two audit codes can't drift into each other on accident.

### Issues Found

#### Critical (Must Fix)
None.

None — the change is narrowly scoped, well-commented, well-tested.

#### Nitpicks (Optional)
None worth filing.

None — the procedures-persister half of #1234 is conditional on Procedure sync existing, which it does not (no `persistProcedure` or `epicProcedureToRow` in the connector). Per #1234's acceptance criteria that bullet only fires "if/when Procedure sync is added in a future PR" and remains tracked.

### Architecture Notes
Follows the established #1200 / #1183 guard pattern verbatim: same `isEpicOwned` predicate, same `upsertMapping` + `logFingerprintSourceConflict` + early return with `conflict: "source_system_conflict"`. No new abstraction introduced. The lab branch was the last fingerprint-hit path in `persistObservation` (and in the whole connector, for the resources that have `source_system`) that lacked the guard.

### Verdict

Local verification: typecheck clean; the full `persistence*` suite passes (49 tests, plus 4 int-tests skipped without DB). 15/15 source-guard tests pass including the 3 new lab cases. CI is green (Build, Test, Typecheck all pass on the latest commit).

PR #1241, opened 2026-05-27
feat(gateway): gate CPT emission behind FHIR_CPT_EMISSION_ENABLED flag

### LINKED ISSUE #939: fhir: compliance review — AMA CPT licensing for FHIR bundle export
## Context

PR #922 (https://github.com/blamechris/carebridge/pull/922) emits CPT codes on `Procedure.code` using the AMA CPT system URL (`http://www.ama-assn.org/go/cpt`). This is FHIR-correct, but the American Medical Association licenses CPT code usage — distributing CPT codes in a FHIR bundle to an external recipient may require an AMA CPT license agreement (end-user agreement, volume license, or similar), depending on the deployment model.

This is a compliance / legal question, not a code defect. Raising it so it's tracked before we ship the FHIR export to any external partner.

## Asks

- [ ] Confirm with CareBridge's legal/compliance whether the current CPT emission path requires an AMA license
- [ ] Document the license status in `docs/` (e.g. `docs/fhir-licensing.md`) alongside the existing HIPAA retention doc
- [ ] If a license is required and not yet obtained: gate CPT emission behind a config flag until the license is in place

## Files

- `services/fhir-gateway/src/generators/procedure.ts` — CPT emission
- `docs/` — new licensing doc

Discovered during review of PR #922.

blamechris:
## Summary

Closes #939's autonomous portion. The remaining work is a legal/compliance decision, not code.

AMA licenses CPT codes. PR #922 introduced emission of CPT codings on `Procedure.code` via the AMA system URL, which is FHIR-correct but may require an AMA CPT license depending on the deployment model (recipients, volume, commercial relationship).

This PR ships the **fail-safe default** so the question can be answered without time pressure: CPT coding is now gated behind `FHIR_CPT_EMISSION_ENABLED`, which defaults to off. `Procedure.code` still carries a useful free-text fallback (`code.text = procedure.name`) so downstream consumers can read the procedure even without the licensed coding.

## Changes

- `services/fhir-gateway/src/generators/procedure.ts` — read `process.env.FHIR_CPT_EMISSION_ENABLED` at call time. Only the literal string `"true"` enables emission; anything else (unset, `"false"`, `"1"`, `"yes"`) keeps the gate closed.
- `services/fhir-gateway/src/__tests__/procedure.test.ts` — 4 new tests covering:
  - flag unset (default) → no CPT coding even when `cpt_code` is set
  - flag = `"false"` → no CPT coding
  - flag = `"1"` → no CPT coding (only literal `"true"` enables)
  - flag = `"true"` → CPT coding emitted as before
  - Existing CPT tests wrapped in setup that explicitly enables the flag.
- `docs/fhir-licensing.md` — documents the gate, the open AMA license question, what needs to happen before flipping the flag, and where the tests live. Pattern matches `docs/hipaa-retention.md` and `docs/fhir-interop.md`.
- `.env.example` — adds `FHIR_CPT_EMISSION_ENABLED` (commented out) under the existing FHIR section.

## What's NOT in this PR

- The legal answer to "does CareBridge need an AMA CPT license?" — that decision still needs a human / legal review. This PR is the conservative scaffolding that lets that decision happen out-of-band without ICD-10 emission, US Core profile conformance, or anything else regressing.
- Once legal clarifies, flipping the flag in production is a one-line env change. No code redeploy required.

## Test plan

- [x] `pnpm --filter @carebridge/fhir-gateway test -- procedure.test` — 27/27 passing (was 22, added 5 incl. parameterised setup)
- [x] Confirmed pre-existing typecheck/test-loader failures (`@lhncbc/ucum-lhc` missing, `body_height` / `bmi` / `head_circumference` / `npi` / `nucc_code` references) exist on `main` and are NOT introduced by this PR — verified by checking out `main` versions of the touched files and re-running.
- [x] `pnpm lint` — clean (only pre-existing clinician-portal warnings)

Refs #939

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -64,3 +64,11 @@ NODE_ENV=development
 # services/epic-connector/README.md for details.
 # EPIC_OBSERVATION_CATEGORIES=vital-signs,laboratory
 # EPIC_MEDICATION_REQUEST_STATUS=active
+
+# FHIR — AMA CPT emission gate (#939)
+# CPT codes are AMA-licensed. Distributing them in outbound FHIR bundles
+# to an external recipient may require an AMA CPT license. This flag
+# defaults to "off" — Procedure.code falls back to the free-text
+# procedure name. Set to the literal string "true" only after the
+# license question is resolved. See docs/fhir-licensing.md.
+# FHIR_CPT_EMISSION_ENABLED=true
--- a/docs/fhir-licensing.md
+++ b/docs/fhir-licensing.md
@@ -1,0 +1,80 @@
+# FHIR Code System Licensing
+
+This document tracks third-party code-system licenses that gate what
+CareBridge is allowed to distribute in outbound FHIR bundles. The
+practical concern is that several widely used FHIR code systems
+(CPT, SNOMED CT, LOINC under some conditions) are licensed by their
+publishers, and including those codings in a FHIR resource sent to
+an external recipient may constitute redistribution.
+
+Each gated system has a corresponding environment flag and a default
+that fails safe (no emission until the license question is resolved).
+
+## CPT (AMA Current Procedural Terminology)
+
+### Status
+
+**Default: emission disabled.** `FHIR_CPT_EMISSION_ENABLED` defaults
+to anything other than the literal string `"true"`, which means the
+fhir-gateway omits CPT codings from `Procedure.code`.
+
+### Context
+
+`services/fhir-gateway/src/generators/procedure.ts` historically
+emitted CPT codings on `Procedure.code` using the AMA CPT system URL
+(`http://www.ama-assn.org/go/cpt`). This is FHIR-correct: the AMA is
+the canonical source for CPT and that system URL is the registered
+identifier.
+
+The American Medical Association licenses CPT code usage. Distributing
+CPT codes in a FHIR bundle to an external recipient may require an
+AMA CPT license agreement (end-user agreement, organisational
+license, or similar). The required license depends on the deployment
+model and the number / nature of recipients.
+
+CareBridge has not yet confirmed which license tier covers its
+intended FHIR export path. Until that is confirmed, emission is
+gated off by default.
+
+### The gate
+
+`services/fhir-gateway/src/generators/procedure.ts` reads
+`process.env.FHIR_CPT_EMISSION_ENABLED` at call time. CPT codings
+are only emitted when the value is exactly the string `"true"`. Any
+other value (unset, `"false"`, `"1"`, `"yes"`, etc.) keeps the gate
+closed.
+
+When the gate is closed, `Procedure.code` still emits a useful
+free-text fallback via the existing `code.text = procedure.name`
+path. Downstream consumers see a Procedure with a human-readable
+description but no licensed coding.
+
+### What needs to happen before flipping the gate
+
+1. **Confirm with CareBridge legal / compliance** whether the
+   intended FHIR export model (which recipients, what volume, what
+   commercial relationship) requires an AMA CPT license.
+2. **If a license is required:** obtain the appropriate AMA license
+   covering distribution. Record the license reference in this
+   document.
+3. **If no license is required** (e.g. recipient is a covered entity
+   with their own CPT license, or the export is intra-organisational
+   only): document the basis in this file with the date and the
+   stakeholder who confirmed.
+4. **Flip the flag in the relevant deployment environments:**
+   ```
+   FHIR_CPT_EMISSION_ENABLED=true
+   ```
+
+### Tests
+
+- `services/fhir-gateway/src/__tests__/procedure.test.ts` covers
+  both the gated-off path (default) and the gated-on path. The
+  gated-off path asserts that `Procedure.code.coding` is undefined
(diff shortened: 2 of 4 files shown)


blamechris:
## Code Review: PR #1241

### Summary

Gates AMA-licensed CPT code emission in outbound FHIR `Procedure.code` behind a new `FHIR_CPT_EMISSION_ENABLED` environment flag, defaulting to off. This is the conservative, code-only half of issue #939 — it lets the licensing question be resolved out of band without holding up FHIR export work. Free-text `code.text = procedure.name` remains so the resource stays useful when the gate is closed.

The PR ships:

- A `isCptEmissionEnabled()` helper that strictly compares `process.env.FHIR_CPT_EMISSION_ENABLED === "true"` (no truthy-coercion).
- 4 new tests covering flag-unset, flag-`"false"`, flag-`"1"` (non-true truthy), flag-`"true"`, plus the existing CPT-emission tests now scoped under a `beforeEach` that enables the flag.
- A new `docs/fhir-licensing.md` documenting the gate, the open AMA license question, and the steps required to flip it, in the same shape as `docs/hipaa-retention.md` and `docs/fhir-interop.md`.
- An `.env.example` entry under the existing FHIR section.

### Strengths

- **Fail-safe default is genuinely fail-safe.** The strict `=== "true"` check is the right call — `Boolean(value)` or truthy coercion would silently enable emission on `"false"`, `"0"`, `"no"`, and the `"1"` test explicitly nails that down. Good discipline.
- **Read at call time, not module init.** This is what makes the tests cleanly mutate the env without re-importing the module, and what makes a future runtime-config rollout (without restart) possible.
- **Documentation matches the repo pattern.** `docs/fhir-licensing.md` mirrors the structure of the existing `hipaa-retention.md` doc — Status → Context → Gate → What needs to happen → Tests → Related. Forward-looking too: explicitly calls out SNOMED CT and LOINC as future licensing concerns.
- **Free-text fallback preserved.** `code.text = procedure.name` keeps the FHIR resource human-readable for downstream consumers when the licensed coding is dropped. The gated-off tests assert both the absence of the coding and the presence of the text.
- **Test isolation is correct.** The `previous = process.env.X / afterEach restore` pattern handles the case where another suite has already set the env var. Cleanly avoids cross-test pollution.
- **JSDoc on `procedure.ts` updated** to reference `#939` and `docs/fhir-licensing.md`, so the next reader of the generator finds the rationale without git-blame archaeology.

### Issues Found

#### Critical (Must Fix)

| _(none)_ | | | |

| _(none in-scope; deferred items below)_ | | | |

#### Nitpicks (Optional)

- `services/fhir-gateway/src/__tests__/procedure.test.ts:50,89` — Two `beforeEach/afterEach` blocks with the same env save/restore shape. Could be factored into a `withCptFlag(value: string | undefined, fn: () => void)` helper if you find yourself repeating the pattern for SNOMED/LOINC later. Not worth changing now — pure repetition cost is low and the explicitness aids reading.

| SNOMED CT and LOINC are also licensed code systems that the new `docs/fhir-licensing.md` calls out, but they're emitted unconditionally from `allergy-intolerance.ts`, `condition.ts`, `medication-request.ts`, `medication-statement.ts`, and `observation.ts`. Same gate pattern likely needed once legal confirms the model. | [#1250](https://github.com/blamechris/carebridge/issues/1250) | Out of scope — this PR is explicitly the CPT-only fail-safe. Mirroring the gate for SNOMED CT / LOINC needs its own legal review and own test coverage per generator. |
| Add a one-time-per-process operator log breadcrumb when CPT suppression is active, so misconfig (`"True"` / `"yes"` / `"1"` instead of `"true"`) is diagnosable from log streams. | [#1251](https://github.com/blamechris/carebridge/issues/1251) | Out of scope — silent suppression matches the conservative default. The breadcrumb is a UX improvement for operators, not a correctness concern. |

### Architecture Notes

This change fits the repo's "fail-safe defaults with explicit opt-in" posture for compliance-gated behaviour — same shape as `CAREBRIDGE_DEV_AUTH` (off unless explicitly `"true"`), and consistent with the `EPIC_OBSERVATION_CATEGORIES` / `EPIC_MEDICATION_REQUEST_STATUS` pattern of reading from env at call time with documented overrides.

It correctly stays inside `services/fhir-gateway` — no cross-service HTTP, no DB-schema change, no event-queue interaction. The only public-API change is that `Procedure.code.coding` may be `undefined` for downstream consumers; that's already a valid state for cpt-less rows so no consumer contract is broken.

One thing worth noting for whoever picks up the SNOMED/LOINC follow-up: the gate logic is currently a module-private helper. If you end up needing three gates (CPT, SNOMED, LOINC) read across five generators, this should probably be a small shared `licensing-gates.ts` module in `services/fhir-gateway/src/` rather than three copy-pastes of `process.env.X === "true"`. Premature now; flag for later.

### Verdict

Clean, well-scoped change. The fail-safe default + the docs + the open license question being explicit (rather than hidden in code review history) is exactly the right way to handle a compliance scaffold. CI is green (Typecheck, Build, Test all SUCCESS), no breaking changes, no architecture violations. The deferred items are genuine follow-ups, not blockers.

Issue #916, opened 2026-04-19
smoke-test: post-Wave-8 end-to-end verification (36 PRs, clinical workflow + caregiver + appointments + notifications + audit IP)


blamechris:
# Smoke test checklist — post-Wave-8

**Context.** Between 2026-04-18 and 2026-04-19, 36 PRs merged across 8 waves covering: clinical-workflow (cosign/amend/version history, care team CRUD), patient-portal (proxy/caregiver, appointment booking, reminders, mobile responsive), AI oversight expansion (9 new rules including Beers/pediatric, contraindications, HL7 panic severity), notifications (audience routing, escalation), access-scope enforcement, allergy override audit, HIPAA IP-capture, and a review-debt sweep. This issue verifies the entire stack boots, talks end-to-end, and the headline flows work.

**How to use this issue.** Each step has a checkbox. Check it when the step passes as described. If it fails, **reply in a comment using the template at the bottom** and leave the box unchecked — I'll fix the root cause and ping you to retry.

**Estimated total time:** 45–90 minutes if everything works, longer if blockers surface.

**Working directory for ALL commands below (unless stated otherwise):** `/Users/blamechris/Projects/carebridge`

---

### 0.1 Sync local main to remote
- [ ] Pass

**Working dir:** `/Users/blamechris/Projects/carebridge`

```bash
git status                    # should be clean — NO uncommitted changes
git checkout main
git pull origin main
git log --oneline -5          # confirm last 5 commits include the Wave 8 merges
```

**Expect:** Output includes PRs like `Merge pull request #915`, `#914`, `#912`, `#911`, `#910` in the last 20 commits. Working tree clean.

**If fails:**
- "uncommitted changes": run `git stash` first (you can `git stash pop` after). If those changes are from prior worktree spill, investigate before stashing.
- "behind origin": `git pull` resolves it.
- Merge conflict on pull: unusual — means you made local commits to main. Let me know.

---

### 0.2 Tool versions
- [ ] Pass

```bash
node --version                # expect v20.x or v22.x (monorepo tested range)
pnpm --version                # expect 9.15.4 (matches packageManager in package.json)
docker --version              # any modern version
docker compose version        # v2.x
```

**Expect:** `pnpm 9.15.4`. Other versions need to be "in the ballpark."

**If fails:** `corepack enable && corepack prepare pnpm@9.15.4 --activate` to fix pnpm. Install the right Node from your version manager (nvm/fnm/volta).

---

### 0.3 Env vars populated
[...]

PR #1244, opened 2026-05-28
fix(ai): cast reviewJobs.created_at to timestamptz in in-flight dedup probe

blamechris:
## Summary

**Production-blocking regression in the AI oversight worker.** Every fresh review job currently fails on the idempotency probe with:

```
operator does not exist: text >= timestamp with time zone
```

The pipeline short-circuits before any rule evaluation or LLM call, so **no clinical flags are ever generated**. Surfaced during the post-Wave-8 smoke test (#916).

## Root cause

`services/ai-oversight/src/services/review-service.ts:137` builds the dedup cutoff as:

```ts
const inFlightCutoff = sql`NOW() - ${IN_FLIGHT_WINDOW_SEC} * interval '1 second'`;
// ...
gte(reviewJobs.created_at, inFlightCutoff),  // line 153
```

- `reviewJobs.created_at` is a **`text`** column storing ISO 8601 Z-form strings (per `CLAUDE.md` convention: "All dates as ISO 8601 strings").
- `inFlightCutoff` evaluates to a **`timestamp with time zone`** in PostgreSQL.
- PostgreSQL has no operator `text >= timestamptz` → throws on every probe call.

Why it wasn't caught earlier:
- `review-service-db-clock-idempotency.test.ts` mocks drizzle's `sql` helper, so it never executes the actual SQL.
- `review-idempotency.test.ts` mocks the entire DB layer; the `gte` mock doesn't validate the runtime types.
- No integration test exercises the dedup probe against real PostgreSQL.

## Fix

Wrap the column reference in an inline `::timestamptz` cast so PostgreSQL has a valid `timestamptz >= timestamptz` comparison:

```ts
gte(sql`${reviewJobs.created_at}::timestamptz`, inFlightCutoff),
```

This preserves the design intent of PR #608 (DB-clock NOW() for worker/DB clock-skew safety) — only the column side is cast. The cast is safe because every `created_at` value is written via `new Date().toISOString()` which always emits parseable ISO 8601 Z-form.

## Smoke-test verification

After applying the fix and letting `tsx watch` hot-reload the worker, the previously-stuck job processed cleanly:

```
[review-worker] Processing job fb6e0975-... — event: note.saved for patient 95d00e2c****
[review-worker] Job fb6e0975-... completed in 291ms
```

A `clinical_flags` row appeared for the patient (placeholder "AI review unavailable" because the smoke env has no real `ANTHROPIC_API_KEY` — the LLM path 401s and falls back to the deterministic-only path, which is the expected dev behavior).

## Regression test

`review-idempotency.test.ts` gets a new test under the existing `processReviewJob — idempotency probe predicate shape` suite:

```ts
it("wraps the in-flight cutoff column side in a sql template (timestamptz cast)", async () => {
  // ... captures gte() calls, asserts first arg is sql-tagged (cast),
  // not the bare drizzle column reference.
});
```

This catches a future regression where someone reintroduces a bare column reference on the text-vs-timestamptz comparison path.

## Test plan

- [x] `pnpm --filter @carebridge/ai-oversight test -- review-idempotency` — 11/11 (10 existing + 1 new)
- [x] `pnpm --filter @carebridge/ai-oversight test -- review-service-db-clock-idempotency` — 7/7 (unchanged; the cutoff expression itself is unchanged)
- [x] `pnpm --filter @carebridge/ai-oversight test` — 939/939
- [x] `pnpm typecheck` — 38/38 tasks clean
- [x] Live smoke verification — job that had failed 7/8 retries completed in 291ms after worker hot-reload

## Related follow-ups (separate issues, not in this PR)

1. **Seed FK violation on a clean DB** — `tooling/seed/index.ts:94` sets `patient_id: dvtPatientId` on the patient-user insert before the patient row exists, violating `users_patient_id_fkey`. The seed has a deferred linking step at line 253 that already handles this; the early set is dead code. Will file separately.
2. **DVT scenario doesn't fire** — even with the SQL fix and the seed data in place, the deterministic `ONCO-VTE-NEURO-001` rule didn't generate a flag for Margaret on `note.saved`. Needs separate investigation.
3. **Smoke-test checklist vs seed mismatch** — Phase 4.1 of #916 expects "DVT flag already firing from seed" but the seed doesn't enqueue review events. The checklist needs an update OR the seed needs to publish events.

Refs #916

### DIFF
--- a/services/ai-oversight/src/__tests__/review-idempotency.test.ts
+++ b/services/ai-oversight/src/__tests__/review-idempotency.test.ts
@@ -267,4 +267,36 @@ describe("processReviewJob — idempotency probe predicate shape", () => {
 
     expect(orMock).toHaveBeenCalled();
   });
+
+  // Regression — the in-flight cutoff is `NOW() - interval` (a
+  // `timestamp with time zone`), but `review_jobs.created_at` is a
+  // `text` column. Without an explicit cast on the column side, the
+  // query errors with `operator does not exist: text >= timestamp with
+  // time zone` and silently bricks every review job. Found during the
+  // post-Wave-8 smoke test (#916). The fix wraps the column reference
+  // in a `sql` template (the cast); this asserts the column side of
+  // the in-flight gte call is no longer the bare drizzle column.
+  it("wraps the in-flight cutoff column side in a sql template (timestamptz cast)", async () => {
+    const drizzle = await import("drizzle-orm");
+    const gteMock = vi.mocked(drizzle.gte);
+
+    try {
+      await processReviewJob(makeEvent({ id: "evt-probe-cutoff-shape" }));
+    } catch {
+      // Ignore downstream errors — we only care about the probe call.
+    }
+
+    // The dedup probe calls gte(<column-or-cast>, inFlightCutoff). The
+    // pre-fix code passed `reviewJobs.created_at` directly (a string
+    // identifier under our mock). The fix passes a sql template
+    // (our mock returns `{ __sql: true }`). Assert that at least one
+    // gte call's first argument is the sql-tagged object.
+    const sqlTaggedColumnCalls = gteMock.mock.calls.filter(
+      ([col]) =>
+        typeof col === "object" &&
+        col !== null &&
+        (col as { __sql?: boolean }).__sql === true,
+    );
+    expect(sqlTaggedColumnCalls.length).toBeGreaterThanOrEqual(1);
+  });
 });
--- a/services/ai-oversight/src/services/review-service.ts
+++ b/services/ai-oversight/src/services/review-service.ts
@@ -150,7 +150,17 @@ export async function processReviewJob(event: ClinicalEvent): Promise<void> {
           inArray(reviewJobs.status, TERMINAL_REVIEW_STATUSES as string[]),
           and(
             eq(reviewJobs.status, "processing"),
-            gte(reviewJobs.created_at, inFlightCutoff),
+            // `reviewJobs.created_at` is a `text` column; the cutoff
+            // above is `timestamp with time zone`. Cast the column at
+            // query time so PostgreSQL has a `timestamptz >= timestamptz`
+            // operator. Without this, every probe call errors with
+            // `operator does not exist: text >= timestamp with time zone`
+            // and the dedup short-circuit silently bricks every review
+            // job. Found during the post-Wave-8 smoke test (#916). PR
+            // #608 deliberately moved to DB-clock NOW() for worker/DB
+            // clock-skew safety — we keep that property and only cast
+            // the column side.
+            gte(sql`${reviewJobs.created_at}::timestamptz`, inFlightCutoff),
           ),
         ),
       ),


blamechris:
## Code Review: PR #1244

### Summary

One-line SQL fix for a production-blocking bug: the in-flight idempotency probe compared `reviewJobs.created_at` (a `text` column) against `inFlightCutoff` (a `timestamp with time zone` expression), which PostgreSQL rejected with `operator does not exist: text >= timestamp with time zone`. Every BullMQ review job was failing on the probe, short-circuiting the pipeline before any flag generation. The fix wraps the column side in `sql\`${reviewJobs.created_at}::timestamptz\`` so PostgreSQL has a valid `timestamptz >= timestamptz` comparison, while preserving the DB-clock NOW() pattern from PR #608 (worker/DB clock-skew safety).

### Strengths

- **Surgical fix.** Only the column side is cast; the cutoff side stays as the SQL `NOW() - interval`, preserving the worker/DB clock-skew invariant from PR #608. No other call sites or query shapes change.
- **Cast safety is sound.** Every `reviewJobs.created_at` value is written via `new Date().toISOString()` (lines 181 and 612 of `review-service.ts`), which always emits Z-form ISO 8601 — Postgres `text::timestamptz` parses this cleanly.
- **No index regression.** There is no btree index on `reviewJobs.created_at` (only `idx_review_jobs_trigger_event` on `(trigger_event_id, status)`), so the column-side cast doesn't disable an index that would otherwise be used.
- **Other comparisons in the file were correctly inspected.** The `gte/lte/gt` calls in `buildPatientContextForRules` (lines 677–731) compare text columns against JS-string cutoffs (`eventAt`, `labCutoff`) — text-vs-text lexicographic compare, which is correct for the ISO Z-form values the platform writes. No collateral fixes needed there.
- **Excellent inline comment** at the call site documenting *why* the cast exists and the production incident that drove it. Future readers won't be tempted to simplify it away.

### Issues Found

#### Critical (Must Fix)

None.

| `services/ai-oversight/src/__tests__/review-idempotency.test.ts` | 292–298 | The assertion is `sqlTaggedColumnCalls.length >= 1`, i.e. "at least one gte call has a sql-tagged first arg." Today this catches a line-153 regression, but only because no other `gte` call wraps its column side in `sql\`\``. If a future change to `buildPatientContextForRules` (e.g. casting `labResults.created_at`) introduces a sql-tagged gte call, this test would pass even if the in-flight cutoff regressed. Consider tightening — either by also asserting both args of that specific call are sql-tagged, OR by adding the cast assertion to `review-service-db-clock-idempotency.test.ts` where the sql template strings/values are captured directly (and you can grep for the rendered `::timestamptz` substring). Tracked as #1254. |
| `services/ai-oversight/src/services/review-service.ts` | — | No real-PG integration test exercises the dedup probe. This is the structural reason the bug shipped (the PR body explicitly acknowledges it). The new mocked regression test cannot catch a future Drizzle column-type/PG-type mismatch, only the specific shape that triggered this incident. Tracked as #1252. |

#### Nitpicks (Optional)

- The new inline comment block on lines 153–162 is verbose (10 lines for a 1-line fix), but the verbosity is justified by the incident severity — future readers will benefit from the breadcrumb to #916. Keep as-is.

| Add a real-PG integration test for the dedup probe | [#1252](https://github.com/blamechris/carebridge/issues/1252) | Out of scope for a production hotfix — integration infrastructure work is larger than the fix itself. |
| Tighten the new regression test so the cast is bound to the in-flight predicate specifically | [#1254](https://github.com/blamechris/carebridge/issues/1254) | Current test is sufficient for the immediate regression; precision improvement can land separately. |
| Seed FK violation on patient-user insert (called out by author) | [#1256](https://github.com/blamechris/carebridge/issues/1256) | Separate concern surfaced during the same smoke-test session. |
| DVT scenario (ONCO-VTE-NEURO-001) doesn't fire for Margaret (called out by author) | [#1257](https://github.com/blamechris/carebridge/issues/1257) | Needs independent investigation; not a regression introduced by this PR. |
| Phase 4.1 checklist (#916) vs seed mismatch (called out by author) | [#1258](https://github.com/blamechris/carebridge/issues/1258) | Documentation/scope alignment issue distinct from the SQL fix. |

### Architecture Notes

The fix preserves the architectural intent of PR #608: both sides of the in-flight freshness comparison resolve in PostgreSQL's clock, so a fast or slow worker JS clock cannot misclassify a live `processing` row as an orphan (or vice versa). The cast is the minimum delta needed to make the operator valid; no rethink of the probe semantics is required.

One mild architectural smell remains, but out of scope for this PR: `reviewJobs.created_at` is stored as `text` (per the project-wide "All dates as ISO 8601 strings" convention in CLAUDE.md), yet the rest of the platform increasingly compares these values against PG `timestamptz` expressions. Casting at query time is workable but creates a pattern others may mimic incorrectly. A future migration to `timestamptz` columns for `review_jobs` time fields would be cleaner, but is correctly out of scope here.

### Verdict

This is a production-blocking hotfix with a tightly-scoped, low-risk change. CI is green (Build, Test, Typecheck all pass). The author has already filed the three sibling concerns surfaced during smoke testing (now tracked as #1256, #1257, #1258). The two test-improvement suggestions (#1252, #1254) are non-blocking and appropriate for follow-up PRs.

Ship it.

PR #1248, opened 2026-05-28
fix(db): drop premature patient_id on demo user (#1245)

### LINKED ISSUE #1245: fix(seed): users_patient_id_fkey violation on clean DB
## Summary

`tooling/seed/index.ts:94` sets `patient_id: dvtPatientId` on the `patient@carebridge.dev` user row, but `dvtPatientId` doesn't exist in the `patients` table at that point (the patient row is inserted at line 103, after the users insert). On a fresh DB this violates `users_patient_id_fkey` and the seed errors out.

```
Key (patient_id)=(82c49e88-...) is not present in table "patients".
constraint_name: users_patient_id_fkey
```

## Why it didn't surface earlier

When the DB had a prior `patient@carebridge.dev` user from an earlier seed run, the `.onConflictDoNothing()` in the users insert silently no-op'd the row (conflict on email unique index), skipping the FK check. So the bug only fires the very first time `pnpm db:seed` runs against a truly clean DB — or after a `TRUNCATE users` reset like the smoke test did.

## Fix

The seed already has a deferred linking step at line 253:

```ts
await db.update(schema.users)
  .set({ patient_id: dvtPatientId })
  .where(eq(schema.users.id, patientUser));
```

The early `patient_id: dvtPatientId` on line 94 is dead — the deferred update handles it correctly after the patient row exists. Drop the early set.

## Repro

```bash
# From a clean DB
docker exec carebridge-postgres-1 psql -U carebridge -d carebridge \
  -c "TRUNCATE users, patients CASCADE"
pnpm db:seed   # fails with users_patient_id_fkey
```

Found during the post-Wave-8 smoke test (#916).

blamechris:
## Summary

`tooling/seed/index.ts` set `patient_id: dvtPatientId` on the `patient@carebridge.dev` user row before the corresponding `patients` row was inserted. On a clean DB this violated `users_patient_id_fkey` and the seed aborted; on a re-seed, `.onConflictDoNothing()` silently swallowed the row and masked the bug — discovered during the post-Wave-8 smoke test (#916) when the smoke-test workflow ran against a freshly-truncated DB.

The deferred update further down in the seed script (around line 253) already sets `patient_id` once the `patients` row exists, so dropping the early set is the complete fix.

## Why no test

The seed package has no test infrastructure — it's a one-off TypeScript script. Adding a real-PG integration test for a single seed run is out of scope for a 1-line fix. The added comment guards against re-introduction (a future contributor would otherwise look at a "Demo Patient" user with no `patient_id` and assume it was an oversight).

## Repro (before fix)

```bash
docker exec carebridge-postgres-1 psql -U carebridge -d carebridge \
  -c "TRUNCATE users, patients CASCADE"
pnpm db:seed   # fails: users_patient_id_fkey
```

After this fix, the same sequence succeeds.

Closes #1245.

## Test plan

- [x] Verified `pnpm typecheck --filter=@carebridge/seed` passes
- [ ] CI green
- [ ] Manual verification: `TRUNCATE users, patients CASCADE; pnpm db:seed` succeeds (already done during smoke test pre-merge)

### DIFF
--- a/tooling/seed/index.ts
+++ b/tooling/seed/index.ts
@@ -91,7 +91,9 @@ async function seed() {
       password_hash: devPasswordHash,
       name: "Demo Patient",
       role: "patient",
-      patient_id: dvtPatientId,
+      // patient_id is set by the deferred update below, after the patients
+      // row exists. Setting it here violates users_patient_id_fkey on a
+      // clean DB (#1245).
       is_active: true,
       created_at: now(),
       updated_at: now(),


blamechris:
## Code Review: PR #1248

### Summary
Drops the premature `patient_id: dvtPatientId` set on the `patient@carebridge.dev` user row at `tooling/seed/index.ts:94`. The FK target (`patients.id = dvtPatientId`) didn't exist yet at that point in the script, so on a clean DB the insert violated `users_patient_id_fkey`. On a re-seed the bug was masked because `.onConflictDoNothing()` silently no-op'd the conflicting email row. The deferred `db.update(schema.users).set({ patient_id: dvtPatientId })` at line 253 (now 255 after the comment) does the linking correctly once the `patients` row exists.

### Strengths
- Minimal, surgical 1-line behavior change — exactly what a clean-DB seed regression needs.
- The added 3-line comment cites the issue number (`#1245`) inline, so a future contributor inspecting the user-row literal and wondering "why is `Demo Patient` missing a `patient_id`?" gets the answer without `git blame`.
- The PR body documents the masking mechanism (`onConflictDoNothing` swallowing the row), the repro steps, and the smoke-test trail (#916) that surfaced it. Good paper trail.
- `Closes #1245` is properly wired (not `Refs` — issue will auto-close on merge).

### Issues Found

#### Critical (Must Fix)
_None._

_None._ Verified the deferred-update site at lines 253-255 still exists, still references `dvtPatientId` and `patientUser` correctly, and is reached unconditionally before `client.end()`. Verified no other premature FK references in the file — every other `patient_id: dvtPatientId` / `patient_id: patient2Id` usage occurs after the corresponding `patients` row insert.

#### Nitpicks (Optional)
_None._

_None._ The PR body's "Why no test" rationale is sound: the seed package has no test harness, standing one up for a single-script integration test isn't warranted by a 1-line fix, and the explanatory comment + `#1245` issue trail are sufficient regression guards. If the seed grows further FK-ordering footguns later, that's the trigger to add a smoke-test-style integration check — not this PR.

### Architecture Notes
No architectural surface touched. This is purely a dev-tooling seed-script ordering fix; no schema change, no migration, no service or shared-package impact. The fix preserves the existing two-phase pattern (insert users with NULL `patient_id`, then `UPDATE` after `patients` row exists) which is the right shape for a self-referential FK like `users.patient_id -> patients.id`.

### Verdict

Correctly diagnosed, minimally fixed, well-documented. The comment-in-code + issue-in-PR-body combination is the right level of guard for a script with no test infrastructure.

PR #1249, opened 2026-05-28
fix(ai): enrich note.saved payload with chief_complaint/HPI/new_symptoms (#1246)

blamechris:
## Summary

`ai-oversight`'s `extractSymptoms()` populates `ctx.new_symptoms` from `event.data.chief_complaint`, `event.data.subjective`, and `event.data.new_symptoms`. The note-service publishers (`createNote`, `updateNote`) were only emitting `{ resourceId, [version] }`, so `ctx.new_symptoms` always came back empty. The `ONCO-VTE-NEURO-001` rule's `hasNeuroSymptom` check then returned `false` — even on a SOAP draft that explicitly listed `headache` + `numbness` + `vision changes` in the New Symptoms checklist for a cancer+VTE patient.

This was found in the post-Wave-8 smoke test (#916), reported as #1246. The SQL regression that previously bricked every review job (#1244) was masking the symptom-extraction gap; once the worker started running cleanly, the empty-symptoms problem became visible.

## Fix

Add a publisher-side helper `extractSymptomFieldsFromSections` in `services/clinical-notes/src/services/note-service.ts` that walks the structured `NoteSection[]` and pulls:

- `chief_complaint` — the first non-empty `chief_complaint` field across all sections (SOAP nests it inside `subjective`; H&P puts it at the top level).
- `subjective` — first non-empty `history_of_present_illness` or `hpi` field.
- `new_symptoms` — concatenation of any `new_symptoms` or `ros` array values.

The result is spread onto `event.data` alongside `resourceId` for both the `createNote` and `updateNote` publish points. Empty / null fields are omitted so the event stays sparse for notes that genuinely lack symptom content.

## Why publisher-side, not consumer-side

Alternatives considered:
- **Worker fetches the note row by `resourceId`.** Cleaner separation (event stays minimal), but pulls more shape responsibility into `ai-oversight` and adds a DB roundtrip per event. Worth considering once we have integration test coverage to validate the SQL.
- **Spread `sections` whole onto `event.data`.** Includes objective/assessment/plan content unnecessarily and grows queue payload size — most fields are PHI that have no rule consumers.

The narrow field-key extractor keeps the payload small and avoids broadcasting non-symptom PHI through BullMQ.

## Tests

Added 4 regression tests in `services/clinical-notes/src/__tests__/note-service.test.ts`:
- `enriches the note.saved payload with chief_complaint, HPI, and new_symptoms (#1246)` — primary regression coverage with realistic SOAP fixture.
[...]

PR #1266, opened 2026-05-28
fix(ai): enrich note.signed + note.amended payloads (#1259, #1260)

### LINKED ISSUE #1259: clinical-notes: enrich note.signed payload with chief_complaint/HPI/new_symptoms
## Context

Identified during review of PR #1249 (the fix for #1246).

PR #1249 enriches `note.saved` events emitted by `createNote` and `updateNote` so that `ai-oversight`'s `extractSymptoms()` can read `chief_complaint`, `subjective` (HPI), and `new_symptoms` from `event.data`. The consumer (`services/ai-oversight/src/services/review-service.ts:1017-1023`) ALSO handles `note.signed`:

```ts
if (
  (event.type === "note.saved" || event.type === "note.signed") &&
  event.data.subjective &&
  typeof event.data.subjective === "string"
) {
  symptoms.push(event.data.subjective);
}
```

…but `signNote` in `services/clinical-notes/src/services/note-service.ts:249-256` still emits:

```ts
data: { resourceId: noteId, signedBy },
```

Result: a clinician who composes a draft and signs it without an intervening `updateNote` cycle (the natural flow for a quick visit) triggers a `note.signed` event with no symptom content, so the rule still won't fire from that path.

## Description

Enrich the `note.signed` emit in `signNote` using the same `extractSymptomFieldsFromSections(existing.sections)` helper introduced in PR #1249. `existing` is already loaded on line 222 for archival purposes, so there's no additional DB roundtrip.

## Original Review Comment

> `signNote` emits `note.signed` with bare `{ resourceId, signedBy }`. `extractSymptoms()` explicitly handles `note.signed` (`review-service.ts:1018`) — so the consumer is waiting for enrichment that never arrives. In real workflows, signing IS the moment a symptom-bearing note enters the record; an unenriched `note.signed` means the rule still won't fire when a clinician composes a draft and signs it without a prior `updateNote` cycle.

## Acceptance Criteria

- [ ] Spread `extractSymptomFieldsFromSections(existing.sections)` onto the `note.signed` event payload in `signNote`
- [ ] Regression test: signing a draft with SOAP `new_symptoms` containing `headache` produces a `note.signed` event with `event.data.new_symptoms` populated
- [ ] No change to the `note.signed` payload shape beyond the added keys (resourceId + signedBy still present)
- [ ] No new DB roundtrip — reuse the row already loaded for archival

## Refs

PR #1249, issue #1246.

### LINKED ISSUE #1260: clinical-notes + ai-oversight: enrich note.amended payload and consume it in extractSymptoms
## Context

Identified during review of PR #1249 (the fix for #1246).

PR #1249 enriches `note.saved` events. The `note.signed` enrichment is tracked separately (linked follow-up).

`amendNote` in `services/clinical-notes/src/services/note-service.ts:411-424` emits a `note.amended` event with `{ resourceId, amendedBy, reason, previousVersion, newVersion }`. The consumer's `extractSymptoms()` in `services/ai-oversight/src/services/review-service.ts:1017-1023` ONLY switches on `note.saved` and `note.signed`:

```ts
if (
  (event.type === "note.saved" || event.type === "note.signed") &&
  event.data.subjective &&
  typeof event.data.subjective === "string"
) {
  ...
}
```

Amendments are HIPAA-significant content changes — a clinician adding "new onset headache + numbness" to a previously signed note via amendment is exactly the cross-specialty signal `ONCO-VTE-NEURO-001` exists to catch. The current contract drops these.

## Description

Two coordinated changes:

1. **Publisher side** (`note-service.ts:411-424`): spread `extractSymptomFieldsFromSections(sections)` onto the `note.amended` event payload. `sections` is the new amended content (the parameter), not the archived prior version.
2. **Consumer side** (`review-service.ts:1017-1023`): add `note.amended` to the event-type guard so `event.data.subjective` is consumed on the amendment path.

## Original Review Comment

> `amendNote` emits `note.amended` with reason/version. `extractSymptoms()` does not currently handle `note.amended`, but amendments can change clinical content (and per HIPAA they may add neurological observations to a previously signed note). Symmetric enrichment + adding `note.amended` to the consumer's switch would close the loop for the amendment path.

## Acceptance Criteria

- [ ] `amendNote` event payload includes chief_complaint / subjective / new_symptoms when present in the amended sections
- [ ] `extractSymptoms()` recognises `note.amended` for both `chief_complaint` and `subjective` paths
- [ ] Regression test in `services/clinical-notes/src/__tests__/note-service.test.ts` covering amendment with added neuro symptoms
- [ ] Regression test in `services/ai-oversight/src/__tests__/review-service.test.ts` (or appropriate test file) asserting `note.amended` event with symptoms populates `ctx.new_symptoms`
- [ ] Cross-PR coordination: ship the consumer change before or with the publisher change so older note-service nodes during deploy don't lose symptoms from amendments

## Refs

PR #1249, issue #1246.

blamechris:
## Summary

PR #1249 enriched `note.saved` events emitted by `createNote` / `updateNote` so ai-oversight's `extractSymptoms()` could populate `ctx.new_symptoms` from `chief_complaint` / `subjective` / `new_symptoms`. The same publisher gap existed on two adjacent event paths:

- **`signNote` → `note.signed`** — the consumer already handled `note.signed` (see `review-service.ts:1018`), but the publisher emitted bare `{resourceId, signedBy}`. Compose-and-sign in one shot (no intervening `updateNote` — the natural quick-visit flow) produced an empty symptom set. (#1259)
- **`amendNote` → `note.amended`** — publisher emitted bare `{resourceId, amendedBy, reason, ...}` AND the consumer's `extractSymptoms` event-type guard didn't even recognise `note.amended`. Amendments adding neuro observations to a previously-signed note (the HIPAA-significant content-change path) were silently dropped. (#1260)

## What this PR does

1. **`signNote`** — spread `extractSymptomFieldsFromSections(existing.sections as NoteSection[])` onto the `note.signed` data. Reuses the row already loaded for archival — no extra DB roundtrip.
2. **`amendNote`** — spread `extractSymptomFieldsFromSections(sections)` onto the `note.amended` data. The `sections` parameter is the new amended content (correct — we want the post-amendment symptoms, not the pre-amendment archive).
3. **`extractSymptoms` in `review-service.ts`** — add `note.amended` to the event-type guard so the `subjective` field is consumed on the amendment path. (`chief_complaint` and `new_symptoms` already short-circuit on field presence and pick up the new payload automatically.)

## What this PR does NOT do

- **`cosignNote` is intentionally NOT enriched.** Cosign is a workflow signal on already-saved content — the symptoms in the cosigned note were already processed when it was originally saved or signed. Re-processing them on cosign would be duplicate work, and from a clinical-safety standpoint the cosign moment is *not* a new-content event.

## Tests

Two new regression tests:

- `enriches the note.signed payload with symptom fields from sections (#1259)` — asserts that signing a draft with realistic SOAP `subjective` content emits a `note.signed` event with `chief_complaint` + `new_symptoms` populated.
- `enriches the note.amended payload with symptom fields from the new sections (#1260)` — asserts that amending a signed note with new neuro symptoms emits a `note.amended` event with the post-amendment symptoms in the payload.

Both follow the mock-DB / mock-event pattern established by PR #1249.

**Caveat (per `feedback_mocked_sql_misses_runtime_types` memory):** these are mock-event tests — they verify the publisher payload shape but cannot prove the rule actually fires end-to-end against real PG. Smoke-verification was just completed for the `note.saved` path (#916 Phase 4.1 — Margaret's `ONCO-VTE-NEURO-001` flag fires after a note save). Equivalent live verification for `note.signed` and `note.amended` is a smoke-test follow-up.

## Verification

- `pnpm --filter=@carebridge/clinical-notes test` — 61/61 pass
- `pnpm --filter=@carebridge/ai-oversight test` — 939/939 pass
- `pnpm typecheck --filter=@carebridge/clinical-notes --filter=@carebridge/ai-oversight` — green

Closes #1259, closes #1260.

## Test plan

- [x] Unit tests for both new emit sites
- [x] Consumer-side change to `extractSymptoms` for `note.amended`
- [x] Existing 1037+ tests still pass
- [ ] CI green
- [ ] Manual smoke (post-merge): create a SOAP note for Margaret, sign without prior update — `ONCO-VTE-NEURO-001` should fire on the `note.signed` event (independent of the `note.saved` from createNote).

### DIFF
--- a/services/ai-oversight/src/services/review-service.ts
+++ b/services/ai-oversight/src/services/review-service.ts
@@ -1023,9 +1023,15 @@ function extractSymptoms(event: ClinicalEvent): string[] {
   if (event.type === "vital.created" && event.data.notes) {
     symptoms.push(event.data.notes as string);
   }
-  // For note events, extract from sections if available
+  // For note events, extract from sections if available. note.amended is
+  // included because amendments are the only way to change content after a
+  // note is signed — a clinician adding "new onset headache + numbness" to a
+  // previously-signed note via amendment is exactly the cross-specialty
+  // signal ONCO-VTE-NEURO-001 exists to catch (#1260).
   if (
-    (event.type === "note.saved" || event.type === "note.signed") &&
+    (event.type === "note.saved" ||
+      event.type === "note.signed" ||
+      event.type === "note.amended") &&
     event.data.subjective &&
     typeof event.data.subjective === "string"
   ) {
--- a/services/clinical-notes/src/__tests__/note-service.test.ts
+++ b/services/clinical-notes/src/__tests__/note-service.test.ts
@@ -517,6 +517,45 @@ describe("signNote", () => {
     expect(event.data.resourceId).toBe(NOTE_ID);
   });
 
+  it("enriches the note.signed payload with symptom fields from sections (#1259)", async () => {
+    // Compose-and-sign in one shot (no intervening updateNote) is the
+    // natural flow for a quick visit. Before #1259's fix, this path emitted
+    // a bare { resourceId, signedBy } payload and ai-oversight's
+    // extractSymptoms returned [] — ONCO-VTE-NEURO-001 never fired.
+    const signedDraftRow = {
+      ...existingRow,
+      sections: [
+        {
+          key: "subjective",
+          label: "Subjective",
+          fields: [
+            {
+              key: "chief_complaint",
+              label: "Chief Complaint",
+              value: "Headache with vision changes",
+              field_type: "text",
+              source: "new_entry",
+            },
+            {
+              key: "new_symptoms",
+              label: "New Symptoms",
+              value: ["headache", "vision changes"],
+              field_type: "multiselect",
+              source: "new_entry",
+            },
+          ],
+        },
+      ],
+    };
+    db.willSelect([signedDraftRow]);
+
+    await signNote(NOTE_ID, PROVIDER_ID);
+
+    const event = emitClinicalEvent.mock.calls[0][0];
+    expect(event.data.chief_complaint).toBe("Headache with vision changes");
+    expect(event.data.new_symptoms).toEqual(["headache", "vision changes"]);
+  });
+
   it("throws when note is not found", async () => {
     db.willSelect([]);
 
@@ -859,6 +898,46 @@ describe("amendNote", () => {
     expect(event.data.resourceId).toBe(NOTE_ID);
     expect(event.data.amendedBy).toBe(AMENDER_ID);
     expect(event.data.reason).toBe(AMEND_REASON);
+  });
+
+  it("enriches the note.amended payload with symptom fields from the new sections (#1260)", async () => {
+    // Amendments are the only way to change a signed note. A clinician
+    // adding "new onset headache + numbness" to a previously-signed note
+    // via amendment is exactly the cross-specialty signal
+    // ONCO-VTE-NEURO-001 exists to catch — but pre-#1260 the consumer's
+    // extractSymptoms didn't even handle note.amended.
+    db.willSelect([signedRow]);
+
+    const amendedSections: NoteSection[] = [
+      {
+        key: "subjective",
+        label: "Subjective",
+        fields: [
+          {
+            key: "chief_complaint",
+            label: "Chief Complaint",
+            value: "Now reporting numbness in left arm",
+            field_type: "text",
+            source: "new_entry",
+          },
+          {
+            key: "new_symptoms",
+            label: "New Symptoms",
+            value: ["numbness", "weakness"],
+            field_type: "multiselect",
+            source: "new_entry",
+          },
+        ],
+      },
+    ];
+
+    await amendNote(NOTE_ID, AMENDER_ID, amendedSections, AMEND_REASON);
+
+    const event = emitClinicalEvent.mock.calls[0][0];
+    expect(event.data.chief_complaint).toBe(
+      "Now reporting numbness in left arm",
+    );
+    expect(event.data.new_symptoms).toEqual(["numbness", "weakness"]);
   });
 
   it("allows amending a cosigned note", async () => {
--- a/services/clinical-notes/src/services/note-service.ts
+++ b/services/clinical-notes/src/services/note-service.ts
@@ -318,7 +318,11 @@ export async function signNote(
     patient_id: existing.patient_id,
     provider_id: existing.provider_id,
     timestamp: now,
-    data: { resourceId: noteId, signedBy },
+    data: {
+      resourceId: noteId,
+      signedBy,
+      ...extractSymptomFieldsFromSections(existing.sections as NoteSection[]),
+    },
   });
 
   return {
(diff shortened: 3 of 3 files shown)


blamechris (commented):
## CareBridge Inspector — PR #1266

### Verdict
APPROVE_WITH_NITS

Tight, surgical fix that closes a real clinical-safety gap: the compose-and-sign-in-one-shot flow and the amend-with-new-symptoms flow were both invisible to ONCO-VTE-NEURO-001. Enrichment lives at the right boundary (publisher), the consumer-side guard is extended symmetrically, and the deliberate cosignNote exclusion is well-reasoned. Existing-keys-vs-spread safety, archive ordering, and the choice of `existing.sections` (sign) vs `sections` parameter (amend) are all correct. CI is green (Build/Test/Typecheck).

### Critical
None.

### Suggestions (non-blocking, follow-ups can be filed)

1. **`buildEventSummary` in `services/ai-oversight/src/workers/context-builder.ts:368-389` has no `case "note.amended"`.** It already handles `note.saved` and `note.signed` with a friendly string but falls through to the `default: \`Clinical event: ${event.type}\`` branch for `note.amended`. The symptom payload now flows in via this PR, but the LLM context line will read "Clinical event: note.amended" instead of "Clinical note amended". Low-impact — the symptoms array is what the rule reads — but worth tightening so the LLM gets a consistent event-summary vocabulary. Will file a follow-up.

2. **Test coverage doesn't exercise the `subjective` field (HPI key).** The two new tests assert `chief_complaint` and `new_symptoms` shape. Those two fields are extracted from `event.data` regardless of `event.type` in `extractSymptoms` (lines 1014-1021 of review-service.ts). The field whose extraction is *gated* by the new `note.amended` guard is `subjective` — extracted from `history_of_present_illness` / `hpi` field keys. So the regression test that most directly proves "extractSymptoms now handles note.amended" isn't here. Recommend adding one assertion using an `hpi` field with string content to lock in the guard expansion. Will file a follow-up.

3. **Cast inconsistency: `existing.sections as NoteSection[]` vs the surrounding `existing.sections as ClinicalNote[\"sections\"]` pattern** (lines 330, 405 in the same file). Both resolve to the same shape, but two casts to the same property in adjacent functions reads as drift. Trivial nit — no follow-up needed.

### Praise

- The publisher-side enrichment (instead of having ai-oversight re-read the note row) is the right call: zero extra DB roundtrip, single source of truth for what "the symptoms in this note" means, and the same `extractSymptomFieldsFromSections` helper is reused across `createNote` / `updateNote` / `signNote` / `amendNote`.
- The decision to NOT enrich `cosignNote` is explicitly justified in the PR body with the right reason (workflow signal, not content change) — this is exactly the kind of judgment-call documentation that prevents future "fix" PRs from over-correcting.
- The `existing.sections` (signNote) vs `sections` parameter (amendNote) distinction is subtle and correct — signing captures pre-state, amending captures post-state. Easy to get wrong; this gets it right.
- Test comments cite the issue numbers and explain the clinical scenario (Margaret-style cancer + new neuro symptom path), which makes the regression intent legible to future readers.

PR #1267, opened 2026-05-28
fix(infra): auto-rebuild workspace package dist/ in dev (#1264)

### LINKED ISSUE #1264: dev: workspace package src changes don't reach api-gateway without manual rebuild
## Summary

Services like \`@carebridge/clinical-notes\` are workspace packages (\`"main": "dist/index.js"\`) with no \`dev\` script. Editing \`services/clinical-notes/src/services/note-service.ts\` does not propagate to the running api-gateway under \`pnpm dev\` until someone manually runs \`pnpm --filter=@carebridge/clinical-notes build\`.

This caused a false-negative smoke-test result today: a fix that was merged to main and was on disk in \`src/\` was silently ignored by the running dev server because the \`dist/\` it imported from was hours stale. The fix appeared not to work; in fact it had never been loaded.

## Repro

1. Make any change to \`services/clinical-notes/src/services/note-service.ts\`
2. \`pnpm dev\` (already running)
3. tsx watch logs no reload for the clinical-notes file
4. The api-gateway / ai-oversight pipeline continues running the pre-change behaviour

## Why it bites

- The grep-the-src verification fool's-gold check (\`grep -n 'myNewHelper' services/clinical-notes/src/services/note-service.ts\`) passes — the source has the fix — but the deployed behaviour doesn't reflect it.
- The compiled \`dist/\` mtime is the only reliable signal. Engineers don't habitually check dist mtimes.
- For services that DO have a tsx-watch \`dev\` script (\`api-gateway\`, \`ai-oversight\`, etc.), hot reload works for their own \`src/\` — but never picks up changes in imported \`@carebridge/*\` packages.

## Options

1. **Add a \`dev\` script to each workspace package** that runs \`tsc --watch --noEmit false\` so dist/ is continuously rebuilt. Lowest-friction; aligns each package with its existing tsc step.
2. **Switch importers to import from \`src/\` in dev** via conditional exports or path mapping. Trickier — needs tsx-aware resolution and may break the production build.
3. **Add a turbo \`dev\` pipeline task** that depends on builds of all upstream packages and re-runs them on src changes. Turbo's task graph already knows about this — likely just a config change.

(3) is probably the cleanest if turbo can do it; (1) is the safest fallback.

## Acceptance criteria

- [ ] Editing any \`@carebridge/*\` workspace package's \`src/\` is reflected in the running \`pnpm dev\` process without a manual \`pnpm build\` step
- [ ] Documented in CLAUDE.md or contributing docs

## Refs

Surfaced today during smoke-verify of merged #1244 / #1248 / #1249 against #916. Real-world impact: smoke-test reported failure for #1249 when the fix had simply not been built.

blamechris:
## Summary

Before this change, library packages under \`packages/*\` and the non-server services under \`services/*\` declared no \`dev\` script. They export from \`dist/\` (\`"main": "dist/index.js"\`), so consumers running \`tsx watch\` on their own \`src/\` were loading workspace packages from a \`dist/\` snapshot frozen at the last \`pnpm build\`. Editing a workspace package's source had no runtime effect until someone manually rebuilt it.

This produced a real false-negative smoke-test result on 2026-05-27: PR #1249's \`note.saved\` payload enrichment was on disk in \`src/\` but the running api-gateway was importing yesterday's \`dist/\`, so the symptom-bearing event payload appeared not to work. After a manual \`pnpm --filter=@carebridge/clinical-notes build\` and restart, the same smoke test passed cleanly (Margaret's \`ONCO-VTE-NEURO-001\` flag fires on note save).

## What this PR does

1. **Adds \`"dev": "tsc --watch --preserveWatchOutput"\`** to every \`@carebridge/*\` package that has a tsc build step but lacked a dev script: 5 services (\`auth\`, \`clinical-data\`, \`clinical-notes\`, \`epic-connector\`, \`fhir-gateway\`, \`patient-records\`) and 9 packages (\`ai-prompts\`, \`db-schema\`, \`logger\`, \`medical-logic\`, \`outbox\`, \`phi-sanitizer\`, \`redis-config\`, \`shared-types\`, \`test-utils\`).
2. **Bumps the root \`pnpm dev\` concurrency** from turbo's default 10 to 25 so all 21 persistent tasks (15 tsc-watch + 4 tsx-watch + 2 Next.js) can run together.
3. **Adds a one-paragraph note to CLAUDE.md** explaining the mtime contract — pointing engineers at \`dist/\` mtime as the diagnostic when a fix doesn't seem to take.

\`tsx watch\` natively notices workspace package \`dist/*.js\` changes through the resolved pnpm symlinks. Verified live during testing: editing any package's \`src/\` now triggers a \`tsc\` rebuild of its \`dist/\`, which in turn causes the consuming service's tsx watch to restart automatically.

## Verification

Started \`pnpm dev\` with the new config and observed:

- Turbo reports \`Running dev in 25 packages\` (was 5 before)
- 15 \`tsc --watch\` processes running alongside 4 \`tsx watch\` services + 2 Next.js dev servers
- Initial startup logs show \`@carebridge/api-gateway:dev: [tsx] change in ./../../packages/shared-types/dist/index.js Restarting...\` — proving the dist propagation works
- All services come up healthy: api-gateway on :4000, ai-oversight on :4001, notifications, scheduling, clinician-portal on :3000, patient-portal on :3001

The reformatted JSON in \`services/epic-connector/\`, \`services/fhir-gateway/\`, \`services/patient-records/\` package.json is from \`jq\` expanding the one-line script/exports objects to multi-line standard formatting — they now match the other 11 package.jsons.

\`pnpm typecheck\` — 38/38 green.

## What this does NOT do

Doesn't address turbo's task graph ordering for dev tasks. All dev tasks start in parallel; if a service's tsx-watch boots before its upstream packages finish their first \`tsc\` build, the service might temporarily fail to import. Verified empirically that this self-corrects within ~5 seconds (tsx-watch retries on the next dist change). If this becomes a real issue, a follow-up would add \`dependsOn: ["^build"]\` to the dev task to require a fresh one-shot build before the watchers start.

Closes #1264.

## Test plan

- [x] \`pnpm typecheck\` — green
- [x] \`pnpm dev\` boots all 25 packages, services come up healthy
- [x] Editing a workspace package \`src/\` triggers tsx-watch reload in consumers
- [ ] CI green

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -23,6 +23,13 @@ pnpm db:seed                  # Seed dev data (DVT scenario patient)
 pnpm dev                      # Start all services + apps
 ```
 
+`pnpm dev` runs `tsc --watch` for every workspace library package alongside
+`tsx watch` for each service. Editing `src/` in any `@carebridge/*` package
+rebuilds its `dist/` and the consuming service hot-reloads automatically.
+(If you ever see a fix that's on disk but isn't taking effect at runtime,
+check that the package's `dist/` mtime is newer than its `src/` — that
+contract being broken is the root cause #1264 was filed to prevent.)
+
 ## Monorepo Structure
 - `packages/` — Shared libraries (types, validators, medical-logic, db-schema, ai-prompts)
 - `services/` — Backend services (api-gateway, clinical-notes, ai-oversight, clinical-data, patient-records, auth, notifications, fhir-gateway)
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "smoke-test": "tsx tooling/smoke-test.ts"
   },
   "devDependencies": {
+    "concurrently": "^9.1.0",
     "turbo": "^2.4.0",
     "typescript": "^5.7.0",
     "tsx": "^4.19.0",
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "private": true,
   "packageManager": "pnpm@9.15.4",
   "scripts": {
-    "dev": "turbo dev",
+    "dev": "turbo dev --concurrency=25",
     "build": "turbo build",
     "lint": "turbo lint",
     "typecheck": "turbo typecheck",
--- a/packages/ai-prompts/package.json
+++ b/packages/ai-prompts/package.json
@@ -13,6 +13,7 @@
   },
   "scripts": {
     "build": "tsc",
+    "dev": "tsc --watch --preserveWatchOutput",
     "typecheck": "tsc --noEmit -p tsconfig.check.json",
     "test": "vitest run",
     "eval": "vitest run --config vitest.config.ts golden-eval",
--- a/packages/db-schema/package.json
+++ b/packages/db-schema/package.json
@@ -13,6 +13,7 @@
   },
   "scripts": {
     "build": "tsc",
+    "dev": "tsc --watch --preserveWatchOutput",
     "typecheck": "tsc --noEmit",
     "test": "node --test --import tsx src/__tests__/*.test.ts",
     "generate": "drizzle-kit generate",
--- a/packages/logger/package.json
+++ b/packages/logger/package.json
@@ -13,6 +13,7 @@
   },
   "scripts": {
     "build": "tsc",
+    "dev": "tsc --watch --preserveWatchOutput",
     "test": "vitest run",
     "typecheck": "tsc --noEmit",
     "clean": "rm -rf dist"
--- a/packages/medical-logic/package.json
+++ b/packages/medical-logic/package.json
@@ -13,6 +13,7 @@
   },
   "scripts": {
     "build": "tsc",
+    "dev": "tsc --watch --preserveWatchOutput",
     "typecheck": "tsc --noEmit -p tsconfig.check.json",
     "test": "vitest run",
     "clean": "rm -rf dist",
--- a/packages/outbox/package.json
+++ b/packages/outbox/package.json
@@ -13,6 +13,7 @@
   },
   "scripts": {
     "build": "tsc",
+    "dev": "tsc --watch --preserveWatchOutput",
     "typecheck": "tsc --noEmit",
     "test": "vitest run",
     "clean": "rm -rf dist"
--- a/packages/phi-sanitizer/package.json
+++ b/packages/phi-sanitizer/package.json
@@ -13,6 +13,7 @@
   },
   "scripts": {
     "build": "tsc",
+    "dev": "tsc --watch --preserveWatchOutput",
     "typecheck": "tsc --noEmit",
     "test": "vitest run",
     "clean": "rm -rf dist"
--- a/packages/redis-config/package.json
+++ b/packages/redis-config/package.json
@@ -13,6 +13,7 @@
   },
   "scripts": {
     "build": "tsc",
+    "dev": "tsc --watch --preserveWatchOutput",
     "typecheck": "tsc --noEmit",
     "clean": "rm -rf dist",
     "test": "vitest run"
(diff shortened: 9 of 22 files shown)


blamechris (commented):
## CareBridge Inspector — PR #1267

### Verdict
REQUEST_CHANGES_OPTIONAL

The library-tier fix is correct and surgical: 15 packages × `tsc --watch --preserveWatchOutput` exactly matches the set of `"main": "dist/..."` libs without an existing dev script (Grep confirms 18 dist-exporting packages; 4 server services already had `tsx watch`, and `portal-shared` / `ui-tokens` export from `src/` so don't need it). Concurrency math (21 = 15 tsc-watch + 4 tsx-watch + 2 Next.js, default 10, bumped to 25) checks out, `turbo.json` already has `dev: { cache: false, persistent: true }` so no Turbo work needed, the jq reformat is cosmetic, and the CLAUDE.md mtime-diagnostic paragraph is accurate and zero-attribution-clean. macOS FD limits (`ulimit -n` = 1048576) are not a real risk.

### Critical
**The same #1264 bug class still exists for cross-service dist exports.** Four services keep `"main": "dist/index.js"` but their only `dev` script is `tsx watch src/server.ts` — which boots their own runtime but never refreshes their `dist/`:
- `services/ai-oversight` — imported by `services/api-gateway/src/routers/ai-oversight.ts:22`
- `services/scheduling` — imported by `services/api-gateway/src/routers/scheduling.ts:43`
- `services/notifications` — declared as a dep of `api-gateway`, `ai-oversight`, `scheduling`
- `services/api-gateway` — declared as a devDep of `portal-shared`, `clinician-portal`, `patient-portal` (likely types-only via `inferRouterInputs`, but still dist-resolved)

So if I edit `services/ai-oversight/src/routers/...` today, ai-oversight's own tsx-watch restarts the worker — but api-gateway will keep loading yesterday's `ai-oversight/dist/index.js` and the tRPC router shape goes stale. The clinical-notes false-negative reproduces here too, just across a different boundary. Fix is the same shape: those four services need *both* `tsx watch src/server.ts` (runtime) AND a parallel `tsc --watch` (dist freshness for consumers) — `pnpm run -r --parallel` or a `concurrently` invocation. Not a blocker for the PR-as-described, but it leaves #1264's premise (dist/mtime contract holds) materially false for the most-imported boundary in the repo.

### Suggestions (non-blocking, follow-ups will be filed)

1. **`incremental: true` in `tsconfig.base.json` for cold-start.** 15 cold `tsc --watch` processes each redo the full program parse on `pnpm dev` startup. `tsc --watch` keeps an in-memory program so steady-state rebuilds are already fast, but the cold-start cost compounds across 15 watchers. Adding `incremental: true` + `.tsbuildinfo` to each package would cut first-boot time noticeably with zero behavioral risk. Follow-up.

2. **Concurrency headroom is thin (4 slots).** PR body acknowledges this. As soon as 4 more workspace libs land, `pnpm dev` will silently queue tasks and a service can boot before its upstream dist exists. Either set concurrency to `auto` / a round number like `50`, or assert headroom in CI by counting persistent dev scripts. Follow-up.

3. **PR body undercounts services touched (says "5 services" then lists six: auth, clinical-data, clinical-notes, epic-connector, fhir-gateway, patient-records).** Cosmetic; the diff itself is correct.

### Praise
- The grep-discoverable invariant — every `"main": "dist/"` package has a `dev` script — is exactly the kind of one-liner future readers can audit. The follow-up issue could add a `pnpm` script to assert it.
- The CLAUDE.md paragraph turns the dist/mtime contract into a debuggable signal ("check that the package's `dist/` mtime is newer than its `src/`"). That's the kind of forensic hint that pays back the second time someone gets bitten.
- Verification log in the PR body cites the exact tsx-restart message (`[tsx] change in ./../../packages/shared-types/dist/index.js Restarting...`) — that's the evidence the symlink-through-pnpm propagation actually works, not just that the watchers spawn.
- Zero attribution clean throughout.

PR #1268, opened 2026-05-28
fix(portal): unique React keys for note version history (#1265)

### LINKED ISSUE #1265: clinician-portal: duplicate React key in note version history (app/notes/[id]/page.tsx:204)
## Summary

\`app/notes/[id]/page.tsx:204\` renders the version-history list with \`<div key={v.version}>\`. Multiple \`note_versions\` rows share the same numeric \`version\` (sign and cosign both archive at \`existing.version\` without bumping — disambiguated only by \`lifecycle_event\`). React emits:

\`\`\`
Encountered two children with the same key, \`1\`. Keys should be unique...
\`\`\`

## Repro

1. Create a SOAP note for Margaret as \`dr.smith\` (v1 draft archived).
2. Sign it (signed archive row at v1).
3. Open the note detail. Console shows the duplicate-key warning.

## Fix

The \`note_versions\` table has \`(note_id, version, lifecycle_event)\` as the disambiguator. The React key should reflect that:

\`\`\`tsx
<div key={\`\${v.version}-\${v.lifecycle_event}\`} className="detail-row">
\`\`\`

Or even more uniquely, use \`v.saved_at\` (ISO timestamp) since each archive is at a distinct moment:

\`\`\`tsx
<div key={\`\${v.version}-\${v.saved_at}\`} className="detail-row">
\`\`\`

## Acceptance criteria

- [ ] React no longer warns about duplicate keys when a note has both \`draft\` and \`signed\` (or \`signed\` and \`cosigned\`) at the same version number
- [ ] Tested by the same smoke-test path that surfaced this

## Refs

Found during smoke-verify against #916 today.

blamechris:
## Summary

The note detail page (\`apps/clinician-portal/app/notes/[id]/page.tsx:204\`) was rendering the version history list with \`<div key={v.version}>\`. \`note_versions\` rows share the same numeric \`version\` because \`signNote\` and \`cosignNote\` both archive at \`existing.version\` without bumping it — disambiguated only by \`lifecycle_event\`. A single note progressing draft → signed produces two rows keyed \`"1"\`, and React warns:

\`\`\`
Encountered two children with the same key, \`1\`. Keys should be unique...
\`\`\`

## Fix

Combine \`version\`, \`lifecycle_event\`, and \`saved_at\` into the key. \`saved_at\` is the ultimate tie-breaker — even if a future bug emitted two rows with the same version+lifecycle_event (retroactive correction, etc.) the ISO timestamp keeps keys unique.

## Verification

- \`pnpm typecheck --filter=@carebridge/clinician-portal\` — green
- Reproduced the warning before the fix by signing any draft note as \`dr.smith\` (caught during the post-Wave-8 smoke-test re-run yesterday)

Closes #1265.

## Test plan

- [x] Typecheck green
- [ ] Manual: sign a draft note → React DevTools should no longer warn about duplicate keys
- [ ] CI green

### DIFF
--- a/apps/clinician-portal/app/notes/[id]/page.tsx
+++ b/apps/clinician-portal/app/notes/[id]/page.tsx
@@ -201,7 +201,10 @@ function NoteDetailContent() {
         <div className="detail-card">
           <div className="detail-card-title">Version History</div>
           {versions.map((v) => (
-            <div key={v.version} className="detail-row">
+            <div
+              key={`${v.version}-${v.lifecycle_event}-${v.saved_at}`}
+              className="detail-row"
+            >
               <span className="detail-label">v{v.version}</span>
               <span className="detail-value">
                 {formatDate(v.saved_at)} — {v.saved_by}


blamechris (commented):
## CareBridge Inspector — PR #1268

### Verdict
Approve-with-suggestions. The composite key is correct and minimal. The fix lands the React warning. However, this PR exposes (but does not introduce) a latent server-side bug in `getNoteById` that has the same root cause as the key collision.

### Critical
1. **Underlying sort instability still present in `getNoteById`.** `services/clinical-notes/src/services/note-service.ts:591` orders the `versions` payload by `desc(noteVersions.version)` only. Within a same-version pair (e.g. signed + cosigned both archived at `v1`), PostgreSQL is free to return the rows in any order — the UI will render the timeline in a non-deterministic sequence across page reloads/cache invalidations. Note `getVersionHistory` at L528 already orders by `asc(saved_at)` for exactly this reason and documents it. `getNoteById` should mirror that (`.orderBy(desc(noteVersions.saved_at))`) so the React list order is stable too. This is the same root cause as the duplicate-key warning, just one layer deeper; fixing only the key papers over the symptom.

### Suggestions
1. **Add a regression test.** `apps/clinician-portal/src/__tests__/` has many UI tests; none cover this. A React Testing Library test that renders `NoteDetailContent` against a mocked `notes.getById` returning two versions with identical `version` and different `lifecycle_event` (`"signed"` + `"cosigned"`) and asserts no console.error warning + both rows present would prevent future regressions in the same surface.
2. **Render the lifecycle event in the row.** Now that `lifecycle_event` is part of the key, the user can see two `v1` rows but cannot tell which is which. A tiny badge (`v1 · signed` / `v1 · cosigned`) would close the original UX gap — the column was added precisely to label these transitions.
3. **`saved_at` precision.** ISO millisecond resolution + sequential awaited inserts within a single call mean true collisions are effectively impossible — but `lifecycle_event` is already in the composite key so the belt-and-suspenders is fine. No change required.

### Praise
- Composite key matches the row-uniqueness tuple `(version, lifecycle_event, saved_at)` exactly; deterministic, no `Math.random`/index fallback.
- PR body cleanly traces the failure mode back to `signNote`/`cosignNote` not bumping `version` — accurate root-cause framing.
- Type-safe: `NoteVersion.lifecycle_event` is declared in `@carebridge/shared-types`, so no `any` cast needed.

PR #1293, opened 2026-05-28
test(ai): PG integration for in-flight dedup probe (#1252)

### LINKED ISSUE #1252: ai-oversight: integration test for in-flight dedup probe against real PostgreSQL
## Context

Identified during review of PR #1244.

PR #1244 fixed a production-blocking SQL bug where `gte(reviewJobs.created_at, inFlightCutoff)` compared a `text` column to a `timestamp with time zone` expression, causing every BullMQ review job to fail with:

```
operator does not exist: text >= timestamp with time zone
```

The reason this shipped is documented in the PR body:

> - `review-service-db-clock-idempotency.test.ts` mocks drizzle's `sql` helper, so it never executes the actual SQL.
> - `review-idempotency.test.ts` mocks the entire DB layer; the `gte` mock doesn't validate the runtime types.
> - No integration test exercises the dedup probe against real PostgreSQL.

The new regression test in PR #1244 asserts on the mocked predicate shape but still does not catch text↔timestamptz type mismatches at the database level. The same class of bug — Drizzle helper returning a different PG type than the schema column — could be reintroduced on any future predicate touching a non-text column.

## Acceptance criteria

- [ ] New `review-service-idempotency-pg.integration.test.ts` (or extend an existing PG integration suite) that:
  - [ ] Spins up a real PostgreSQL (Testcontainers or the existing dev compose stack)
  - [ ] Applies the production schema (`drizzle/*.sql`)
  - [ ] Inserts a `review_jobs` row with status `processing` and a known `created_at`
  - [ ] Calls `processReviewJob` against a re-delivered event and asserts no second `review_jobs` row is created (dedup short-circuit)
  - [ ] Calls `processReviewJob` for a stale `processing` row and asserts processing proceeds (orphan recovery)
- [ ] Test runs in CI on the existing PG-required workflow
- [ ] Cross-link from this issue in the next PR description

## Notes

Per `feedback_mocked_sql_misses_runtime_types` (user memory) the ai-oversight unit-test suite mocks Drizzle wholesale — this class of bug is structurally impossible to catch in those tests.

## Origin

From-review follow-up for PR #1244.

> "No integration test exercises the dedup probe against real PostgreSQL." — PR #1244 author, explicit acknowledgement in the PR body.

blamechris:
## Summary

Adds a real-PostgreSQL integration test for the review-service in-flight dedup probe — the SQL surface where PR #1244 fixed a production-blocking `text >= timestamp with time zone` operator-mismatch error. Closes #1252.

## Why integration is needed

Every existing review-service test under `services/ai-oversight/src/__tests__/` mocks `drizzle-orm` (see `review-idempotency.test.ts` and `review-service-db-clock-idempotency.test.ts`): they validate the constructed SQL **shape** in JavaScript but never send it to a real PostgreSQL planner. That structural gap is exactly why the production bug from #1244 shipped — and per the user memory `feedback_mocked_sql_misses_runtime_types`, the same class of bug (Drizzle helper returning a different PG type than the schema column) can be reintroduced on any future predicate touching a non-text column.

The new test reconstructs the exact dedup probe query from `review-service.ts` and runs it against a real PostgreSQL instance. The bug PR #1244 fixed reproduces cleanly at the DB level if the `::timestamptz` cast on `reviewJobs.created_at` is reverted:

```
operator does not exist: text >= timestamp with time zone
```

Mutation-verified locally: removing the cast from the test probe fails all 4 cases with that exact error message.

## What's covered

- `does not throw 'operator does not exist'` — primary type-bug guard
- `finds a fresh 'processing' row inside the in-flight window` — cast preserves match semantics
- `ignores a stale 'processing' row` — orphan-recovery branch still works
- `matches a terminal-status row regardless of created_at age` — terminal-status OR branch unaffected

## Harness

Follows the established pattern from `packages/outbox/src/__tests__/batch-disjointness.int.test.ts` and `services/epic-connector/src/__tests__/persistence-source-system.int.test.ts`:

- Gated on `TEST_DATABASE_URL` via `describe.skipIf(!TEST_URL)` — no-op locally without docker-compose up
- Forces `DATABASE_URL` to `TEST_DATABASE_URL` for safety (test inserts seed rows)
- Generates a per-process throwaway `PHI_ENCRYPTION_KEY` before importing `@carebridge/db-schema` so the encrypted `patients.name` column can be written without leaking a real secret into CI
- Targeted cleanup via `id LIKE 'test-1252-%'` so the test never touches dev seed data
- Closes the `postgres-js` client in `afterAll` so Vitest doesn't hang

CI already supplies `TEST_DATABASE_URL` from the `test` job in `.github/workflows/ci.yml`, so no workflow changes are needed.

## Test plan

- [x] `pnpm typecheck` — 38/38 tasks clean
- [x] `pnpm lint` — 21/21 tasks clean
- [x] `pnpm --filter @carebridge/ai-oversight test` — 943/943 passing (37 files including new one)
- [x] Locally with `TEST_DATABASE_URL` set — 4/4 new tests pass
- [x] Locally without `TEST_DATABASE_URL` — 4/4 skip cleanly
- [x] Mutation test: removing the `::timestamptz` cast from the test's probe builder reproduces `operator does not exist: text >= timestamp with time zone` on all 4 cases

## Refs

- Closes #1252
- Guards the fix shipped in #1244

### DIFF
--- a/services/ai-oversight/src/__tests__/review-service-in-flight-probe.int.test.ts
+++ b/services/ai-oversight/src/__tests__/review-service-in-flight-probe.int.test.ts
@@ -1,0 +1,252 @@
+/**
+ * Integration test for issue #1252 — exercise the review-service in-flight
+ * dedup probe against a real PostgreSQL instance, not a vi.mock("drizzle-orm")
+ * stub.
+ *
+ * Why this test exists
+ * --------------------
+ * PR #1244 fixed a production-blocking bug where the probe's
+ * `gte(reviewJobs.created_at, inFlightCutoff)` predicate compared a `text`
+ * column to a `timestamp with time zone` expression, causing every BullMQ
+ * review job to fail with:
+ *
+ *     operator does not exist: text >= timestamp with time zone
+ *
+ * The bug shipped because every existing review-service test under
+ * services/ai-oversight/src/__tests__/ mocks `drizzle-orm` (see
+ * `review-idempotency.test.ts` and `review-service-db-clock-idempotency.test.ts`):
+ * the SQL is constructed in JS but never sent to a real planner that would
+ * reject the type mismatch. The fix wraps the column side in a
+ * `${col}::timestamptz` cast — but only a real-PG test can prove the cast is
+ * still in place. Per the user memory `feedback_mocked_sql_misses_runtime_types`,
+ * the same class of bug (Drizzle helper returning a different PG type than the
+ * schema column) could be reintroduced on any future predicate touching a
+ * non-text column. This file is the structural guard.
+ *
+ * Gating
+ * ------
+ * Gated on TEST_DATABASE_URL — no-op locally without docker-compose up.
+ * CI's `test` job (.github/workflows/ci.yml) supplies it after running
+ * migrations. Mirrors the harness shape already used by
+ * `packages/outbox/src/__tests__/batch-disjointness.int.test.ts` and
+ * `services/epic-connector/src/__tests__/persistence-source-system.int.test.ts`
+ * so behaviour is consistent across the repo.
+ */
+import { describe, it, expect, beforeAll, beforeEach, afterAll } from "vitest";
+import crypto from "node:crypto";
+import { and, eq, gte, inArray, or, sql } from "drizzle-orm";
+
+// Safety invariant identical to the other *.int.test.ts harnesses: force
+// DATABASE_URL to TEST_DATABASE_URL so a developer with DATABASE_URL pointing
+// at a dev/prod DB cannot accidentally mutate it. The probe under test is
+// read-only, but the test inserts seed rows and cleans up afterwards.
+const TEST_URL = process.env.TEST_DATABASE_URL;
+if (TEST_URL) {
+  if (process.env.DATABASE_URL && process.env.DATABASE_URL !== TEST_URL) {
+    console.warn(
+      `[review-service-in-flight-probe.int] Overriding DATABASE_URL (${process.env.DATABASE_URL}) ` +
+        `with TEST_DATABASE_URL for test safety.`,
+    );
+  }
+  process.env.DATABASE_URL = TEST_URL;
+}
+
+// `patients.name` is an encryptedText column, so seeding a patient (needed to
+// satisfy review_jobs.patient_id FK) requires a PHI_ENCRYPTION_KEY. Match the
+// pattern from `persistence-source-system.int.test.ts`: generate a per-process
+// throwaway key BEFORE the dynamic import of @carebridge/db-schema so the
+// encryption module's memoised key read picks it up.
+if (TEST_URL && !process.env.PHI_ENCRYPTION_KEY) {
+  process.env.PHI_ENCRYPTION_KEY = crypto.randomBytes(32).toString("hex");
+}
+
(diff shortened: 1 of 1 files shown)


blamechris (commented):
## CareBridge Inspector — PR #1293 (Harness Pattern)

**Verdict:** APPROVE. This is a well-considered first PG integration test for `services/ai-oversight` and actually improves on the existing int-test pattern in two important ways. Safe to merge.

### Critical
None.

### Suggestions (non-blocking)

1. **Constant drift risk.** The test re-declares `IN_FLIGHT_WINDOW_SEC = 150` and `TERMINAL_REVIEW_STATUSES = [...]` as local literals (lines ~71-73). Both are already exported from `services/ai-oversight/src/services/review-service.ts` (lines 83-87, 103). If a future PR widens the window to e.g. 180s or adds `llm_circuit_breaker` to the terminal set, the production query updates but the test still passes a stale window/status set — the probe shape silently drifts from production. Importing the exported constants would make the test self-syncing.

2. **Probe shape duplication.** `runInFlightProbe()` is a verbatim copy of `review-service.ts` lines 137-168. The inline-copy approach is justified in the file header (avoids dragging in Claude/BullMQ deps), but a single internal export — e.g. `buildInFlightProbeQuery(triggerEventId, db)` from `review-service.ts` — would let the test exercise the *actual* production SQL builder and would close the drift gap on (1) as a bonus. Worth considering as a follow-up.

3. **Cleanup runs `beforeEach`, not `afterAll`.** If a test process crashes mid-suite, `test-1252-*` rows linger in the DB until the next run's `beforeEach`. Not a correctness issue (next run cleans them), and matches the existing pattern, but an `afterAll` mirror would leave the DB tidy.

### Praise

- **Marker-prefixed cleanup (`test-1252-`) is a real harness improvement.** `batch-disjointness.int.test.ts` TRUNCATEs the whole table; `persistence-source-system.int.test.ts` does `DELETE ... LIKE '%'`. Both break parallel int tests sharing the DB. The prefixed `LIKE` here is the right pattern for the future and worth documenting as the canonical approach.
- **First test (`does not throw`) catches the bug even with zero rows.** The `text >= timestamptz` error fires at planner time, so the empty-table assertion is a genuine, non-tautological mutation probe — verified against the schema (`created_at: text("created_at")` in `packages/db-schema/src/schema/ai-flags.ts:109`).
- **`beforeAll` fail-fast on `SELECT 1 FROM review_jobs`** — if CI ever forgets `pnpm db:migrate` the test surfaces a readable error instead of silently skipping. Important guard.
- **Per-process `PHI_ENCRYPTION_KEY` set before the dynamic `db-schema` import.** Correct ordering — `getKey()` memoises on first read.
- **`TEST_DATABASE_URL` override of `DATABASE_URL` with a structured warning** matches the established safety invariant.
- **`describe.skipIf` + `$client.end()` in `afterAll`** — local `pnpm test` without docker-compose stays green, CI Vitest doesn't hang on the open postgres-js handle.

Workflow check: `.github/workflows/ci.yml` lines 50-79 provision the postgres:16-alpine service, set `TEST_DATABASE_URL`, run `pnpm db:migrate`, then `pnpm test`. The test will execute (not silently skip) in CI. Confirmed.

